PR #2, opened 2021-09-23
pass pendings txns to Stake and GlpSwap

gkrasulya:


src/components/Glp/GlpSwap.js
```diff
@@ -131,11 +131,11 @@ export default function GlpSwap(props) {
   })
 
   const tokenAllowanceAddress = swapTokenAddress === AddressZero ? nativeTokenAddress : swapTokenAddress
-  const { data: tokenAllowance, mutate: updateTokenAllowance } = useSWR([active, chainId, tokenAllowanceAddress, "allowance", account, glpManagerAddress], {
+  const { data: tokenAllowance, mutate: updateTokenAllowance } = useSWR([active, chainId, tokenAllowanceAddress, "allowance", account || AddressZero, glpManagerAddress], {
```
gkrasulya, line 134: ideally we should not request data at all in case we don't have `account`. and in some other places as well
or is there some idea behind?

it easy to not request – useSWR won't send requests if you pass falsie `key` argument

xvi10: oh i see, will keep that in mind for the future requests

xvi10 (approved):

PR #7, opened 2021-09-30
fix debug error

gkrasulya:
reviewed & tested.
impacted parts:
- sell GLP works, error message doesn't double
- any contract call should show debug info for not identified error

checked:
- sell GLP works
- in bad case message is not duplicated and correct error is shown
- tested errors with mocked error with compound feature

### DIFF
--- a/src/components/Glp/GlpSwap.js
+++ b/src/components/Glp/GlpSwap.js
@@ -482,10 +482,6 @@ export default function GlpSwap(props) {
     .then(async () => {
       handleFulfilled();
     })
-    .catch((e) => {
-      console.error(e)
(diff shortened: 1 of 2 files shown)


src/components/Glp/GlpSwap.js
```diff
@@ -482,10 +482,6 @@ export default function GlpSwap(props) {
     .then(async () => {
       handleFulfilled();
     })
-    .catch((e) => {
-      console.error(e)
-      toast.error("Sell failed.")
-    })
```
gkrasulya, line 488: it duplicates error message
`callContract` func will toast this error too

xvi10 (approved):

PR #33, opened 2021-11-23
show liquidation popup

gkrasulya:


src/components/Exchange/TradeHistory.js
```diff
@@ -41,6 +45,28 @@ function getOrderActionTitle(action) {
   return `${actionDisplay} Order`
 }
 
+function renderLiquidationTooltip(liquidationData) {
+  const minCollateral = liquidationData.size.mul(BASIS_POINTS_DIVISOR).div(MAX_LEVERAGE)
```
gkrasulya, line 49: small nuance, for older liquidations that may be not accurate (since we had lower max leverage before)
i don't think it's critical but it's possible to save actual max leverage un subgraph for each liquidation

PR #47, opened 2021-12-18
Avalanche Integration

xvi10:


gdev8317:
i've merged all my changes for Avalanche into `avax`. i guess i edited other files but in case i broke something please tell me

midasdev711:
> i've merged all my changes for Avalanche into `avax`. i guess i edited other files but in case i broke something please tell me

I see you removed yarn.lock. I added it.

PR #71, opened 2022-02-04
Add Connect Wallet button for Coinbase Wallet.

0xAdamn:
Unified some css. Added Coinbase Wallet image.

I saw this request in the discord and thought it may be a good first PR  https://discord.com/channels/837652500948713502/838541637045977099/937385090349142027


xvi10:
Hi @0xAdamn, thank you very much for this pr, seems there are some conflicts with the master branch, could you resolve them? We will test and merge this pr after

0xAdamn:
Hi @xvi10 - I've merged upstream master in. Should be all good now.

gdev8317:
@0xAdamn hey! I'm sorry for such big delays in our replies, we're a bit swamped with urgent tasks. 

i've reviewed and tested the PR. Everything looks good except one bug i found. Steps to reproduce:
- Connect Coinbase Wallet
- Disconnect wallet (from user menu in top bar)
- Connect MetaMask
- Refresh

After refresh i still have Coinbase Wallet connected. Probably it caches somewhere. Will you be able to look into this case?

And again thank you for this change, hope we will merge it soon!

0xAdamn:
@gdev8317 @xhiroz Looks like this is a known bug/feature with the version of @web3-react/core that is being used (v6).

v6 is now deprecated/maintenance mode and v8 upgrade is recommended. 

However, it also sounds like this is actually not currently supported functionality anyway as deactivate() which disconnects the wallet is a superficial action only provided for each connector as the disconnection of the wallet can only actually be done by the user in the wallet (i.e. external to the app). (ref: see linked comment/issue below)

Looks like there may be a (hacky/temporary) if the lib is upgraded to v8 (see last comment in linked thread below) but that is beyond the scope of this PR imo.

Let me know your thoughts.

ref: https://github.com/NoahZinsmeister/web3-react/issues/377#issuecomment-1018717549

0xAdamn:
I've just worked on this a bit more and found another way around it - I'm removing the previous wallet details from localStorage on disconnect which forces/emulates a proper disconnect.

Check it out and let me know what you think

PR #77, opened 2022-02-09
add helper for rpc issues on arbitrum

mark-antony1:
- user should see pop up when they log in with metamask for arbitrum but are having rpc connectivity issues


gdev8317:
thank you for PR and sorry for late reply! we've implemented similar functionality in separate PR

PR #101, opened 2022-03-15
convert rem from 15.5px to 16px

midasdev711:


src/Shared.css
```diff
@@ -8,7 +8,7 @@ body,
   color: white;
   font-family: "Relative";
   letter-spacing: 0.5px;
-  font-size: 15.5px;
+  font-size: 16px;
```
jonathan-s, line 11: Though to be fair, if you want it to be easy to overview and understand changing the base font-size to 10px would be better and then adjusting all the rem values accordingly.

Issue #120, opened 2022-04-05
Site SEO file changes cause site name to be "Lee Robinson"


ikmckenz:
Recent changes to the site SEO left the site name as "Lee Robinson". As far as I can tell, this doesn't impact the site visually at all, but it does impact third party tools that interact with GMX. 

For example when using Fireblocks to connect to GMX, the platform is listed as "Lee Robinson":

![gmx_fireblocks](https://user-images.githubusercontent.com/13459320/161833982-d747cd35-f535-4654-9142-756fb260aff2.jpg)

Fix should be pretty simple, looks like the code is only in one place.

PR #121, opened 2022-04-05
Change SEO site name to GMX

ikmckenz:
Small change to fix SEO site name

Fixes issue #120


xvi10:
Will merge, thank you for this

PR #124, opened 2022-04-09
Add referrals page

vipineth:
- [x] Add referral page
- [x] Add support for creating users referral code
- [x] Add support for user to register a referral code
- [x] Update user's referral code
- [x] Display stats like rebate distributions


src/views/Referrals/Referrals.js
```diff
@@ -0,0 +1,601 @@
...
+    trades: 0,
+    volume: bigNumberify(0),
+  };
+};
+
+const TRADERS = "Traders";
+const AFFILIATES = "Affiliates";
+let TAB_OPTIONS = [TRADERS, AFFILIATES];
+
+function getDollarValue(value) {
```
gdev8317, line 51: i guess we use "usd" everywhere, may be better to use it here as well for consistency

src/views/Referrals/Referrals.js
```diff
@@ -0,0 +1,601 @@
...
+  return `$${formatAmount(value, USD_DECIMALS, 2, true, "0.00")}`;
+}
+
+function getErrorMessage(value) {
+  let trimmedValue = value.trim();
+  let invalid = /\s/;
+  if (invalid.test(trimmedValue)) {
+    return "The referral code can't contain spaces.";
```
gdev8317, line 59: it should not be bigger than 31 as well (it's bytes 32 bytes in contract, but ether's formatBytes32String accepts 31 only afaik)

src/views/Referrals/Referrals.js
```diff
@@ -0,0 +1,601 @@
...
+  }
+  return "";
+}
+
+function Referrals({ connectWallet, setPendingTxns, pendingTxns }) {
+  const { active, account, library } = useWeb3React();
+  const { chainId } = useChainId();
+  const { infoTokens } = useInfoTokens(library, chainId, active, undefined, undefined);
```
gdev8317, line 67: i guess last 2 args can be omitted

src/views/Referrals/Referrals.js
```diff
@@ -0,0 +1,601 @@
...
+  let [activeTab, setActiveTab] = useState(TRADERS);
+  const referralsData = useReferralsData(chainId, account);
+  const [referralsDataState, setReferralsDataState] = useState(referralsData);
+  const { userReferralCode } = useUserReferralCode(library, chainId, account);
+
+  useEffect(() => {
+    setReferralsDataState(referralsData);
```
gdev8317, line 74: mmm thinking if it can work unexpectedly in some case.

so here you're setting "initial" value when data from subgraph is loaded
then when user adds code `referralsDataState` is updated
in theory if `referralsData` changes it can override that updated state 🤔

src/views/Referrals/Referrals.js
```diff
@@ -0,0 +1,601 @@
...
+}
+
+function ReferrersStats({ referralsDataState, infoTokens, chainId, library, setPendingTxns, pendingTxns }) {
+  let [referralCode, setReferralCode] = useState("");
+  let [isAdding, setIsAdding] = useState(false);
+  let [isAddReferralCodeModalOpen, setIsAddReferralCodeModalOpen] = useState(false);
+  let [error, setError] = useState("");
+
+  const [, copyToClipboard] = useCopyToClipboard();
+  let open = () => setIsAddReferralCodeModalOpen(true);
```
gdev8317, line 236: const? :)

src/Addresses.js
```diff
@@ -130,6 +130,7 @@ const CONTRACTS = {
     PositionRouter: "0x3D6bA331e3D9702C5e8A8d254e5d8a285F223aba",
 
     UniswapGmxEthPool: "0x80A9ae39310abf666A87C743d6ebBD0E8C42158E",
+    Referral: "0xe6fab3f0c7199b0d34d7fbe83394fc0e0d06e99d",
```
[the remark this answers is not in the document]
vipineth: Fixed!

src/App.js
```diff
@@ -312,6 +322,15 @@ function FullApp() {
   const triedEager = useEagerConnect(setActivatingConnector);
   useInactiveListener(!triedEager || !!activatingConnector);
 
+  let query = useRouteQuery();
```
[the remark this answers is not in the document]
vipineth: Sure!

src/views/Referrals/Referrals.js
gdev8317: > I was wondering if apolloClient gives a way to re-fetch the data programmatically.

we may probably use `useSWR` there to make it more flexible. or as a simple solution we may pass some `key` to use useUserReferralCode to use in `useEffect`'s dependencies

atm all subgraph data in UI is just loaded once and never updates. probably it's not critical, i don't think it makes sense  to update rebates data realtime (it doesn't change that often) 🤔

vipineth: Yeah, because of that reason only I though to just use optimistic update. Because when a user adds new referral code it's not going to have any volume. So in future when they refresh the page they will get the updated data.

xvi10 (approved):

PR #133, opened 2022-04-27
Fix TraderJOE URL

obito:
Fix the TraderJOE URL to redirect to the trade page with the token already input.

The current link just return to home.


xvi10:
will merge this, thank you

PR #134, opened 2022-04-28
Implement multi language with lingui/react

midasdev711:


src/locales/en/messages.js
```diff
@@ -0,0 +1,24 @@
...
+    "Home": "Home",
+    "Trade": "Trade",
+    "Connect": "Connect",
+    "Connect Wallet": "Connect Wallet",
+    "MetaMask not detected.": "MetaMask not detected.",
+    "Install MetaMask": "Install MetaMask",
+    ", and use GMX with its built-in browser": ", and use GMX with its built-in browser",
+    " to start using GMX": " to start using GMX",
+    "Coinbase Wallet not detected.": "Coinbase Wallet not detected.",
+    "Install Coinbase Wallet": "Install Coinbase Wallet"
```
gdev8317, line 22: Imo better to use more "key-like" keys for mappings. otherwise eventually after updating texts for several times we will end up having weird mappings that don't make sense

e.g. imagine after a while we will update text ", and use GMX with its built-in browser" to ", and gain access to trust-less trading platform". and in a mapping file it will be:
```
", and use GMX with its built-in browser": ", and gain access to trust-less trading platform"
```
it might confuse both us and people who will translate it. i've had an experience with that approach and with product in active development in can become an issue in future

i believe approach with more technical key names fits better here. e.g.
```
"instalMetamaskButton": "Install MetaMask",
"installMetamaskButtonDescription": ", and use GMX with its built-in browser"
```

also namespaces can be used here. like:
```
"nav.home": "Home"
```

package.json
```diff
@@ -8,6 +8,9 @@
     "@ethersproject/providers": "^5.0.23",
     "@ethersproject/units": "^5.0.10",
     "@headlessui/react": "dev",
+    "@lingui/core": "^3.13.3",
```
gdev8317, line 11: i haven't used that lib before. i encountered with react-i18next and react-intl, i guess they are more widely used. do you think `lingui` better fits our needs?

midasdev711: Same here.
https://github.com/gmx-io/gmx-interface/pull/130
Please check above pr. I used react-i18next.

src/locales/en/messages.js
gdev8317: afaik you've discussed that with @xvi10 
let's wait for him in case there are some arguments against

xvi10: didn't discuss the naming convention, I think your recommendation makes sense

package.json
xvi10: picked lingui cos i think the syntax is preferable, it is easier to read and edit with the default language, vs with i18n which would require checking two files to figure out the content

src/locales/en/messages.js
xvi10: @midasdev711 re-discussed with @gdev8317, let's use the text as the key for now, i thought the syntax for the translation file would be something like this though: https://github.com/Uniswap/interface/blob/main/src/locales/de-DE.po#L100

PR #138, opened 2022-05-01
add PageNotFound page

vipineth:
- [ ] Add Page Not Found page


src/App.js
```diff
@@ -802,6 +803,12 @@ function FullApp() {
             <Route exact path="/debug">
               <Debug />
             </Route>
+            <Route exact path="/debug">
+              <Debug />
+            </Route>
```
gdev8317, line 808: i guess second Router with path=/debug is not required?

PR #139, opened 2022-05-03
change font for changing values to mono

vipineth:


src/components/Exchange/ExchangeTVChart.js
```diff
@@ -440,31 +437,35 @@ export default function ExchangeTVChart(props) {
...
-					</div>
-					<div className="ExchangeChart-additional-info">
-						<div className="ExchangeChart-info-label">24h Low</div>
-						<div>
-							{!low && "-"}
-							{low && low.toFixed(2)}
-						</div>
-					</div>
+          <div>
+            <div className="ExchangeChart-main-price font-mono">
```
gkrasulya, line 441: maybe add that font to `.ExchangeChart` class in styles file? at first glance adding `font-mono` to different elements in views looks like a unnecessary duplication

gdev8317, line 441: maybe add that font to .ExchangeChart class in styles file? at first glance adding font-mono to different elements in views looks like a unnecessary duplication

PR #141, opened 2022-05-09
Referrals refactor

vipineth:
- [x] Break Referrals component into smaller components
- [x] Remove common feature into a separate component 
- [x] Lots of minor improvements


src/views/Referrals/CreateAffiliateCode.js
```diff
@@ -0,0 +1,173 @@
...
+
+  const buttonError = getButtonError();
+
+  function getPrimaryText() {
+    if (buttonError) {
+      return buttonError;
+    }
+
+    if (isProcessing) {
+      return `Creating...`;
```
gdev8317, line 102: let's use `"` quotes by default? and ` only in case we need to use variables

gdev8317: does prettier support such rule?

src/views/Referrals/CreateAffiliateCode.js
```diff
@@ -0,0 +1,173 @@
...
+    }
+    if (takenStatus === "other") {
+      setIsProcessing(false);
+    }
+
+    if (takenStatus === "none" || takenStatus === "other") {
+      const ownerOnOtherNetwork = takenInfo[chainId === ARBITRUM ? "ownerAvax" : "ownerArbitrum"];
+      try {
+        const tx = await handleCreateReferralCode(referralCode);
+        if (typeof callAfterSuccess === "function") {
```
gdev8317, line 132: `callAfterSuccess` should be either undefined or function? i think it better to just check `if (callAfterSuccess) ...`

in case parent function passed callAfterSuccess of incorrect type it's better to throw an error, and parent function code will be fixed

src/views/Referrals/CreateAffiliateCode.js
```diff
@@ -0,0 +1,173 @@
...
+  }
+
+  async function handleSubmit(event) {
+    event.preventDefault();
+    setIsProcessing(true);
+    const { status: takenStatus, info: takenInfo } = await getReferralCodeTakenStatus(account, referralCode, chainId);
+    if (takenStatus === "all" || takenStatus === "current") {
+      setIsProcessing(false);
+    }
+    if (takenStatus === "other") {
```
gdev8317, line 124: it can be rewritten like `if (["all", "current", "other"].includes(takenStatus)) ...`. no need to have 2 separate checks, right?

src/views/Referrals/JoinReferralCode.js
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    if (isSubmitting) {
+      return "Adding...";
+    }
+    if (debouncedReferralCode === "") {
+      return "Enter Referral Code";
+    }
+    if (isValidating) {
+      return `Checking code...`;
```
gdev8317, line 56: let's replace ` -> " here and couple lines below?

src/views/Referrals/CreateAffiliateCode.js
gdev8317: and inside [JoinReferralCode](https://github.com/gmx-io/gmx-interface/pull/141/files#diff-4d09bd609cef3a6437b9da1f2d741182d917e13b4c104342dd0d771a09b7ac25)

src/views/Referrals/JoinReferralCode.js
```diff
@@ -0,0 +1,151 @@
...
+      return false;
+    }
+    return true;
+  }
+
+  async function handleSubmit(event) {
+    const isEdit = type === "edit";
+    event.preventDefault();
+    setIsSubmitting(true);
+    const referralCodeHex = encodeReferralCode(referralCode);
```
gdev8317, line 81: i guess we discussed that already – i was thinking about moving all encoding/decoding logic to api/referrals so views should not know about it and handle it (in general i want to keep less logic inside our components/views, because sometimes they're becoming too large and complex).
is it difficult to change, what do you think?

src/components/Referrals/AddAffiliateCode.js
```diff
@@ -0,0 +1,170 @@
...
+    setIsProcessing(true);
+    const { status: takenStatus, info: takenInfo } = await getReferralCodeTakenStatus(account, referralCode, chainId);
+    if (["all", "current", "other"].includes(takenStatus)) {
+      setIsProcessing(false);
+    }
+
+    if (takenStatus === "none" || takenStatus === "other") {
+      const ownerOnOtherNetwork = takenInfo[chainId === ARBITRUM ? "ownerAvax" : "ownerArbitrum"];
+      try {
+        const tx = await handleCreateReferralCode(referralCode);
```
gdev8317, line 128: sorry missed that last time, can we move `handleCreateReferralCode` to Api/referrals.js?

gdev8317 (approved):
i think it's okay to merge into `staging`

gdev8317:
@vipineth if you're not going to add some more changes to this PR let's merge it

PR #143, opened 2022-05-16
Leverage trading to usd

vipineth:
---
  - https://app.asana.com/0/0/1202229483086812


gdev8317:
some first thoughts:
1. i guess the main issue is that it's unclear for me that it is a USD value inside "Long" input 
image
2. also we should limit this value to N symbols 
image

PR #144, opened 2022-05-17
Staging

xvi10:


xvi10:
merged https://github.com/gmx-io/gmx-interface/pull/128

xvi10:
merged https://github.com/gmx-io/gmx-interface/pull/142

src/Helpers.js
```diff
@@ -2701,3 +2702,7 @@ export function useDebounce(value, delay) {
   );
   return debouncedValue;
 }
+
+export function isDevelopment() {
+  return !window.location.origin?.includes("gmx.io");
```
gdev8317, line 2707: just curious why using `origin` here instead of `hostname`? both work, origin contains protocol as well

src/components/Exchange/ConfirmationBox.js
```diff
@@ -149,6 +178,10 @@ export default function ConfirmationBox(props) {
   };
 
   const getPrimaryText = () => {
+    if (decreaseOrdersThatWillBeExecuted.length > 0 && !isTriggerWarningAccepted) {
+      return `Accept confirmation of trigger orders`;
+    }
```
gdev8317, line 183: both `getPrimaryText` and `isPrimaryEnabled` use `getError()`. so it should be sufficient to add that check to `getError` func end return error text. so this works for any kind of error

PR #146, opened 2022-05-25
feat: add ecosystem projects

vipineth:


src/views/Ecosystem/Ecosystem.js
```diff
@@ -384,6 +383,38 @@ export default function Ecosystem() {
...
+                  <img src={arbitrumIcon} alt="arbitrumIcon" />
+                  <img src={avalancheIcon} alt="avalancheIcon" />
+                </div>
+              </div>
+              <div className="App-card-divider"></div>
+              <div className="App-card-content">
+                <div className="App-card-row">
+                  <div className="label">Link</div>
+                  <div>
+                    <a href="http://gmx.defisims.com/" target="_blank" rel="noopener noreferrer">
```
gdev8317, line 399: let's use url with http*s*

gdev8317: (and on next line as well)

xvi10:
i think defisims is inside the list already on master

PR #150, opened 2022-05-31
Restyle dashboard

dospore:
Restyle dashboard


gdev8317:
@dospore no worries :)

PR #151, opened 2022-06-02
fix: remove bignumber form localstorage

vipineth:


src/Api/referrals.js
```diff
@@ -47,7 +47,7 @@ export function decodeReferralCode(hexCode) {
 }
 
 export function encodeReferralCode(code) {
-  let final = code.replace(/[^\w\s_]/g, ""); // replace everything other than numbers, string  and underscor to ''
+  const final = code.replace(/[^\w\s_]/g, ""); // replace everything other than numbers, string  and underscor to ''
```
gkrasulya, line 50: @xvi10  just noticed that regexp will not replace any space characters. is that on purpose?

gdev8317, line 50: @xvi10 that regexp won't replace any space-characters as well, is it on purpose?

src/components/Referrals/referralsHelper.js
```diff
@@ -66,6 +66,21 @@ export const tierDiscountInfo = {
   2: 10,
 };
 
+export function deserializeSampleStats(input) {
+  const parsedData = JSON.parse(input);
```
gdev8317, line 70: could you check please what will happen in case there is bad value store in local storage? e.g. incorrect json. will useLocalStorage fail in this case?

we should treat localstorage as a user input in general and handle incorrect input so the app still works

gdev8317: and also handle case when there is empty object or some fields are missing. 
so the goal is to not fail in case of any data there

vipineth: Sure!

src/Api/referrals.js
```diff
@@ -342,34 +342,50 @@ export async function getReferralCodeOwner(chainId, referralCode) {
...
+      }
+    }
+    getUserReferralCode();
+  }, [account, contract, userReferralCodeInLocalStorage]);
 
   return {
     userReferralCode,
     userReferralCodeString,
-    mutateUserReferralCode,
+    userReferralCodeForExchange,
```
gdev8317, line 388: i think better to return boolean flag like "isCodeAttached" or something like this. with this flag other app components can understand if the code is already on-chain or not. and SwapBox can decide if it wants to use it inside `increasePosition` call

imo with `userReferralCodeForExchange` API lib knows too much about it's "user" – SwapBox

src/Api/referrals.js
```diff
@@ -342,34 +317,45 @@ export async function getReferralCodeOwner(chainId, referralCode) {
...
+      ],
+    { fetcher: fetcher(library, ReferralStorage) }
+  );
+  const [isCodeAttached, userReferralCode, userReferralCodeString] = useMemo(() => {
+    if (onChainCode && !isHashZero(onChainCode)) {
+      return [true, onChainCode, decodeReferralCode(onChainCode)];
+    } else if (localStorageCodeOwner && !isAddressZero(localStorageCodeOwner)) {
+      return [false, localStorageCode, decodeReferralCode(localStorageCode)];
+    }
+    return [ethers.utils.AddressZero];
```
gdev8317, line 352: i guess first element should be either true/false or it should be an empty array

src/Api/referrals.js
```diff
@@ -342,34 +317,45 @@ export async function getReferralCodeOwner(chainId, referralCode) {
...
+  }, [localStorageCode, localStorageCodeOwner, onChainCode]);
 
   return {
     userReferralCode,
     userReferralCodeString,
-    mutateUserReferralCode,
+    isCodeAttached,
```
gdev8317, line 358: maybe `savedOnChain`? i don't want to bother with naming however for new developers without knowing the context "is attached" might be unclear when they see it in some component's code, makes sense?

src/components/Exchange/SwapBox.js
```diff
@@ -176,8 +175,8 @@ export default function SwapBox(props) {
   const [isSubmitting, setIsSubmitting] = useState(false);
   const [modalError, setModalError] = useState(false);
   const [isHigherSlippageAllowed, setIsHigherSlippageAllowed] = useState(false);
-  const { userReferralCode } = useUserReferralCode(library, chainId, account);
-  const [userReferralCodeInLocalStorage] = useLocalStorage(REFERRAL_CODE_KEY);
+  const { isCodeAttached } = useUserReferralCode(library, chainId, account);
+  const userReferralCodeInLocalStorage = window.localStorage.getItem(REFERRAL_CODE_KEY);
```
gdev8317, line 179: now `getReferralCodeTakenStatus` already returns code from localstorage, i think better to use value from there

src/components/Referrals/referralsHelper.js
gdev8317: mmm i guess in case input is incorrect json this code will fail? i think we should at least wrap it in try/catch

src/hooks/useEffectDebugger.js
```diff
@@ -0,0 +1,36 @@
+import { useEffect, useRef } from "react";
```
gdev8317, line 1: btw i like the idea of storing such utility hooks in separate place, Helpers are pretty big already 👍

src/components/Referrals/referralsHelper.js
```diff
@@ -1,15 +1,16 @@
...
+  USD_DECIMALS,
   MAX_REFERRAL_CODE_LENGTH,
   ARBITRUM,
   AVALANCHE,
-  isAddressZero,
 } from "../../Helpers";
 import { encodeReferralCode, getReferralCodeOwner } from "../../Api/referrals";
 
 export const REFERRAL_CODE_REGEX = /^\w+$/; // only number, string and underscore is allowed
+export const REGEX_VERIFY_BYTES32 = new RegExp(/0x[0-9a-f]{64}/);
```
gdev8317, line 13: - i guess `new RegExp` is not needed here, it's already regexp because of leading and trailing slashes
- better to add `^` at the beginning and `$` in the end otherwise strings like "!@&#^%!@# 0xfac123.. !@#!@#" would be valid

src/Api/referrals.js
```diff
@@ -342,34 +317,45 @@ export async function getReferralCodeOwner(chainId, referralCode) {
   const referralStorageAddress = getContract(chainId, "ReferralStorage");
   const provider = getProvider(null, chainId);
   const contract = new ethers.Contract(referralStorageAddress, ReferralStorage.abi, provider);
-  const codeOwner = await contract.codeOwners(referralCode);
-  return codeOwner;
+  try {
```
gdev8317, line 320: no need to change it in this PR, but i think in future would be good to use `useSWR` for all data retrievals

for example useSWR has retries out-of-the-box. imagine in this specific case something goes wrong, request fails and it will be never requested again. and we will display in UI that code has no owner whereas it might be not true

src/components/Referrals/referralsHelper.js
gdev8317: is there a chance that in case some fields are missing the code that uses it will throw error?

src/components/Referrals/referralsHelper.js
```diff
@@ -92,14 +109,13 @@ export function getUSDValue(value, decimals = 2) {
 }
 
 export function getCodeError(value) {
-  const trimmedValue = value.trim();
-  if (!trimmedValue) return "";
+  if (!value) return "";
```
gdev8317, line 112: why do you think it's better to not use trimmed value?

src/components/Referrals/referralsHelper.js
vipineth: We are adding the data in local storage in an Array.  You can look at the updated code below. This will check the data to be an array and the different values inside array to be an object. 

```js
export function deserializeSampleStats(input) {
  const parsedData = JSON.parse(input);
  if (!Array.isArray(parsedData)) return [];
  return parsedData.map((data) => {
    if (typeof data !== "object") return {};
    return Object.keys(data).reduce((acc, cv) => {
      const currentValue = data[cv];
      if (currentValue?.type === "BigNumber") {
        acc[cv] = bigNumberify(currentValue.hex || 0);
      } else {
        acc[cv] = currentValue;
      }
      return acc;
    }, {});
  });
}
```

src/Api/referrals.js
vipineth: Yes for sure!

src/Api/referrals.js
vipineth: Yeah sure `attachedOnChain` makes sense.

src/components/Exchange/SwapBox.js
vipineth: You mean `useUserReferralCode` instead of getReferralCodeTakenStatus I think. I will update accordingly

src/components/Referrals/referralsHelper.js
vipineth: Sure! I have updated it to `/^0x[0-9a-f]{64}$/`

src/Api/referrals.js
vipineth: Sure!

vipineth: I will add a task for that and update it later.

src/components/Referrals/referralsHelper.js
vipineth: Sorry, I was thinking of spaces are included in the bytes32. I just checked spaces doesn't affect the bytes32.

I was thinking to show the error when there is any space in the input.

vipineth: I will bring that back.

src/components/Referrals/referralsHelper.js
gdev8317: i mean if i add manually incorrect value to localstorage e.g. `[{time: Date.now()}]` then page will be broken
image

localstorage data should be verified as any other user input like query params so it can't break the page. it's just some thing to keep in mind when we're using some value that user can change directly and validate it by default.
tbh i'm not 100% sure if every other handles that, but hope it does

gdev8317 (approved):

PR #157, opened 2022-06-09
Fix tooltip scroll

vipineth:


src/components/Tooltip/TooltipWithPortal.js
```diff
@@ -0,0 +1,102 @@
...
+function Portal({ children }) {
+  const root = document.body;
+  const el = document.createElement("div");
+
+  useEffect(() => {
+    root.appendChild(el);
+    return () => root.removeChild(el);
+  }, [el, root]);
+
+  return createPortal(children, el);
```
gkrasulya, line 16: interesting, could you tell what's happening here? i've never used `createPortal`

gkrasulya:
@netlify

src/components/Tooltip/TooltipWithPortal.js
vipineth: Portals is a way to push element into DOM that will be outside of the components hierarchy. The issue we were facing was no scroll on the element with overflow property. So, an extra DOM element outside of the hierarchy will fix this issue.

Here we are dynamically pushing an element suing portal because that will be outside of the hierarchy.

PR #159, opened 2022-06-12
fix glp sell cap reach disabled button

gdev8317:
Right now it's impossible to sell GLP for USDT. this is caused by validation of max cap for token. this validation should be applied for buying only


gdev8317:
123

xvi10:
I see, will close then, thank you

PR #217, opened 2022-07-22
Position share new

vipineth:
- [x] Implement position sharing using canvas


src/components/Exchange/PositionShare.js
```diff
@@ -21,6 +21,7 @@ import { useAffiliateCodes } from "../../Api/referrals";
 import SpinningLoader from "../Common/SpinningLoader";
 
 const ROOT_SHARE_URL = getRootShareApiUrl();
+// const ROOT_SHARE_URL = "https://share.gmx.io";
```
xvi10, line 24: could remove this

src/components/Exchange/PositionShare.js
```diff
@@ -115,7 +124,8 @@ function PositionShare({ setIsPositionShareModalOpen, isPositionShareModalOpen,
   );
 }
 
-function PositionShareCard({ positionRef, position, userAffiliateCode, uploadedImageInfo }) {
+function PositionShareCard({ positionRef, position, userAffiliateCode, uploadedImageInfo, uploadedImageError }) {
+  console.log({ uploadedImageError });
```
xvi10, line 128: think this should be removed as well

PR #231, opened 2022-08-07
[Unsigned]  fix asset dropdown hover

vipineth:
- [x] Fix asset dropdown hover 
- [x] Change USD to $ on SwapBox 
- [x] Add new Ecosystem project
- [x] Fix the style of tooltip on Save on Fees  
- [x] Fix the style of tooltip on GLP Index Composition

---
  - https://app.asana.com/0/0/1202835647852404


src/views/Dashboard/TooltipCard.js
```diff
@@ -7,6 +7,7 @@ export default function TooltipCard({
   arbitrum,
   showDollar = true,
   decimalsForConversion = USD_DECIMALS,
+  symbol,
```
gdev8317, line 10: i missed PR in which TooltipCard was added, i think views should not contain components that are used inside other views or other components. 

e.g. Exchange view requires it
`import { TooltipCardRow } from "../../views/Dashboard/TooltipCard";`

Can we move TooltipCard to components/Tooltip dir?

vipineth: Sure I will do that. It just stated as to be used only on the DashboardV2 later we wanted to use it everywhere, to make all the tooltip look similar.

Done!

PR #239, opened 2022-08-10
Receive token change feature

divhead:
Task: https://app.asana.com/0/1201404209663627/1201719254714513/f

Pr preview for "not enough liquidity case" https://github.com/gmx-io/gmx-interface/pull/240

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 /node_modules
 /.pnp
 .pnp.js
+.vscode
 
 # testing
 /coverage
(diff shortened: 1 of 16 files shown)


divhead:
Comments from 

> positionFee and fundingFee is denominated in USD
> swapFee is denominated in token
> swapFeeUsd is denominated in USD

> maybe to avoid confusion better to name swapFee for USD denominated and swapFeeToken (swapFeeInToken?), what do you think?

**Fixed**

> here you don't check if receiveToken is defined, however on line 675 you validate if receiveToken is defined.

> if it can be undefined we should check it here? or no need to check it on line 675?

**Fixed**

PR #240, opened 2022-08-10
[For tests only] Receive token select liquidity case

divhead:


src/components/Exchange/PositionSeller.js
```diff
@@ -74,6 +88,30 @@ function getTokenAmount(usdAmount, tokenAddress, max, infoTokens) {
...
+  const fromInfo =  getTokenInfo(infoTokens, fromTokenAddress)
+  const toInfo = getTokenInfo(infoTokens, toTokenAddress)
+
+  const maxInUsd = fromInfo.maxUsdgAmount
+    ?.sub(fromInfo.usdgAmount)
+      .mul(expandDecimals(1, USD_DECIMALS))
+      .div(expandDecimals(1, USDG_DECIMALS));
+
+  const maxIn = maxInUsd?.mul(expandDecimals(1, fromInfo.decimals)).div(fromInfo.maxPrice).toString();
+  const maxOut = toInfo.availableAmount?.gt(toInfo.poolAmount?.sub(toInfo.bufferAmount))
```
gdev8317, line 101: `gt()` would fail if you pass undefined to it. i.e. in case availableAmount != undefined, but toInfo.poolAmount is undefined it will raise error

src/components/Exchange/PositionSeller.js
```diff
@@ -114,6 +153,18 @@ export default function PositionSeller(props) {
...
   const positionRouterAddress = getContract(chainId, "PositionRouter");
   const nativeTokenSymbol = getConstant(chainId, "nativeTokenSymbol");
+  const toTokens = getTokens(chainId);
+
+  const [savedRecieveTokenAddress, setSavedRecieveTokenAddress] = useLocalStorageByChainId(
+    chainId,
+    `${CLOSE_POSITION_RECEIVE_TOKEN_KEY}-${position?.indexToken?.symbol}-${position.isLong ? 'long' : 'short'}`
+  );
+
+  const [swapToken, setSwapToken] = useState(() => (
```
gdev8317, line 163: i'm thinking what if address stored in local storage is invalid? let's consider local storage as another source of user data – not trust it

src/components/Exchange/PositionSeller.js
```diff
@@ -270,19 +340,52 @@ export default function PositionSeller(props) {
...
+
+      if (feeBasisPoints) {
+        const baseFromAmountUsd = fromAmount.mul(BASIS_POINTS_DIVISOR).div(position.leverage)
+        swapFeeUsd = baseFromAmountUsd.mul(feeBasisPoints).div(BASIS_POINTS_DIVISOR)
+        swapFee = getTokenAmount(swapFeeUsd, collateralToken.address, false, infoTokens)
+      }
+    }
+
+    totalFees = totalFees
+      .add(positionFee || bigNumberify(0))
```
gdev8317, line 377: it's okay to use simple numbers as an argument to add/sub/other BigNumber methods
jfyi

src/components/Exchange/PositionSeller.js
```diff
@@ -270,19 +340,52 @@ export default function PositionSeller(props) {
       }
     }
 
+    receiveToken = (allowReceiveTokenChange && swapToken) 
+      ? swapToken
+      : collateralToken;
+
     receiveAmount = receiveAmount.add(collateralDelta);
+    receiveAmount = receiveAmount.mul(bigNumberify(1000000));
```
gdev8317, line 348: may be better to define a constant for 1000000?

src/components/Exchange/PositionSeller.js
```diff
@@ -270,19 +340,52 @@ export default function PositionSeller(props) {
...
+        infoTokens,
+        undefined,
+        undefined,
+        usdgSupply,
+        totalTokenWeights,
+        true
+      );
+
+      if (feeBasisPoints) {
+        const baseFromAmountUsd = fromAmount.mul(BASIS_POINTS_DIVISOR).div(position.leverage)
```
gdev8317, line 370: position.leverage might fluctuate based on pnl https://github.com/gmx-io/gmx-interface/blob/master/src/views/Exchange/Exchange.js#L299

user can enable a flag "include pnl in leverage calculation" in settings. i guess in this case calculation may be incorrect?

src/components/Exchange/PositionSeller.js
```diff
@@ -557,16 +664,23 @@ export default function PositionSeller(props) {
...
+
     if (nextHasProfit && !minProfitTimeExpired && !isProfitWarningAccepted) {
       if ((position.isLong && priceLimit.lt(profitPrice)) || (!position.isLong && priceLimit.gt(profitPrice))) {
         priceLimit = profitPrice;
       }
     }
+    const path = [tokenAddress0];
 
-    const withdrawETH = collateralTokenAddress === AddressZero || collateralTokenAddress === nativeTokenAddress;
+    if ((receiveToken?.address !== tokenAddress0) && allowReceiveTokenChange) {
```
gdev8317, line 675: in case `receiveToken` is undefined that condition can be true. and on next line it will fail? because it will try to read property of undefined

src/components/Exchange/PositionSeller.js
```diff
@@ -270,19 +340,52 @@ export default function PositionSeller(props) {
...
+        const baseFromAmountUsd = fromAmount.mul(BASIS_POINTS_DIVISOR).div(position.leverage)
+        swapFeeUsd = baseFromAmountUsd.mul(feeBasisPoints).div(BASIS_POINTS_DIVISOR)
+        swapFee = getTokenAmount(swapFeeUsd, collateralToken.address, false, infoTokens)
+      }
+    }
+
+    totalFees = totalFees
+      .add(positionFee || bigNumberify(0))
+      .add(fundingFee || bigNumberify(0))
+      .add(swapFeeUsd || bigNumberify(0))
```
gdev8317, line 379: positionFee and fundingFee is denominated in USD
swapFee is denominated in token
swapFeeUsd is denominated in USD

maybe to avoid confusion better to name `swapFee` for USD denominated and `swapFeeToken` (swapFeeInToken?), what do you think?

src/components/Exchange/PositionSeller.js
```diff
@@ -557,16 +664,23 @@ export default function PositionSeller(props) {
...
     }
+    const path = [tokenAddress0];
 
-    const withdrawETH = collateralTokenAddress === AddressZero || collateralTokenAddress === nativeTokenAddress;
+    if ((receiveToken?.address !== tokenAddress0) && allowReceiveTokenChange) {
+      path.push(receiveToken.address)
+    }
+
+    const withdrawETH = path.length === 1 
+      && (receiveToken.address === nativeTokenAddress || receiveToken.address === AddressZero)
```
gdev8317, line 680: here you don't check if receiveToken is defined, however on line 675 you validate if `receiveToken` is defined. 

if it can be undefined we should check it here? or no need to check it on line 675?

src/components/Exchange/TokenSelector.js
```diff
@@ -103,8 +107,24 @@ export default function TokenSelector(props) {
...
               balanceUsd = balance.mul(info.maxPrice).div(expandDecimals(1, token.decimals));
             }
+
+            const tokenState = getTokenState(info)|| {};
+
             return (
-              <div className="TokenSelector-token-row" onClick={() => onSelectToken(token)} key={token.address}>
+              <div 
+                key={token.address}
+                className={cx('TokenSelector-token-row', {disabled: tokenState?.disabled})}
```
gdev8317, line 116: i think `?` is unnecessary here, tokenState is always defined?

src/components/Exchange/TokenSelector.js
```diff
@@ -103,8 +107,24 @@ export default function TokenSelector(props) {
...
+                onClick={() => !tokenState.disabled && onSelectToken(token)}
+              >
+                {tokenState.disabled && tokenState.message && 
+                  <Tooltip
+                    className="TokenSelector-tooltip"
+                    handle={<div style={{width: '100%', height: '100%'}} />}
```
gdev8317, line 122: and remove inline styles here?

src/components/Exchange/PositionSeller.js
divhead: it's a hardcode for tests)

src/components/Exchange/PositionSeller.js
divhead: If address is invalid in this line swapToken will be initiated as undefined, it's ok

```
const [swapToken, setSwapToken] = useState(() => (
    savedRecieveTokenAddress 
      ? toTokens.find(token => token.address === savedRecieveTokenAddress)
      : undefined
  ));
 ```

PR #244, opened 2022-08-14
[Unsigned]  Upgrade react script to v5

vipineth:
[x] Webpack 5 no longer includes node polyfills. Several dev dependencies needed to be added.
[x] There is no way to configure webpack in CRA without ejecting or using react-app-rewired.
[x] We are using `react-app-rewired`
[x] Change require to not use `default`

---
  - https://app.asana.com/0/0/1202793736284910


.env
```diff
@@ -0,0 +1 @@
+GENERATE_SOURCEMAP=false
```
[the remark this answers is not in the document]
divhead: Solved.
This does not seem to affect debugging code in dev mode

References:
https://github.com/facebook/create-react-app/pull/11752
https://github.com/facebook/create-react-app/discussions/11767

divhead (dismissed):

.env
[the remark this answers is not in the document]
vipineth: Thanks for updating!

config-overrides.js
```diff
@@ -0,0 +1,32 @@
+const webpack = require("webpack");
+const ModuleScopePlugin = require("react-dev-utils/ModuleScopePlugin");
+
+module.exports = function override(config) {
+  config.resolve.plugins = config.resolve.plugins.filter((plugin) => !(plugin instanceof ModuleScopePlugin));
+  // https://github.com/lingui/js-lingui/issues/1195
+  config.module.rules[0].oneOf.splice(2, 0, {
```
divhead, line 7: Perhaps we should add a more detailed comment on what rule is changed here and why?

Or more explicitly, redefine this like:
config.module.rules[0] = {...} ?

vipineth: https://github.com/gmx-io/gmx-interface/pull/244/commits/33b99b68f2e270434cdbc3ffbe6d854487393ab5

Refactored and added comment explaining the loader we are adding to webpack ocnfig.

.env
gdev8317, line 1: will it have effect for local development?

vipineth: No it only applies to production. You can check this link https://create-react-app.dev/docs/advanced-configuration/

PR #251, opened 2022-08-17
Wallet rpc errors

divhead:


src/Api/index.js
```diff
@@ -964,27 +964,50 @@ export function executeDecreaseOrder(chainId, library, account, index, feeReceiv
...
+
 export function extractError(ex) {
   if (!ex) {
     return [];
   }
+
   const message = ex.data?.message || ex.message;
+  const code = ex.code;
+
   if (!message) {
```
gdev8317, line 995: is it possible that `code` exists but `message` is falsie?

src/Api/index.js
```diff
@@ -964,27 +964,50 @@ export function executeDecreaseOrder(chainId, library, account, index, feeReceiv
...
+      const matchMessage = pattern.msg && message.includes(pattern.msg);
+
+      if (matchCode || matchMessage) {
+        return [message, type, ex.data];
       }
     }
   }
-  return [message];
+
+  return [message, ex.data];
```
gdev8317, line 1010: it should probably be `[message, null, ex.data];`?

PR #257, opened 2022-08-20
[Unsigned] Setup 1inch swap on capacity reached

vipineth:
---
  - https://app.asana.com/0/0/1202477937360852


src/components/Exchange/SwapBox.js
```diff
@@ -783,8 +783,15 @@ export default function SwapBox(props) {
     if (!fromTokenInfo || !fromTokenInfo.minPrice) {
       return [t`Incorrect network`];
     }
-    if (fromTokenInfo && fromTokenInfo.balance && fromAmount && fromAmount.gt(fromTokenInfo.balance)) {
-      return [t`Insufficient ${fromTokenInfo.symbol} balance`];
+
+    if (
+      !savedShouldDisableOrderValidation &&
```
gdev8317, line 788: can we use either separate setting or rename this one? not sure which option is better, we can discuss how we can handle such things in future

src/components/Exchange/SwapBox.js
```diff
@@ -1058,36 +1073,40 @@ export default function SwapBox(props) {
   };
 
   const renderErrorModal = () => {
-    const inputCurrency = fromToken.address === AddressZero ? "ETH" : fromToken.address;
+    const nativeToken = getNativeToken(chainId).symbol;
+    const inputCurrency = fromToken.address === AddressZero ? nativeToken : fromToken.address;
     let outputCurrency;
     if (isLong) {
-      outputCurrency = toToken.address === AddressZero ? "ETH" : toToken.address;
+      outputCurrency = toToken.address === AddressZero ? nativeToken : toToken.address;
```
gdev8317, line 1080: small request, can we use `nativeToken.symbol` instead of `nativeToken`? all (i guess) token variables here represent token object, so other developers might not expect it to be string. just for improved readability, what do you think?

gdev8317: or name it nativeTokenSymbol?

src/views/BuyGlp/BuyGlp.js
```diff
@@ -9,7 +9,7 @@ import "./BuyGlp.css";
 import { useChainId } from "../../Helpers";
 import { getNativeToken } from "../../data/Tokens";
 
-import { Trans } from '@lingui/macro'
+import { Trans } from "@lingui/macro";
```
gdev8317, line 12: wondering why prettier updates these files only now? we haven't changed formatting rules/styleguide, right?

PR #268, opened 2022-08-29
Add typescript

divhead:


src/react-app-env.d.ts
```diff
@@ -0,0 +1 @@
+/// <reference types="react-scripts" />
```
gdev8317, line 1: what is it?

src/typings.d.ts
```diff
@@ -0,0 +1,7 @@
+import "@web3-react";
```
gdev8317, line 1: it defines window.ethereum so TypeScript won’t fail because of it, right? and inside that file we can define other (only global?) types?

tsconfig.json
```diff
@@ -0,0 +1,32 @@
+{
+  "compilerOptions": {
```
gdev8317, line 2: config is more or less intuitive, maybe some of parameters worth commenting? or it's a usual ts config?

src/react-app-env.d.ts
divhead: This is like a re-export of global types from create-react-app. For example imported images are considered as strings (import icon from './icon.png';  icon: string)

Just generated by create-react-app automatically.

src/typings.d.ts
divhead: Yes, you can create declaration files like "something.d.ts" in the src folder, it will  add global types in the project. For example, types for global variables like window, ENV_*, Object.values, etc. 

Also you can extend or override libraries, for example 
```import {useTranslate} from 'i18n'``` could has types with translation keys from locale configs.

https://www.typescriptlang.org/docs/handbook/declaration-files/by-example.html

tsconfig.json
divhead: This is default ts config with a few changes:
1. Strict check for implicit any has been disabled (to avoid marking all untyped code with the word "any")
2. Added path alias to allow write absolute imports like "import Button from 'components/Button"

gdev8317 (dismissed):

vipineth (approved):

PR #269, opened 2022-08-29
Structure refactor

divhead:
The first step in structure changes.

Description can be found here:
https://opaque-magician-0fe.notion.site/GMX-refactor-e2d88c0e63e74813b4655b03b4e4d30a

In this PM, a base structure is created without reorganizing the code in Helpers to avoid a lot of conflicts


src/components/APRLabel/APRLabel.js
```diff
@@ -12,7 +12,7 @@ import {
   getVestingData,
   getStakingData,
   getProcessedData,
-} from "../../Helpers";
+} from "../../lib/Helpers";
```
gdev8317, line 15: let's use absolute paths for imports (e.g. `from "lib/Helpers"`)

src/components/APRLabel/APRLabel.js
```diff
@@ -22,9 +22,9 @@ import GlpManager from "../../abis/GlpManager.json";
 
 import { useWeb3React } from "@web3-react/core";
 
-import { useGmxPrice } from "../../Api";
+import { useGmxPrice } from "../../domain";
```
gdev8317, line 25: i've mentioned that in README discussion, maybe better to not use domain/index? maybe move everything from index.js to common.js and import explicitly `from "domain/common"`? not sure

gdev8317 (dismissed):

vipineth (approved):

PR #285, opened 2022-08-31
fix icon on buy gmx and update button

vipineth:
Fix icons on buy GMX page


src/components/Common/Button.js
```diff
@@ -1,29 +1,14 @@
 import cx from "classnames";
 import "./Button.css";
 
-export default function Button({ href, imgSrc, children, onClick, className, size = "lg", align = "center", ...rest }) {
+export function ButtonWithExternalLink({ href, imgSrc, children, className, size = "lg", align = "center", ...rest }) {
```
gdev8317, line 4: do you think it's necessary to rename button? i guess `Button` works as well? and it's shorted

also not sure why you removed `default`, i guess all components are exported with `default` so it's slightly shorted to import them. i guess it would be good to keep it the same for consistency?

vipineth: Yeah I just felt we are exporting multiple kinds of from this file. Like connect wallet, ButtonWithExternalLink and maybe in future we will add few more if needed.

The main reason behind renaming the Button to ButtonWithExternalLink is, earlier we were using it at multiple places like header, Buy page and other places. Now we only use at BuyGmx page.

gdev8317: thank you

if i'm not wrong according to new structure/guides each component should have separate file anyway.

gdev8317 (approved):

divhead (approved):

gdev8317:
> The base branch requires all commits to be signed

@vipineth could you take a look at it?

gdev8317 (approved):

gdev8317 (approved):

PR #288, opened 2022-08-31
show redirect modal on token card click on homepage

vipineth:


src/components/TokenCard/TokenCard.js
```diff
@@ -37,9 +37,9 @@ export default function TokenCard({ showRedirectModal }) {
   const BuyLink = ({ className, to, children, network }) => {
     if (isHome && showRedirectModal) {
       return (
-        <div className={cx("a", className)} onClick={() => showRedirectModal(to)}>
+        <HeaderLink to={to} className={cx(className)}>
```
gdev8317, line 40: i guess HeaderLink is intended to be used inside Header? 

HeaderLink was moved to separate file so it can be imported, no need to pass it as a prop
i think we can rename it (and move away from Header component) and reuse everywhere in other components

vipineth: Yes for sure. I will update based on the new refactoring. HeaderLink contains the validation for opening the Launch App modal. So, I was using it here.

divhead:
Need to replace the HeaderLink, as G mentioned, but otherwise ok

divhead (dismissed):

src/components/TokenCard/TokenCard.js
vipineth: Done!

divhead: HeaderLink now required props showRedirectModal and redirectPopupTimestamp (Previously, these variables were taken from the scope of App.js, but it was implicit dependency)

src/pages/Stake/StakeV2.js
```diff
@@ -287,7 +287,6 @@ function UnstakeModal(props) {
           <div className="Exchange-swap-section-top">
             <div className="muted">
               <div className="Exchange-swap-usd">
-                Unstake
                 <Trans>Unstake</Trans>
```
divhead, line 290: There is need to update lingui configs?

divhead (approved):

PR #289, opened 2022-09-01
Fix min pool amount

divhead:
https://app.asana.com/0/1201404209663627/1202881629558381/f

1. Added bufferAmount checks for targetToken
2. disabled max out checks for Wrap/Unwrap
3. Added collateral symbol to the modal as @xhiroz  requested

### DIFF
--- a/src/components/Exchange/PositionSeller.js
+++ b/src/components/Exchange/PositionSeller.js
@@ -87,6 +87,21 @@ function getTokenAmount(usdAmount, tokenAddress, max, infoTokens) {
   return usdAmount.mul(expandDecimals(1, info.decimals)).div(max ? info.minPrice : info.maxPrice);
 }
 
+function shouldSwap(collateralToken, receiveToken) {
+  // If position collateral is WETH in contract, then position.collateralToken is { symbol: “ETH”, isNative: true, … }
+  // @see https://github.com/gmx-io/gmx-interface/blob/master/src/pages/Exchange/Exchange.js#L162
+  // meaning if collateralToken.isNative === true in reality position has WETH as a collateral
+  // and if collateralToken.isNative === true and receiveToken.isNative === true then position’s WETH will be unwrapped and user will receive native ETH
+  const isCollateralWrapped = collateralToken.isNative;
+
+  const isSameToken =
+    collateralToken.address === receiveToken.address || (isCollateralWrapped && receiveToken.isWrapped);
+
+  const isUnwrap = isCollateralWrapped && receiveToken.isNative;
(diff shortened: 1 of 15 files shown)


src/components/Exchange/PositionSeller.js
```diff
@@ -396,11 +393,25 @@ export default function PositionSeller(props) {
...
 
+    convertedReceiveAmount = getTokenAmount(receiveAmount, receiveToken.address, false, infoTokens);
+
+    // Check swap limits (max in / max out)
+    if (allowReceiveTokenChange && swapToken && swapToken.address !== collateralToken.address) {
       const collateralInfo = getTokenInfo(infoTokens, collateralToken.address);
+      const swapTokenInfo = getTokenInfo(infoTokens, swapToken.address);
+
+      const isWrapOrUnwrap =
+        (collateralToken.isNative && swapTokenInfo.isWrapped) || (collateralToken.isWrapped && swapTokenInfo.isNative);
```
gdev8317, line 406: 1. inside contract collateralToken can be one of WBTC, WETH, LINK, UNI (OR WAVAX, WBTC, WETH on Avalanche)
if on contract size position.collateraToken == WETH then in UI position.collateralToken will be `{ isNative: true, address: AddressZero, ... }` (https://github.com/gmx-io/gmx-interface/blob/bc8252b035cf0959f97e61afcad4059b161f1ad0/src/views/Exchange/Exchange.js#L162)
So collateralToken.isWrapper is always `false` (or more precisely is `undefined`)

2. The only possible (from contracts perspective) "wrap-unwrap" operation for position with collateralToken == WETH is to unwrap WETH and send native ETH to user. I'm thinking what is correct check on UI side. at first glance it's `isUnwrap = collateralToken.isNative  && swapTokenInfo.isNative` 🤔. Does `isWrapOrUnwrap` work for you now?

src/components/Exchange/PositionSeller.js
```diff
@@ -1034,7 +1053,7 @@ export default function PositionSeller(props) {
             </div>
             <div className="Exchange-info-row">
               <div className="Exchange-info-label">
-                <Trans>Collateral</Trans>
+                <Trans>Collateral</Trans> ({collateralToken.symbol})
```
gdev8317, line 1056: let's move `({collateralToken.symbol})` inside <Trans></Trans>. e.g. in theory in some languages token symbol should be at the beginning or in the middle of the phrase

gdev8317 (approved):

gdev8317 (approved):

PR #291, opened 2022-09-01
Update lingui config

vipineth:
- Add few properties to `.linguirc`
- Add all `messages.js` (compiled file) to `.gitignore`
- Add pseudolocal to test

### DIFF
--- a/.linguirc
+++ b/.linguirc
@@ -6,7 +6,8 @@
     "ru",
     "ja",
     "en",
-    "fr"
+    "fr",
+    "pseudo"
   ],
   "sourceLocale": "en",
   "catalogs": [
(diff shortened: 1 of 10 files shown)


divhead:
Will we add a pre-commit hook here? Or is there something else we need to discuss?

.linguirc
```diff
@@ -17,5 +18,13 @@
...
-  "format": "po"
+  "formatOptions": {
+    "lineNumbers": false
+  },
+  "fallbackLocales": {
+    "default": "en"
+  },
+  "format": "po",
+  "orderBy": "messageId",
+  "pseudoLocale": "pseudo"
```
gdev8317, line 29: could you tell what is pseudo?

vipineth: Pseudo is basically a way to text if the translation is implemented properly or not when not even a single translation is done. I was using this to test when the translation was not done. 

https://lingui.js.org/guides/pseudolocalization.html

PR #293, opened 2022-09-02
add glp swap error modal

vipineth:
Restore the error modal for glp swap.

---
  - https://app.asana.com/0/0/1202881629558395


src/components/Glp/GlpSwap.js
```diff
@@ -432,7 +441,13 @@ export default function GlpSwap(props) {
 
     if (isBuying) {
       const swapTokenInfo = getTokenInfo(infoTokens, swapTokenAddress);
-      if (swapTokenInfo && swapTokenInfo.balance && swapAmount && swapAmount.gt(swapTokenInfo.balance)) {
+      if (
+        !savedShouldDisableOrderValidation &&
```
gdev8317, line 445: order validation in glp swap?

gdev8317: if i understand correctly, you're using this setting for testing.

i think it's better to create separate setting for this, otherwise it might confuse others who will work with that code

vipineth: We are using this setting for testing and removing the available balance check. It only works on the development. In production the option will not even be visible. It also helps xhiroz to test the app properly. Otherwise extreme conditions are hard to test.

We have used this at multiple place in SwapBox. Maybe we can improve the name to say something like test, dev etc. Please let me know your thoughts.

src/App.js
```diff
@@ -854,6 +854,7 @@ function FullApp() {
                   savedSlippageAmount={savedSlippageAmount}
                   setPendingTxns={setPendingTxns}
                   connectWallet={connectWallet}
+                  savedShouldDisableOrderValidation={savedShouldDisableOrderValidation}
```
divhead, line 857: About naming - should the GlpSwap component know that this prop is "saved" somewhere? Maybe just shouldDisableOrderValidation?

src/components/Glp/GlpSwap.js
```diff
@@ -464,6 +479,23 @@ export default function GlpSwap(props) {
     return [false];
   };
 
+  const renderErrorModal = useCallback(() => {
+    const inputCurrency = swapToken.address;
+    const oneInchUrl = `https://app.1inch.io/#/${chainId}/swap/${inputCurrency}`;
+    const label = `${swapToken.symbol} Capacity Reached`;
+    return (
+      <Modal isVisible={!!modalError} setIsVisible={setModalError} label={label} className="Error-modal">
```
divhead, line 487: Boolean(modalError) is more clear than !!modalError

divhead: Regarding the new structure - could you please move this component to a new file? 

Maybe I didn't quite understand something, but now it looks like a regular component, why is it written as a callback?

src/App.js
vipineth: The value is prefixed with `saved` because it's saved in `localStorage`. This is a common patter that has been used throughout the whole app/

src/components/Glp/GlpSwap.js
gdev8317: yup, i understand how it works. i mean it's name may be confusing in this context. initially it was added to disable orders validation and in UI in settings it's "Disable order validations". and i guess in master it is used for order price validation only? 

i'm not sure if it makes sense to create new flags/checkboxes for each new case, so let's just rename it (and change text in UI). maybe something like "savedShouldDisableValidationForTesting" or something like this? do you have better name in mind?

src/components/Glp/GlpSwap.js
vipineth: `Boolean()` is more explicit and makes sense. I will update that.

That's correct `useCallback` does not make sense here. Removed
I feel it makes sense to leave it where it is because it's small and very specific. Moving as a new component will make it look more complicated I feel. But please let me know your thoughts.

src/components/Glp/GlpSwap.js
vipineth: The name makes sense! I will update to `savedShouldDisableValidationForTesting`

gdev8317 (approved):

src/components/Glp/GlpSwap.js
vipineth: Makes sense. I will updated it

PR #295, opened 2022-09-06
fix capacity display

xvi10:


package.json
```diff
@@ -49,7 +49,7 @@
     "react-use": "17.3.1",
     "recharts": "2.1.8",
     "swr": "0.4.2",
-    "typescript": "4.7.4",
+    "typescript": "^4.8.2",
```
divhead, line 52: Sorry,  are we bringing up a version of the timescript here?
Also, i think a specific version would be better as the other packages

PR #297, opened 2022-09-08
fix icon on buy glp and remove link to gmx.io

vipineth:
- Show logo on BuyGMX page
- Remove link to gmx.io from header
- Update ecosystem user profile
- Show redirect modal when clicked on on `Launch Exchange` on homepage

---
  - https://app.asana.com/0/0/1202898414999789
  - https://app.asana.com/0/0/1202881629558472
  - https://app.asana.com/0/0/1202881629558486


src/components/Common/Button.js
```diff
@@ -1,29 +1,13 @@
 import cx from "classnames";
 import "./Button.css";
 
-export default function Button({ href, imgSrc, children, onClick, className, size = "lg", align = "center", ...rest }) {
+export default function Button({ href, imgSrc, children, className, size = "lg", align = "center", ...rest }) {
```
divhead, line 4: mm, why do we turn a button into a link, without the onClick Prop?

vipineth: Updated the component to `ButtonLink` and also converted it to `tsx`. Please have a look and let me know your thougths.

src/components/Common/ConnectWalletButton.tsx
```diff
@@ -0,0 +1,20 @@
+import { ReactNode } from "react";
```
gdev8317, line 1: TypeScript :)

gdev8317 (approved):
one thing i've found: on the home page logo opens redirect popup

vipineth:
> one thing i've found: on the home page logo opens redirect popup

Converted to normal link so, it will not display the redirect popup.

src/components/Common/Button.js
gdev8317: > mm, why do we turn a button into a link, without the onClick Prop?

so far Button component was used for "link-like" buttons only. e.g. on /earn page it's just `<button className="App-button-option App-card-option" onClick={} />`

i think it's better to have one universal Button component that accepts either url or onClick. and it can be reused everywhere. so maybe keep name "Button" and in some future we will extend its functionality and use it in components like StakeV2

makes sense?

vipineth: Yeah it does. Updated Button component to support both as of now will change or update it in future based on our need.

PR #301, opened 2022-09-09
update lingui config and add pre-commit

vipineth:
- Update .linguirc to not tract line number in `.po` files to remove unnecessary conflicts
- Add all messages.js (compiled file) to .gitignore
- Add lingui extract  to `pre-commit` hook

---
  - https://app.asana.com/0/0/1202948224553666


divhead (dismissed):

gkrasulya:
thanks for good description 🙂 it helps

gdev8317:
thanks for good description :) it helps

package.json
```diff
@@ -100,7 +101,7 @@
     "prettier": "2.5.1"
   },
   "lint-staged": {
-    "**/*.js?(x)": [
+    "**/*.{js, jsx, ts, tsx}": [
```
gdev8317, line 104: does it work okay with spaces? afaik native sh/bash won't allow it. probably it's different syntax here, just wondering

gdev8317 (dismissed):

package.json
vipineth: It makes sense to up remove the spaces. I will update to `src/**/*.{js,ts,jsx,tsx}`

PR #302, opened 2022-09-11
Show Buy GLP error modal

vipineth:
- Add buy GLP error modal

---
  - https://app.asana.com/0/0/1202881629558395


src/components/Glp/SwapErrorModal.tsx
```diff
@@ -0,0 +1,75 @@
...
+  useEffect(() => {
+    setLowestFeeToken(
+      getLowestFeeTokenForBuyGlp(
+        chainId,
+        glpAmount,
+        glpPrice,
+        usdgSupply,
+        totalTokenWeights,
+        infoTokens,
+        swapToken?.address
```
gdev8317, line 50: here we check if swapToken is defined, however on line 53 it's just `swapToken.address`

i think better:
- if swapToken can be undefined then it should be handled in dependency array as well (and i guess in Props?)
- if it can no be undefined – then just remove `?` to make it consistent

src/config/Tokens.js
```diff
@@ -432,3 +432,9 @@ export function getTokenBySymbol(chainId, symbol) {
 export function getWhitelistedTokens(chainId) {
   return TOKENS[chainId].filter((token) => token.symbol !== "USDG");
 }
+
+export function getVisibleTokens(chainId) {
+  return getWhitelistedTokens(chainId)
+    .filter((t) => !t.isWrapped)
```
gdev8317, line 438: let's check both conditions inside single callback? in this specific case array is small and it won't affect performance much of course, but just a good thing to avoid double iteration over array imo

src/lib/legacy.js
```diff
@@ -697,6 +697,39 @@ export function getBuyGlpFromAmount(toAmount, fromTokenAddress, infoTokens, glpP
   return { amount: fromAmount, feeBasisPoints };
 }
 
+export function getLowestFeeTokenForBuyGlp(
```
gdev8317, line 700: i guess it's better to avoid adding anything new to legacy.js?

src/lib/legacy.js
```diff
@@ -697,6 +697,39 @@ export function getBuyGlpFromAmount(toAmount, fromTokenAddress, infoTokens, glpP
...
+        MINT_BURN_FEE_BASIS_POINTS,
+        TAX_BASIS_POINTS,
+        true,
+        usdgSupply,
+        totalTokenWeights
+      );
+      return { ...token, fees };
+    })
+    .filter((t) => t.address !== fromTokenAddress)
+    .filter((t) => t.fees)
```
gdev8317, line 729: let's change it to single filter call as well?

src/components/Glp/SwapErrorModal.tsx
vipineth: Done!

src/config/Tokens.js
vipineth: Done!

src/lib/legacy.js
vipineth: Done!

src/lib/legacy.js
vipineth: Done!

src/config/Tokens.js
```diff
@@ -434,7 +434,10 @@ export function getWhitelistedTokens(chainId) {
 }
 
 export function getVisibleTokens(chainId) {
-  return getWhitelistedTokens(chainId)
-    .filter((t) => !t.isWrapped)
-    .filter((t) => !t.isTempHidden);
+  return getWhitelistedTokens(chainId).filter((token) => {
+    if (token.isWrapped || token.isTempHidden) {
+      return false;
```
gdev8317, line 439: probably a matter of taste, maybe juts `(token) => !token.isWrapped && !token.isTempHidden`?

src/config/Tokens.js
```diff
@@ -432,3 +432,12 @@ export function getTokenBySymbol(chainId, symbol) {
 export function getWhitelistedTokens(chainId) {
   return TOKENS[chainId].filter((token) => token.symbol !== "USDG");
 }
+
+export function getVisibleTokens(chainId) {
+  return getWhitelistedTokens(chainId).filter((token) => {
+    if (token.isWrapped || token.isTempHidden) {
```
gdev8317, line 438: probably matter of taste, maybe better `(token) => !token.isWrapper && !token.isTempHidden`? i think it's a bit easier to grasp the filter logic (at least for me)

src/components/Glp/SwapErrorModal.tsx
```diff
@@ -0,0 +1,75 @@
...
+      glpPrice,
+      usdgSupply,
+      totalTokenWeights,
+      infoTokens,
+      swapToken.address
+    );
+    setLowestFeeToken(lowestFeeToken);
+  }, [chainId, glpAmount, glpPrice, usdgSupply, totalTokenWeights, infoTokens, swapToken.address]);
+
+  if (!lowestFeeToken || !swapToken) return "";
```
gdev8317, line 55: not sure if it makes any different, i usually do `return null;` to not render component. maybe we can use consistent `return null` everywhere in code?

vipineth: Sure!

gdev8317 (approved):

gdev8317 (dismissed):

src/components/Glp/SwapErrorModal.tsx
```diff
@@ -0,0 +1,92 @@
...
+  chainId: number;
+  glpAmount: BigNumber;
+  usdgSupply: BigNumber;
+  totalTokenWeights: BigNumber;
+  glpPrice: BigNumber;
+  swapUsdMin: BigNumber;
+  infoTokens: InfoTokens;
+};
+
+interface LowestFeeToken extends Token {
```
divhead, line 25: 1. It's better to use this style:  `type LowestFeeToken = Token & {...additionalProps}` just for a more minimalistic and consistent codestyle
2. Looks like this type should be defined in /domain, somewhere near getLowestFeeTokenForBuyGlp function
3. I thinking, is it really necessary to define a new type of token and call it so specialized (LowestFeeToken)? 

There is a risk that this way we will have a lot of inherited token types with different field sets, which will confuse

It seems to me that here we can do something like this:

```
const {token: Token, fees: number, amountLeftToDeposit: BigNumber} = getLowestFeeTokenForBuyGlp(...)

```

Or

```
const {token: Token, feesSpects: {...}} = getLowestFeeTokenForBuyGlp(.)
```

src/domain/tokens/types.ts
```diff
@@ -4,46 +4,68 @@ export type Token = {
...
-  isStable?: boolean;
-  isNative?: boolean;
-  isShortable?: boolean;
+  isStable: boolean;
   name: string;
   symbol: string;
+  isTempHidden?: boolean;
+  isWrapped?: boolean;
+  baseSymbol?: string;
 };
```
divhead, line 13: Could you add a line break here? :)

divhead (dismissed):

src/domain/tokens/types.ts
gdev8317: i guess better to move such rules to prettier config or at least formalize it somehow? otherwise it's very subjective

i think it's good to add some newlines for readability, but unless we don't have such rule better to not block PRs with it imo

gdev8317 (approved):

PR #303, opened 2022-09-11
Added styling to disabled button on Swap Box

lucasregueiroo:
Added Styling so when the button isn't clickable inside the Swap Box on the Trade page


vipineth:
We are closing this PR. We already have the style for the disabled state of button.

PR #305, opened 2022-09-13
Add 1inch on swap

vipineth:
- Show error modal for no liquidity on SwapBox


src/components/Exchange/NoLiquidityErrorModal.tsx
```diff
@@ -0,0 +1,67 @@
...
+        below to do so:
+      </div>
+      <br />
+      {isShort && (
+        <div>
+          Alternatively you can select a different "Collateral In" token.
+          <br />
+          <br />
+        </div>
+      )}
```
gdev8317, line 60: let's move this line "Alternatively you can select a different "Collateral In" token." to the end

gdev8317 (approved):

PR #306, opened 2022-09-13
make all tooltip follow the same UI pattern

vipineth:
Signed commits from #231 

- [x] Change USD to $ on SwapBox 
- [x] Make all the tooltips follow similar style 
- [x] Add new Ecosystem project
---
  - https://app.asana.com/0/0/1202835647852404


src/components/Tooltip/TooltipCard.js
```diff
@@ -0,0 +1,67 @@
...
+  );
+}
+
+export function TooltipCardRow({ label, value, showDollar = true, values }) {
+  function renderValue() {
+    if (values && Array.isArray(values)) {
+      return (
+        <ul className="Tooltip-row-values">
+          {values.map((v) => (
+            <li key={v}>{v}</li>
```
gdev8317, line 49: let's use index for key? value can be big, not sure if it's okay

src/components/Tooltip/TooltipCard.js
```diff
@@ -0,0 +1,67 @@
...
+          {showDollar && "$"}
+          {formatAmount(total, decimalsForConversion, 0, true)}
+          {!showDollar && symbol && " " + symbol}
+        </span>
+      </p>
+    </>
+  );
+}
+
+export function TooltipCardRow({ label, value, showDollar = true, values }) {
```
gdev8317, line 43: not sure, maybe better to use one `value` prop and check if it's an array?

2 separate props `value` and `values` might be confusing, what do you think?

src/pages/Stake/StakeV2.js
```diff
@@ -1515,42 +1535,52 @@ export default function StakeV2({ setPendingTxns, connectWallet }) {
...
-                              </div>
+                              <TooltipCardRow
+                                label={`${nativeTokenSymbol} Total APR`}
+                                showDollar={false}
+                                value={`${formatKeyAmount(
+                                  processedData,
+                                  "gmxAprForNativeTokenWithBoost",
+                                  2,
+                                  2,
+                                  true
```
gdev8317, line 1574: i feel like sometimes prettier doesn't help with readability

src/components/Tooltip/TooltipCard.js
```diff
@@ -0,0 +1,67 @@
+import { formatAmount, USD_DECIMALS } from "../../lib/legacy";
+
+export default function TooltipCard({
```
divhead, line 3: could you please make it .tsx and add type Props?

src/components/Tooltip/TooltipCard.js
```diff
@@ -0,0 +1,67 @@
+import { formatAmount, USD_DECIMALS } from "../../lib/legacy";
+
```
divhead, line 2: 1. Hm, I think the name Tooltip card is too abstract here, i mean it sounds like a general tooltip for any content like onboarding, errors, etc. But, in fact this tooltip contains a strict structure and is designed only to show some statistics.

Let's rename it please.

2. I'm also a little confused by the fact that we have a components/Tooltip/TooltipCard structure that contains a very specialized TooltipCard that is exported by default, and a TooltipCardRow that is also used in different places

If I were doing this, I would name the folder something more specific, like components/PoolStatsTooltip/PoolStatsTooltip.tsx (not sure, you can probably think of a better one)

3. It seems to me that it is better not to mix different kinds of imports unnecessarily. I personally would do everything as an export function, but it's up to you.)

vipineth: Makes sense! Refactored

src/components/Tooltip/TooltipCard.js
```diff
@@ -0,0 +1,67 @@
...
+export default function TooltipCard({
+  title,
+  total,
+  avax,
+  arbitrum,
```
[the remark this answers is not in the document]
vipineth: Done

gdev8317 (approved):

gdev8317 (approved):

PR #313, opened 2022-09-17
changed referral link so that trader attribution works

jonomnom:
image

Previously this link goes to https://gmx.io/#/?ref=jonomnom . When the referred trader tries to trade on the app (app.gmx.io), they will lose their affiliation with the affiliate. Going directly to http://app.gmx.io/#/?ref=jonomnom lets the affiliation stay with the trader (does not need to do extra steps) and the trader does not have to sign a transaction.

The potential downsides with the proposed changes are:
 -  the referral link does not go to the landing page
 - using localstorage to save the referral link is temporary

### DIFF
--- a/src/components/Referrals/AffiliatesStats.js
+++ b/src/components/Referrals/AffiliatesStats.js
@@ -145,7 +145,9 @@ function AffiliatesStats({
                           <span className="referral-text ">{stat.referralCode}</span>
                           <div
                             onClick={() => {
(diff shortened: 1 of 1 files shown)


vipineth:
Thank you so much @jonomnom, for the PR 🫡

Because gmx.io is much easier for new users to comprehend, we have "https://gmx.io/#/?ref=jonomnom" instead of "https://app.gmx.io/#/?ref=jonomnom".

I am closing this PR

PR #315, opened 2022-09-20
Refactor step three

divhead:
https://app.asana.com/0/1201404209663627/1203006564569510/f

**Changes:**
1. Moved and typed infrastructure code for data access from legacy files
   1.1 subgraph -> `lib/subgraph` + `config/subgraph.ts` (urls)
   1.2 wallets -> `lib/wallets`
   1.3 localStorage -> `lib/localStorage`+ `config/localStorage.ts` (keys constants)
   1.4 backend -> `config/backend.ts` (urls)
   1.5 rpc providers -> `lib/rpc`

2. Moved widely used utils and constants
  2.1 All about chains (chainIds, metadata, etc) -> `config/chains.ts`
  2.2 Utilities for working with numbers (formatting, parsing, bigNumberify) -> `lib/numbers.ts`

3. Completed structure and types for `domain/tokens` including utils and requests

4. Added re-exports for lib and domain modules so that the rest of the code does not depend on the folder structure of the modules

5. Typescript types for all config files

6. Added ts-check in pre-commit hook

7. Moved locales to src after discussion with G

8. Added SCSS support


src/config/tokens.ts
```diff
@@ -391,48 +552,3 @@ for (const chainId of CHAIN_IDS) {
...
-
-export function getNativeToken(chainId) {
-  return NATIVE_TOKENS_MAP[chainId];
-}
-
-export function getTokens(chainId) {
-  return TOKENS[chainId];
-}
-
-export function isValidToken(chainId, address) {
```
gdev8317, line 407: does it make sense to keep these getters here in config?

src/domain/tokens/utils.ts
```diff
@@ -1,18 +1,97 @@
...
+
+export function getWhitelistedTokens(chainId: number) {
+  return TOKENS[chainId].filter((token) => token.symbol !== "USDG");
+}
+
+export function getVisibleTokens(chainId: number) {
+  return getWhitelistedTokens(chainId).filter((token) => !token.isWrapped && !token.isTempHidden);
+}
+
+export function getUsd(
```
gdev8317, line 73: not sure if `getUsd` should be inside tokens utils. logically it related to not tokens only?

src/config/subgraph.ts
```diff
@@ -0,0 +1,11 @@
+export const SUBGRAPH_URLS = {
+  chainLink: "https://api.thegraph.com/subgraphs/name/deividask/chainlink",
+
+  arbitrumStats: "https://api.thegraph.com/subgraphs/name/gmx-io/gmx-stats",
+  avalancheStats: "https://api.thegraph.com/subgraphs/name/gmx-io/gmx-avalanche-stats",
```
gdev8317, line 5: i think it make sense to split these values by chainId here (in case it's related to specific chain)? e.g.
```
{
  [ARBITRUM]: {
    stats: "https://api.thegraph.com/subgraphs/name/gmx-io/gmx-stats",
    referrals: ""https://api.thegraph.com/subgraphs/name/gmx-io/gmx-arbitrum-referrals",
    // ...
  },
  [AVALANCHE]: { 
    // ...
  },
  // ...
}
```

if there are a lot of places to update we can do it next time

src/domain/tokens/useInfoTokens.ts
```diff
@@ -0,0 +1,219 @@
...
+import { Token, TokenInfo } from "./types";
+import { BigNumber } from "ethers";
+import { Web3ReactContextInterface } from "@web3-react/core/dist/types";
+import { bigNumberify, expandDecimals } from "../../lib/numbers";
+
+export function useInfoTokens(
+  library: Web3ReactContextInterface["library"],
+  chainId: number,
+  active: boolean,
+  tokenBalances: BigNumber[] | undefined,
```
gdev8317, line 23: oh it's not the same is `tokenBalances?: BigNumber[]` ?

package.json
```diff
@@ -104,6 +105,9 @@
     "src/**/*.{js,ts,jsx,tsx}": [
       "eslint --fix ./src",
       "prettier --write"
+    ],
+    "src/**/*.{ts,tsx}": [
+      "tsc-files --noEmit src/react-app-env.d.ts src/typings.d.ts"
```
gdev8317, line 110: mmm how does that work? it runs check for 2 files only?

src/domain/tokens/utils.ts
divhead: It depends on tokenAddress and InfoTokens, uses token structure inside, looks like this is the function that applies specifically to tokens :)

I don't really like that it depends on orderOption as well, as it could be made more abstract so that it only depends on triggerPrice. But overall it looks like a function from the token domain. 

Also, in practice I think it's convenient to have functions like getTokenAmountFromUsd and getUsd (from token amount) in one place

src/domain/tokens/useInfoTokens.ts
divhead: Hm, in this place it's true, fixed it

but in general it is not quite the same thing: if you define a parameter as `tokenBalances?`
 it will be considered optional, and in that case you cannot specify the following parameters as mandatory as in the screenshot.

tokenBalances: BigNumber[] | undefined allows to explicitly specify undefined instead of parameter and types are not broken 
 
 Screenshot 2022-09-29 at 22 13 29

src/config/subgraph.ts
divhead: This is not used directly, getting the right by chainId is done in lib/subgraph/clients, also there is additional subgraphs that do not depend on chainId (nissohVault, chainLink)

I think that It's unnecessary right now Because it would complicate the syntax: 
`createClient(SUBGRAPH_URLS[ARBITRUM].stats)`

Screenshot 2022-09-29 at 22 41 22

package.json
divhead: No :) No) lint-staged passes all changed files to this command, the first two are global type declarations, which must be included in every run

After your comment I realized that lint-staged is not suitable because it only checks changed files and may miss errors in dependent files that have not been changed, I added checking the entire project in each commit, this is longer but more correct

src/domain/tokens/useInfoTokens.ts
```diff
@@ -0,0 +1,219 @@
...
+  fundingRateInfo: BigNumber[] | undefined,
+  vaultPropsLength: number | undefined
+) {
```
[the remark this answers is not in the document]
divhead: Added, and, I realized that if look in detail, in TokenInfo almost all fields are optional, because now the code is written so that you can get TokenInfo without waiting for the request and then there will only be fields from the config

src/config/Fees/index.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  for (let i = 0; i < data.length; i++) {
+    const item = data[i];
+    list.push({
+      from: item.to - SECONDS_PER_WEEK,
+      to: item.to,
+      feeUsd: item.feeUsd,
+    });
+  }
+  return list;
```
vipineth, line 24: Do you think `.map` will be a better way to write this? Maybe `.map` will make it more readable.

vipineth (approved):

src/domain/tokens/utils.ts
gdev8317: > I don't really like that it depends on orderOption as well

agree on that. if you think it makes sense it in this PR – let's do it. otherwise we can do it later

gdev8317 (approved):

gkrasulya (approved):

gdev8317 (approved):

gdev8317 (approved):

PR #318, opened 2022-09-25
fix download on iOS and mobile devices

vipineth:
[x] Fix position sharing download on iOS devices (safari)
[x] Fix background not rendering while uploading the image to cloudinary
[x] Update the `html-to-image` package
[x] Fix share link when user referral is not present

---
  - https://app.asana.com/0/0/1203041786811698
  - https://app.asana.com/0/0/1202881629558318


src/domain/downloadImage.ts
```diff
@@ -0,0 +1,33 @@
+export default async function downloadImage(dataURI: string, filename: string) {
```
divhead, line 1: Hi! downloadImage seems to be in the lib, not in the domain

More about the differences:
1. lib is abstract utilities and infrastructure code, not directly related to GMX, this function fits here

2) Domain - product/business logic, i.e. processing specific user scenarios in GMX context (there should be code about tokens, positions, orders, various calculations etc.)

vipineth: Sure, done!

src/components/Exchange/PositionShare.js
```diff
@@ -53,7 +53,11 @@ function PositionShare({ setIsPositionShareModalOpen, isPositionShareModalOpen,
     (async function () {
       const element = positionRef.current;
       if (element && userAffiliateCode.success && sharePositionBgImg && positionToShare) {
-        const image = await toJpeg(element, config);
+        // We have to call the toJpeg function multiple times to make sure the canvas renders all the elements like background image
+        // @refer https://github.com/tsayen/dom-to-image/issues/343#issuecomment-652831863
+        const image = await toJpeg(element, config)
+          .then(async () => await toJpeg(element, config))
```
divhead, line 59: it's nesessary to add async / await here?

if toJpeg returns a promise, you can just write:
.then(() => toJpeg())
.then(() => toJpeg())

vipineth: I tried with this but the canvas doesn't load the background image on iOS (safari). It works okay with download but with uploading the image it doesn't. With async await it does. I will try and find a better way going forward.

divhead (dismissed):

gdev8317 (dismissed):

gdev8317 (approved):

PR #321, opened 2022-09-29
Update translation text

vipineth:
[x] Update translation of all components
[x] Create component for external link and replace every `a` tag with `target="_blank"`
[x] Add a test language for development

---
  - https://app.asana.com/0/0/1202881629558465


src/components/Common/ExternalLink.tsx
```diff
@@ -0,0 +1,17 @@
+import React from "react";
```
divhead, line 1: Maybe move it to components/ExternalLink/ExternalLink.tsx?

I think the 'Common' folder is a bit redundant and it is better to try to keep the components flat, unless it is a case where there is a very complex component with internal child components.

src/components/Common/SEO.js
```diff
@@ -1,12 +1,11 @@
 import { Helmet } from "react-helmet";
 
-import { t } from "@lingui/macro";
-
 function SEO(props) {
   const { children, ...customMeta } = props;
   const meta = {
-    title: t`GMX | Decentralized Perpetual Exchange`,
-    description: t`Trade spot or perpetual BTC, ETH, AVAX and other top cryptocurrencies with up to 30x leverage directly from your wallet on Arbitrum and Avalanche.`,
```
divhead, line 9: hm, sorry, it's ok that we remove `t` from here? Then the title and description will be in English only?

src/components/Exchange/PositionSeller.js
```diff
@@ -725,9 +727,11 @@ export default function PositionSeller(props) {
...
     ];
 
-    const successMsg = t`Requested decrease of ${position.indexToken.symbol} ${
-      position.isLong ? "Long" : "Short"
-    } by ${formatAmount(sizeDelta, USD_DECIMALS, 2)} USD.`;
+    const successMsg = t`Requested decrease of ${position.indexToken.symbol} ${longOrShortText} by ${formatAmount(
+      sizeDelta,
+      USD_DECIMALS,
+      2
+    )} USD.`;
```
divhead, line 734: Should we add a non breaking space here? 
And, in other places like {amount} {symbol} when it can cause unexpected line breaks

divhead: Not critial just if you know some 'danger' places

divhead (approved):

src/components/Exchange/ConfirmationBox.js
```diff
@@ -378,16 +318,18 @@ export default function ConfirmationBox(props) {
       4,
       true
     );
+    const longOrShortText = select(existingOrder.isLong, { true: "Long", false: "Short" });
```
gdev8317, line 321: wondering why we need `select` in this case? is it the same?

```
const longOrShortText = existingOrder.isLong ? t`Long` : t`Short`?
```

just looks a bit simpler imo

gdev8317: like here https://github.com/gmx-io/gmx-interface/pull/321/files#diff-d8826fb03229e798252f7c2101f6fbae00a01996892c9f7337d8eb147465b1fbR332

gdev8317: image

src/pages/Exchange/Exchange.js
```diff
@@ -801,9 +816,7 @@ export const Exchange = forwardRef((props, ref) => {
         type="button"
         onClick={onMultipleCancelClick}
       >
-        <Trans>
-          Cancel {cancelOrderIdList.length} {orderText}
-        </Trans>
+        <Plural value={cancelOrderIdList.length} one="Cancel # order" other="Cancel # orders" />
```
gdev8317, line 819: for one it will be "Cancel 1 order" or "Cancel order"? I guess the last is preferable

src/components/Exchange/ConfirmationBox.js
vipineth: Makes sense👍 I was using `select` when we can have more than 2 options. Here it will never have more than 2 options.

src/pages/Exchange/Exchange.js
vipineth: Fixed!

gdev8317 (approved):

PR #328, opened 2022-10-06
share referral code link based on the root url

vipineth:
[x] Add root url of the domain when we copy the sharable link for referral code

The idea is to make the interface being deployed to any URL in future and not depend on just app.gmx.io. Like few days back `runonflux` team deployed GMX interface on an IFPS and when you copy the sharable link it still goes to `app.gmx.io` it should go to the `runonflux` domain. So now the link copy will depend on the domain origin.

---
  - https://app.asana.com/0/0/1203036542861572

### DIFF
--- a/src/domain/common.js
+++ b/src/domain/common.js
@@ -3,3 +3,14 @@ export function get1InchSwapUrl(chainId, from, to) {
   if (!from && !to) return rootUrl;
   return `${rootUrl}/${from}/${to}`;
 }
+
+export function getRootUrl() {
+  if (!window.location.origin) {
+    window.location.origin =
+      window.location.protocol +
+      "//" +
+      window.location.hostname +
+      (window.location.port ? ":" + window.location.port : "");
+  }
+  return window.location.origin;
+}
--- a/src/domain/common.js
+++ b/src/domain/common.js
@@ -3,14 +3,3 @@ export function get1InchSwapUrl(chainId, from, to) {
   if (!from && !to) return rootUrl;
   return `${rootUrl}/${from}/${to}`;
 }
-
-export function getRootUrl() {
-  if (!window.location.origin) {
-    window.location.origin =
-      window.location.protocol +
-      "//" +
-      window.location.hostname +
-      (window.location.port ? ":" + window.location.port : "");
-  }
-  return window.location.origin;
-}
--- a/src/lib/legacy.js
+++ b/src/lib/legacy.js
@@ -1414,20 +1414,7 @@ export function importImage(name) {
 export function getTwitterIntentURL(text, url = "", hashtag = "") {
   let finalURL = "https://twitter.com/intent/tweet?text=";
   if (text.length > 0) {
-    if (Array.isArray(text)) {
-      finalURL += text
-        .map((t) =>
-          encodeURIComponent(t.replace(/[\r\n]+/g, " "))
-            .replace(/\*%7C/g, "*|URL:")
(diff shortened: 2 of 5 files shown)


src/domain/common.js
```diff
@@ -3,3 +3,14 @@ export function get1InchSwapUrl(chainId, from, to) {
   if (!from && !to) return rootUrl;
   return `${rootUrl}/${from}/${to}`;
 }
+
+export function getRootUrl() {
+  if (!window.location.origin) {
+    window.location.origin =
```
divhead, line 9: Hm, does window.location.origin need to be mutated here?
Maybe just return a value? like const url = ...; return url

Also, seems more like 'lib' function than 'domain', i think it could be lib/url.ts 


It's not very obvious right now, but I hope in the future it will be something like this:

**domain** folder - this is logic specific to GMX, such as tokens, orders, positions, etc.

**lib** is low-level technical and infrastructure code. 

Your function `getRootUrl` could be in any other web project or from npm, then it goes in the 'lib' folder, you can also treat it as the 'utils' folder.

Why we separate these layers:
1. it's easier to find specific code without delving into the codebase, e.g:
1.1 I need some specific query for the GMX contract, for example for token statistics, and I want to see some data types and how they are organized - this should be somewhere in 'domain'.
1.2 I need some kind of url formatter or basic callContract function to implement the new query - then it's in 'lib'.

2 It helps to keep the code clear and control dependencies. Ideally the 'lib' code should be unaware of GMX, and remain flexible and reusable, the 'domain' code should contain clear business rules and logic around our data structures without a lot of low-level code about the environment (e.g. browser), frameworks, etc.

vipineth: Got it. Thanks for taking time to explain. 🙏

src/lib/legacy.js
```diff
@@ -1414,9 +1414,19 @@ export function importImage(name) {
...
   let finalURL = "https://twitter.com/intent/tweet?text=";
   if (text.length > 0) {
-    finalURL += encodeURIComponent(text.replace(/[\r\n]+/g, " "))
-      .replace(/\*%7C/g, "*|URL:")
-      .replace(/%7C\*/g, "|*");
+    if (Array.isArray(text)) {
+      finalURL += text
+        .map((t) =>
+          encodeURIComponent(t.replace(/[\r\n]+/g, " "))
+            .replace(/\*%7C/g, "*|URL:")
```
divhead, line 1421: Maybe create a silmple separated function like const formatText = (t) => t.replace(...)?

Just to improve readability

divhead:
not critial comments

divhead (dismissed):

src/lib/legacy.js
gdev8317: or we can remove `if` and use `(Array.isArray(text) ? text : [text]).map( ... )`

src/lib/legacy.js
```diff
@@ -1414,9 +1414,19 @@ export function importImage(name) {
 export function getTwitterIntentURL(text, url = "", hashtag = "") {
   let finalURL = "https://twitter.com/intent/tweet?text=";
   if (text.length > 0) {
-    finalURL += encodeURIComponent(text.replace(/[\r\n]+/g, " "))
-      .replace(/\*%7C/g, "*|URL:")
-      .replace(/%7C\*/g, "|*");
```
gdev8317, line 1419: can we add some comment explaining why we need to replace it please?

gdev8317 (approved):

PR #350, opened 2022-10-26
update font sizes

vipineth:
Update font-size based on this document (https://docs.google.com/spreadsheets/d/1-1hKihNDH483_YeeP55eniM6AQyFDlW08wb02gJ-1BE/edit#gid=0)

---
  - https://app.asana.com/0/0/1202477937360858


divhead:
Did I understand correctly that the goal is to make the design system more consistent?

Maybe we could think about putting font sizes into variables (we have scss in the project now)?

divhead (dismissed):

gdev8317 (approved):

PR #352, opened 2022-11-01
add check for contract address trades

vipineth:
[x] Don't allow ETH withdraw when the user is interaction with a Contract wallet

---
  - https://app.asana.com/0/0/1203247464865808


src/components/Exchange/PositionSeller.js
```diff
@@ -357,6 +360,14 @@ export default function PositionSeller(props) {
 
     receiveToken = isSwapAllowed && swapToToken ? swapToToken : collateralToken;
 
+    // receiveToken does not need to change for STOP order
+    if (isSwapAllowed && isUserAContractAddress && isAddressZero(receiveToken.address)) {
```
divhead, line 364: Sorry, what about the case when user couldn’t change the token (orderOption === STOP)?

src/components/Exchange/PositionsList.js
```diff
@@ -105,6 +106,7 @@ export default function PositionsList(props) {
   const [isPositionShareModalOpen, setIsPositionShareModalOpen] = useState(false);
   const [ordersToaOpen, setOrdersToaOpen] = useState(false);
   const [isHigherSlippageAllowed, setIsHigherSlippageAllowed] = useState(false);
+  const isUserAContractAddress = useIsAddressAContract();
```
divhead, line 109: Maybe we should rename it somehow.)

Like isContractWallet / isContractAccount / isAccountAContract

Just isAddressAContract does not give any information about the address being checked

src/lib/wallets/useIsAddressAContract.ts
```diff
@@ -0,0 +1,17 @@
+import { useEffect, useState } from "react";
+import { useWeb3React } from "@web3-react/core";
+
+export default function useIsAddressAContract() {
+  const { active, account, library } = useWeb3React();
+  const [isContract, setIsContract] = useState(false);
```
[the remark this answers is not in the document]
divhead: or create additional state  variable `isLoading`

src/components/Exchange/PositionSeller.js
divhead: As I understand it, in this case, if there is a long with collateral in ETH, then when triggerOrder is triggered by the actual user will get WETH or how will this happen?

Is there a similar check on the backend and should we give an explanation here?
@gdev8317

src/components/Exchange/PositionsList.js
vipineth: 👍

src/components/Exchange/PositionSeller.js
vipineth: Order book contract is not affected by this. So we don't need check for STOP order. I asked X earlier about this.

src/components/Exchange/PositionSeller.js
```diff
@@ -358,6 +361,14 @@ export default function PositionSeller(props) {
 
     receiveToken = isSwapAllowed && swapToToken ? swapToToken : collateralToken;
 
+    // receiveToken does not need to change for STOP order
+    if (isSwapAllowed && isContractAccount && isAddressZero(receiveToken.address)) {
+      const wrappedToken = toTokens.find((t) => t.baseSymbol === nativeTokenSymbol);
+      setSwapToToken(wrappedToken);
+      setSavedRecieveTokenAddress(wrappedToken.address);
```
gdev8317, line 368: i think better to change state (call set* functions) inside e.g. some useEffect hook? otherwise by calling setState inside component render function you can in theory go into recursion and this is a common pattern to change state in handlers or hooks in general?

src/lib/wallets/useIsContractAccount.ts
```diff
@@ -0,0 +1,17 @@
...
+import { useWeb3React } from "@web3-react/core";
+
+export default function useIsContractAccount() {
+  const { active, account, library } = useWeb3React();
+  const [isContract, setIsContract] = useState<boolean | null>(null);
+
+  useEffect(() => {
+    if (!active) return;
+    (async function () {
+      const code = await library.getCode(account);
```
gdev8317, line 11: should we memoize it? EOA address can't become contract and vice versa. so no need to send request again 

or maybe i'm over-optimizing, there are a lot of non-optimized things anyway :)

src/components/Exchange/PositionSeller.js
vipineth: Updated!

gdev8317 (dismissed):

gdev8317 (approved):

PR #353, opened 2022-11-01
update poolAmount to managedAmount on glpswap

vipineth:
[x] Update GLP Swap to use `managedAmount` instead of `poolAmount`

---
  - https://app.asana.com/0/0/1203247464865849
  - https://app.asana.com/0/0/1203199570847028

### DIFF
--- a/src/components/Stake/GMXAprTooltip.tsx
+++ b/src/components/Stake/GMXAprTooltip.tsx
@@ -1,0 +1,52 @@
+import { Trans } from "@lingui/macro";
+import StatsTooltipRow from "components/StatsTooltip/StatsTooltipRow";
+import { formatKeyAmount } from "lib/numbers";
+
+export default function GMXAprTooltip({ processedData, nativeTokenSymbol }) {
+  return (
+    <>
+      <StatsTooltipRow
+        label="Escrowed GMX APR"
+        showDollar={false}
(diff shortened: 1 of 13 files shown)


src/components/Stake/GMXAprTooltip.tsx
```diff
@@ -0,0 +1,52 @@
+import { Trans } from "@lingui/macro";
+import StatsTooltipRow from "components/StatsTooltip/StatsTooltipRow";
+import { formatKeyAmount } from "lib/numbers";
+
+export default function GMXAprTooltip({ processedData, nativeTokenSymbol }) {
```
divhead, line 5: The `processedData` structure does not look very clear, I think it would be good to add `type Props` in this component 🙏

src/components/StatsTooltip/StatsTooltipRow.tsx
```diff
@@ -1,7 +1,7 @@
 import "./StatsTooltip.css";
 type Props = {
   label: string;
-  value: number | [number];
+  value: number | string | [string] | [number];
```
divhead, line 4: The `[string]` type means "Array with exactly one string element".

If you want to say "Array of strings," you must use the type `string[]`

Screenshot 2022-11-04 at 19 38 43

vipineth: Resolved! Sorry silly mistake 🥲

PR #354, opened 2022-11-02
fix showing stable for short trigger orders

vipineth:
[x] Show stable asset for short trigger orders.
[x] Show no value when trigger price is not added

---
  - https://app.asana.com/0/0/1203247464865844

### DIFF
--- a/src/components/Exchange/PositionSeller.js
+++ b/src/components/Exchange/PositionSeller.js
@@ -835,7 +835,7 @@ export default function PositionSeller(props) {
 
   const shouldShowExistingOrderWarning = false;
 
-  if (!isSwapAllowed && receiveToken.isStable && !triggerPriceUsd) {
+  if (orderOption === STOP && !triggerPriceUsd) {
     receiveAmount = bigNumberify(0);
     convertedReceiveAmount = bigNumberify(0);
   }
--- a/src/components/Exchange/PositionSeller.js
+++ b/src/components/Exchange/PositionSeller.js
@@ -382,8 +382,7 @@ export default function PositionSeller(props) {
 
     if (orderOption === STOP) {
       convertedReceiveAmount = getTokenAmountFromUsd(infoTokens, receiveToken.address, receiveAmount, {
-        overridePrice: triggerPriceUsd,
-        isShort: !position?.isLong,
+        overridePrice: receiveToken.isStable && triggerPriceUsd,
       });
     } else {
       convertedReceiveAmount = getTokenAmountFromUsd(infoTokens, receiveToken.address, receiveAmount);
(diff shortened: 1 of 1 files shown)


src/components/Exchange/PositionSeller.js
```diff
@@ -835,6 +836,10 @@ export default function PositionSeller(props) {
 
   const shouldShowExistingOrderWarning = false;
 
+  if (!isSwapAllowed && !position.isLong && !triggerPriceUsd) {
```
Twllerm, line 839: I think it would make more sense to use the orderOption === STOP condition here instead of isSwapAllowed

Because isSwapAllowed is not related to triggerPrice and it's not really clear why this condition is here until you dive into the code

vipineth: Done!

src/components/Exchange/PositionSeller.js
```diff
@@ -382,7 +382,7 @@ export default function PositionSeller(props) {
 
     if (orderOption === STOP) {
       convertedReceiveAmount = getTokenAmountFromUsd(infoTokens, receiveToken.address, receiveAmount, {
-        overridePrice: triggerPriceUsd,
+        overridePrice: !receiveToken.isStable && triggerPriceUsd,
```
divhead, line 385: Oh, I'm afraid that this place can be confusing and generate bugs in the future since the logic is not intuitive so far

But I would wait for a comment from G

I thought, it would make more sense to rewrite it like this, so there would be less confusion
(Clearly indicate that here we get the **collateral**,  explicitly specify the order type "Long", add comment)
```
convertedReceiveAmount = getTokenAmountFromUsd(infoTokens, collateralToken.address, receiveAmount, {
        overridePrice: triggerPriceUsd,
        // For Shorts, the collateral is a stable coin, it should not depend on the triggerPrice 
        overridePrice: order.isLong  && triggerPriceUsd,
}
```

vipineth: Makes sense.

I think we started with `isLong` moved to `isStable` because it would reduce the calculation in case where the position is long and relieve token is a stable. IDK if this makes sense.

gdev8317: > I think we started with isLong moved to isStable because it would reduce the calculation in case where the position is long and relieve token is a stable.

* not sure i got it. in this specific case isStable can be true if only isLong == false and vice versa (you can't change receive token for STOP orders, i.e. if received token is stable then it's a short)
* but tbh i'm not sure using `isLong` will make it much more clear
* i think condition on line should be updated https://github.com/gmx-io/gmx-interface/pull/354/files#diff-09f2f719db7be699f5435ad3ac7a5d22a5e9a3b5d78cf45c72efaf8beb8d283eL383 instead of changing `overridePrice` param? because else-branch does exactly what we need

gdev8317: some comment might help as well, a bit tricky place, i've spent 20 minutes just to understand whats happening here :)

vipineth: Sure, I will add the comments to explain.

vipineth: I have added a comment and changed the condition at  https://github.com/gmx-io/gmx-interface/pull/354/files#diff-09f2f719db7be699f5435ad3ac7a5d22a5e9a3b5d78cf45c72efaf8beb8d283eL383  as G mentioned.

gdev8317 (approved):

PR #362, opened 2022-11-17
Fix davatar rpc errors

divhead:
Using custom etherium rpc provider for Davatar

https://app.asana.com/0/1201404209663627/1203385478429923/f


src/lib/rpc/index.ts
```diff
@@ -32,3 +33,21 @@ export function getFallbackProvider(chainId: number) {
     { chainId }
   );
 }
+
+export function useJsonRpcProvider(p: { chainId: number }) {
```
gdev8317, line 37: probably better to just pass chainId: number argument instead of object?

PR #370, opened 2022-11-28
TradingView Implementation

vipineth:
---
  - https://app.asana.com/0/0/1201266503318581


src/components/Exchange/ExchangeTVChart.js
```diff
@@ -151,29 +79,60 @@ export default function ExchangeTVChart(props) {
     setChartToken(tmp);
   }, [swapOption, fromToken, toToken, chainId]);
 
-  const symbol = chartToken ? (chartToken.isWrapped ? chartToken.baseSymbol : chartToken.symbol) : undefined;
-  const marketName = chartToken ? symbol + "_USD" : undefined;
-  const previousMarketName = usePrevious(marketName);
+  // const symbol = chartToken ? (chartToken.isWrapped ? chartToken.baseSymbol : chartToken.symbol) : undefined;
+  // const marketName = chartToken ? symbol + "_USD" : undefined;
+  // const previousMarketName = usePrevious(marketName);
```
gdev8317, line 84: will we need this code in future? or better to just remove it?

gdev8317:
chart doesn't work for me here https://deploy-preview-370--superb-tarsier-e2aa29.netlify.app/#/trade

that's probably caused by content-type-security and/or x-frame-options headers, the same issue may happen in production i guess

chart is embedded as iframe?

image

gdev8317:
there are also some network request errors, does TV send analytics? is it possible to remove these requests?

image

src/components/Exchange/ExchangeTVChart.js
```diff
@@ -151,29 +79,60 @@ export default function ExchangeTVChart(props) {
...
+    }
+    return positions
+      .filter((p) => p.indexToken.address === chartToken.address)
+      .map((position) => {
+        return {
+          open: {
+            price: parseFloat(formatAmount(position.averagePrice, USD_DECIMALS, 2)),
+            title: `Open ${position.indexToken.symbol} ${position.isLong ? "Long" : "Short"}`,
+          },
+          liq: {
```
gdev8317, line 127: might be better to use full name like "liquidationPrice" for concistency? i guess we don't abbreviate it in other places

src/components/Exchange/ExchangeTVChart.js
```diff
@@ -464,13 +314,18 @@ export default function ExchangeTVChart(props) {
...
-          <div className="ExchangeChart-bottom-controls">
-            <Tab options={Object.keys(CHART_PERIODS)} option={period} setOption={setPeriod} />
-          </div>
-          {candleStatsHtml}
-        </div>
-        <div className="ExchangeChart-bottom-content" ref={chartRef}></div>
+        {chartToken.symbol && chainId && (
+          <TVChartContainer
+            currentPositions={currentPositions}
+            currentOrders={currentOrders}
```
gdev8317, line 320: if i understand correctly idea is to have separate `TVChartContainer` component without domain logic related to orders/positions to keep it pure, e.g. it does not calculate liquidation prices for positions

i think it can accept prop `lines` (or probably some better name) instead of currentPositions/currentOrders:
* TVChartContainer would be more "pure"
* currentPosition and currentOrders are almost the same, so it would remove some duplication

src/components/TVChartContainer/TVChartContainer.tsx
```diff
@@ -0,0 +1,142 @@
...
+    return () => {
+      lines.forEach((line) => line?.remove());
+    };
+  }, [currentPositions, savedShouldShowPositionLines, currentOrders, drawLineOnChart]);
+
+  useEffect(() => {
+    if (chartReady && tvWidgetRef.current && symbol !== tvWidgetRef.current?.activeChart?.().symbol()) {
+      tvWidgetRef.current.setSymbol(symbol, tvWidgetRef.current.activeChart().resolution(), () => {});
+    }
+  }, [symbol, chartReady, period]);
```
gdev8317, line 79: `tvWidgetRef.current` is not part of hook dependencies. 

in case hook was trigger with tvWidgetRef.current === undefined it would never be called again until symbol/chartReady/period are updated

how does that work? maybe there is no need to check `tvWidgetRef.current`?

src/components/TVChartContainer/TVChartContainer.tsx
```diff
@@ -0,0 +1,142 @@
...
+
+  useEffect(() => {
+    if (chartReady && tvWidgetRef.current) {
+      if (isMobile && tvWidgetRef.current.activeChart().getCheckableActionState("drawingToolbarAction")) {
```
gdev8317, line 83: 1. i think it's good to avoid nested `if`s
hook may start with `if (!chartReady || !tvWidgetRef.current || !isMobile) { return }`

2. and the same question for `tvWidgetRef.current` – it's not part of deps

src/components/TVChartContainer/TVChartContainer.tsx
gdev8317: at first glance if `chartReady` is true then tvWidgetRef is defined? and btw why ref is preferable here instead of doing something like `setState(tvWidget)`?

src/components/TVChartContainer/charting_library/charting_library.d.ts
```diff
@@ -0,0 +1,2435 @@
+// Generated by dts-bundle-generator v5.3.0
```
gdev8317, line 1: this is original file from TV library, right? it can't be moved to `public/charting_library` dir because typescript won't see it? 

just would be good to not move it to our components if possible

src/components/TVChartContainer/charting_library/package.json
```diff
@@ -0,0 +1,8 @@
+{
```
gdev8317, line 1: do we need this?

src/components/TVChartContainer/constants.ts
```diff
@@ -0,0 +1,125 @@
...
+  "header_in_fullscreen_mode",
+];
+const enabledFeatures = [
+  "side_toolbar_in_fullscreen_mode",
+  "header_in_fullscreen_mode",
+  "hide_resolution_in_legend",
+  "items_favoriting",
+];
+
+export class SaveLoadAdapter {
```
gdev8317, line 52: should it be moved outside "constants"?

src/components/TVChartContainer/constants.ts
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  saveChart(chartData) {
+    if (!chartData.id) {
+      chartData.id = Math.random().toString();
+    } else {
+      this.removeChart(chartData.id);
+    }
+
+    chartData.timestamp = new Date().valueOf();
```
gdev8317, line 88: interesting, i usually use just `Date.now()`. probably matter of taste, but maybe use it since it looks a bit simpler?

src/domain/prices.ts
```diff
@@ -56,7 +67,25 @@ function fillGaps(prices, periodSeconds) {
   return newPrices;
 }
 
-async function getChartPricesFromStats(chainId, symbol, period) {
+export async function getLimitChartPricesFromStats(chainId, symbol, period, limit = 1) {
+  if (["WBTC", "WETH", "WAVAX"].includes(symbol)) {
```
gdev8317, line 71: code to normalize symbol name is duplicated, let's move it to separate function?

src/domain/prices.ts
```diff
@@ -158,7 +181,7 @@ function getCandlesFromPrices(prices, period) {
   }));
 }
 
-function getChainlinkChartPricesFromGraph(tokenSymbol, period) {
+export function getChainlinkChartPricesFromGraph(tokenSymbol, period) {
   if (["WBTC", "WETH", "WAVAX"].includes(tokenSymbol)) {
     tokenSymbol = tokenSymbol.substr(1);
   }
```
gdev8317, line 187: oh, i think there was bug – it doesn't handle "BTC.b", could you please fix it here?

src/domain/tradingview/helper.ts
```diff
@@ -0,0 +1,119 @@
...
+  } else {
+    const newBar = {
+      time: currentCandleTime,
+      open: lastBar.close,
+      close: averagePriceValue,
+      high: Math.max(lastBar.close, averagePriceValue),
+      low: Math.min(lastBar.close, averagePriceValue),
+      ticker,
+    };
+    lastBar = newBar;
```
gdev8317, line 116: is there a reason to override lastBar value? or `let lastBar ...` can be changed to `const lastBar ...`?

src/domain/tradingview/useTVDatafeed.ts
```diff
@@ -0,0 +1,101 @@
...
+            .map((t) => t.symbol);
+
+          return {
+            name: symbol,
+            type: "crypto",
+            description: symbol + " / USD",
+            ticker: symbol,
+            session: "24x7",
+            minmov: 1,
+            pricescale: 100, // 	or 100
```
gdev8317, line 37: 100 or 100? :)

src/domain/tradingview/useTVDatafeed.ts
```diff
@@ -0,0 +1,101 @@
...
+            timezone: "Etc/UTC",
+            has_intraday: true,
+            has_daily: true,
+            currency_code: "USD",
+            visible_plots_set: true,
+            isStable: stableTokens.includes(symbol),
+          };
+        };
+        const symbol = onSymbolResolvedCallback(await symbolInfo(symbolName));
```
gdev8317, line 46: looks a bit hack-ish to call some callback here. is it possible to move this callback to e.g. TVChartContainer? or maybe you can explain why it should be here?

src/domain/tradingview/useTVDatafeed.ts
```diff
@@ -0,0 +1,101 @@
...
+
+        return symbol;
+      },
+
+      getBars: async (symbolInfo, resolution, periodParams, onHistoryCallback, onErrorCallback) => {
```
gdev8317, line 51: and same question for callbacks here

vipineth:
> chart doesn't work for me here https://deploy-preview-370--superb-tarsier-e2aa29.netlify.app/#/trade
> 
> that's probably caused by content-type-security and/or x-frame-options headers, the same issue may happen in production i guess
> 
> chart is embedded as iframe?
> 
> image

Yes it's imbedded as a iframe. We will need to update the content-type-security

vipineth:
> there are also some network request errors, does TV send analytics? is it possible to remove these requests?
> 
> image

This is only happening on the netlify build. Maybe coming from netlify.

src/components/Exchange/ExchangeTVChart.js
vipineth: Sure!

src/domain/prices.ts
vipineth: Sure!

src/domain/tradingview/useTVDatafeed.ts
vipineth: Removed!

src/domain/tradingview/useTVDatafeed.ts
vipineth: I have refactored! 

I am using await for fix this error form TV:

 `resolveSymbol` should return result asynchronously. Use `setTimeout` with 0 interval to execute the callback function.

src/components/TVChartContainer/charting_library/package.json
vipineth: I will remove them. I left them to some predefined types. We are not using them right now. If needed we will create.

src/components/TVChartContainer/charting_library/charting_library.d.ts
vipineth: Removed

src/components/TVChartContainer/charting_library/package.json
vipineth: I will remove

src/domain/prices.ts
vipineth: Done

src/components/TVChartContainer/TVChartContainer.tsx
vipineth: I thought of using the state to store the `tvWidget` but I was facing couple of issue.

- I feel using `ref` is a better choice because we are storing the instance of the widget once created and the instance can be manipulated. TradingView charting library is an external system and it makes sense to store it in ref.
- When I was testing to store the widget instance into a state some of the overrides were not working properly. Like one below. With ref it works normally but with state the color does't apply.

```js
 tvWidget?.applyOverrides({
        "paneProperties.background": "#16182e",
        "paneProperties.backgroundType": "solid",
  });
```

src/config/localStorage.ts
```diff
@@ -12,5 +12,6 @@ export const DISABLE_ORDER_VALIDATION_KEY = "disable-order-validation";
 export const SHOULD_SHOW_POSITION_LINES_KEY = "Exchange-swap-should-show-position-lines";
 export const REFERRAL_CODE_KEY = "GMX-referralCode";
 export const REFERRALS_SELECTED_TAB_KEY = "Referrals-selected-tab";
+export const TV_SAVE_LOAD_CHARTS = "tv-save-load-charts";
```
divhead, line 15: Maybe add _KEY for consistency?

src/components/TVChartContainer/constants.ts
```diff
@@ -0,0 +1,124 @@
...
+  "paneProperties.vertGridProperties.style": 2,
+  "paneProperties.horzGridProperties.color": "rgba(35, 38, 59, 1)",
+  "paneProperties.horzGridProperties.style": 2,
+  "mainSeriesProperties.priceLineColor": "#3a3e5e",
+  "scalesProperties.textColor": "#fff",
+  "scalesProperties.lineColor": "#16182e",
+  ...chartStyleOverrides,
+};
+
+export const disabledFeaturesOnMonile = ["header_saveload", "header_fullscreen_button"];
```
divhead, line 29: I think there's a typo, do you mean disabledFeaturesOnMo**b**ile?

src/domain/tradingview/helper.ts
```diff
@@ -0,0 +1,120 @@
...
+  getStablePriceData,
+  timezoneOffset,
+} from "domain/prices";
+import { CHART_PERIODS, USD_DECIMALS } from "lib/legacy";
+import { formatAmount } from "lib/numbers";
+
+export const supportedResolutions = { 5: "5m", 15: "15m", 60: "1h", 240: "4h", "1D": "1d" };
+const LAST_BAR_REFRESH_INTERVAL = 15000; // 15 seconds
+
+export function getPeriodFromResolutions(value, object = supportedResolutions) {
```
divhead, line 16: I think it's better to split the code here:
1. Put the requests in requests.ts
2. Simple utils which just format something could be be placed in utils.ts

It is also worth  add types to all functions here and in the useTVDatafeed hook

src/domain/tradingview/helper.ts
```diff
@@ -0,0 +1,120 @@
...
+  }
+}
+
+export async function getHistoryBars({ ticker, resolution, chainId, isStable, countBack }) {
+  const period = supportedResolutions[resolution];
+  const bars = isStable ? getStablePriceData(period, countBack) : await getTokenChartPrice(chainId, ticker, period);
+  return bars.map(formatTimeInBar);
+}
+
+const getLastBarAfterInterval = (function () {
```
divhead, line 70: The closure looks a little dangerous here - The variables startTime, lastBar, lastTicker, lastPeriod are initiated when this file is initialized and we have no way to control / reset them

I'm not sure that this could be a practical problem right now, but in theory it could be a place that is hard to debug. (for example, if we will need to use TVChart in several places)

Maybe we could pass these variables through parameters and store them in a state or ref? Or at least create a closure in the hook and pass this function as a callback to getLiveBar

src/domain/tradingview/helper.ts
divhead: But, why redefine lastBar here?
This is a clear function without a cycle, every time you call it, **lastBar** will be re-initialized, it seems `lastBar = newBar` shouldn't have any effect

Correct me if I'm wrong, but otherwise it looks like magic)

src/domain/tradingview/helper.ts
```diff
@@ -0,0 +1,120 @@
...
+    return lastBar;
+  };
+})();
+
+export async function getLiveBar({ ticker, resolution, chainId }) {
+  const period = supportedResolutions[resolution];
+  if (!ticker || !period || !chainId) return;
+
+  const periodSeconds = CHART_PERIODS[period];
+  const currentCandleTime = (Math.floor(Date.now() / 1000 / periodSeconds) * periodSeconds + timezoneOffset) * 1000;
```
divhead, line 92: Maybe leave a comment on this formula? Right now it's not very clear how it counts

src/components/TVChartContainer/constants.ts
```diff
@@ -0,0 +1,124 @@
...
+  "items_favoriting",
+];
+
+export class SaveLoadAdapter {
+  charts: any[];
+  setTvCharts: (a: any[]) => void;
+  onSelectToken: (token: Token) => void;
+  chainId: number;
+
+  constructor(chainId, charts, setTvCharts, onSelectToken) {
```
divhead, line 58: Can we add types to the parameters in the functions here, if it's not too hard?

src/components/TVChartContainer/constants.ts
```diff
@@ -0,0 +1,124 @@
...
+        return Promise.resolve();
+      }
+    }
+
+    return Promise.reject();
+  }
+
+  saveChart(chartData) {
+    if (!chartData.id) {
+      chartData.id = Math.random().toString();
```
divhead, line 83: it's okay to mutate chartData here? It will change outside this function

src/components/TVChartContainer/constants.ts
divhead: I think it could be a SaveLoadAdapter.ts file

src/components/TVChartContainer/TVChartContainer.tsx
```diff
@@ -0,0 +1,142 @@
...
+
+    return () => {
+      if (tvWidgetRef.current) {
+        tvWidgetRef.current.remove();
+        tvWidgetRef.current = null;
+        setChartReady(false);
+      }
+    };
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
divhead, line 139: this hook runs only a the first render, so, datafeed will be set only once, Is that okay?

src/components/TVChartContainer/charting_library/charting_library.d.ts
divhead: Probably we will create some utils that uses types from here, or create different version of TvChartContainer (e.g. for synthetics) maybe put it into src/charting_library.d.ts - it's a common way to place typings for external libraries

public/tradingview-chart.css
```diff
@@ -0,0 +1,104 @@
+.theme-dark:root {
```
divhead, line 1: Is it our custom styles? Can we put it into TVChartContainer.scss?
It's not obvious that we have some editable files in /public folder

src/components/TVChartContainer/TVChartContainer.tsx
```diff
@@ -0,0 +1,142 @@
...
+          .setBodyTextColor("#fff")
+          .setLineColor("#3a3e5e")
+          .setBodyBackgroundColor("#3a3e5e")
+          .setBodyBorderColor("#3a3e5e");
+      }
+    },
+    [chartReady]
+  );
+
+  useEffect(() => {
```
divhead, line 58: I think it would be more readable to give a name for function inside effect, e.g
```
useEffect(function updateLines() {...}, [...])
```

src/domain/tradingview/helper.ts
vipineth: Sorry I was mistaken. Earlier the structure was such that the lastBar was outside the function. So we needed to replace with the new values. I will fix.

public/tradingview-chart.css
vipineth: The docs says the URL needs to be absolute or relative path to static/public folder. The chart is created using the IFrame element.

src/components/TVChartContainer/TVChartContainer.tsx
vipineth: Yes! Otherwise chart will re-load again and again.

src/components/TVChartContainer/constants.ts
vipineth: Yeah, it's okay based on the documentation. It's also just for storing to the localhost.

src/domain/tradingview/helper.ts
vipineth: Done!

src/domain/tradingview/useTVDatafeed.ts
```diff
@@ -0,0 +1,91 @@
...
+        try {
+          const bars = await getHistoryBars(chainId, ticker, resolution, isStable, countBack);
+          const filteredBars = bars.filter((bar) => bar.time >= from * 1000 && bar.time < toWithOffset * 1000);
+          if (filteredBars.length > 0) {
+            onHistoryCallback(filteredBars, { noData: false });
+          } else {
+            onHistoryCallback(filteredBars, { noData: true });
+          }
+        } catch {
+          onErrorCallback("Something went wrong!");
```
gdev8317, line 68: let's use more specific error message?

src/domain/tradingview/useTVDatafeed.ts
```diff
@@ -0,0 +1,91 @@
...
+        }
+        const { ticker, isStable } = symbolInfo;
+        if (activeTicker.current !== ticker) {
+          activeTicker.current = ticker;
+        }
+
+        try {
+          const bars = await getHistoryBars(chainId, ticker, resolution, isStable, countBack);
+          const filteredBars = bars.filter((bar) => bar.time >= from * 1000 && bar.time < toWithOffset * 1000);
+          if (filteredBars.length > 0) {
```
gdev8317, line 62: maybe just this?
```
onHistoryCallback(filteredBars, { noData: filteredBars.length === 0 });
```

src/domain/tradingview/useTVDatafeed.ts
```diff
@@ -0,0 +1,91 @@
...
+      },
+      resolveSymbol: async (symbolName, onSymbolResolvedCallback) => {
+        const stableTokens = getTokens(chainId)
+          .filter((t) => t.isStable)
+          .map((t) => t.symbol);
+        const symbolInfo = {
+          name: symbolName,
+          type: "crypto",
+          description: symbolName + " / USD",
+          ticker: symbolName,
```
gdev8317, line 33: is there a public documentation for widget? would be good to leave link here so it's easier to understand what this config fields mean

src/components/TVChartContainer/TVChartContainer.tsx
gdev8317: > I feel using ref is a better choice because we are storing the instance of the widget once created and the instance can be manipulated. TradingView charting library is an external system and it makes sense to store it in ref.

not sure i understand this. currently in production we use setState to store a reference to TV, why is it different here?

> When I was testing to store the widget instance into a state some of the overrides were not working properly. Like one below. With ref it works normally but with state the color does't apply.

i guess that should not depend on whether it's a setState or useRef. if `tvWidget` has a reference to correct object then it should work? wondering how can setState affect it?

> Because we are storing the widget instance inside a ref which can be mutable we can't use them in dependency array.

i believe react checks reference, not its values recursively? 

in general it might not be critical, just trying to understand why ref is better here

gdev8317: and for this comment https://github.com/gmx-io/gmx-interface/pull/370#discussion_r1071696707, is there are issue i've described?

src/components/Exchange/ExchangeTVChart.js
gdev8317: pinging in case this comment was missed

src/domain/tradingview/useTVDatafeed.ts
vipineth: Sorry the repo that contains the docs is not public.

src/components/Exchange/ExchangeTVChart.js
vipineth: Sorry I missed this one. This one is really good idea. I have updated the code based on this.

src/components/TVChartContainer/TVChartContainer.tsx
vipineth: > and for this comment [#370 (comment)](https://github.com/gmx-io/gmx-interface/pull/370#discussion_r1071696707), is there are issue i've described?

Because we are using the function `setSymbol` from `tvWidgetRef.current` and it can be undefined I am using that so we can remove the error `'tvWidgetRef.current' is possibly 'null'`. 

There are two things `chartReady` that is stored in react state and `tvWidget` which is stored in ref. As of now I am using ref and I don't know why moving from ref for tvWidget to state doesn't work as with the ref (https://github.com/gmx-io/gmx-interface/pull/370#discussion_r1080861097). I am using chartReady to re-render the component on change of tvWidget. Once I find the answer for the above I will remove chartReady.

src/components/TVChartContainer/TVChartContainer.tsx
vipineth: chartReady is the state and it depends on the `tvWidgetRef` so using that instead.

src/domain/tradingview/useTVDatafeed.ts
vipineth: It's just a async function. I have updated it to function for better readability.

src/domain/prices.ts
```diff
@@ -11,6 +11,13 @@ import { formatAmount } from "lib/numbers";
 
 const BigNumber = ethers.BigNumber;
 
+function getNormalizedSymbol(tokenSymbol) {
+  if (["WBTC", "WETH", "WAVAX", "BTC.b"].includes(tokenSymbol)) {
+    return tokenSymbol.includes(".") ? tokenSymbol.split(".")[0] : tokenSymbol.substr(1);
```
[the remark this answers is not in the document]
vipineth: Done!

src/components/TVChartContainer/constants.ts
vipineth: Fixed!

src/domain/tradingview/helper.ts
vipineth: Done!

src/components/TVChartContainer/TVChartContainer.tsx
vipineth: Done!

src/domain/tradingview/useTVDatafeed.ts
vipineth: Sorry! Updated

src/domain/tradingview/utils.ts
```diff
@@ -0,0 +1,13 @@
+export const supportedResolutions = { 5: "5m", 15: "15m", 60: "1h", 240: "4h", "1D": "1d" };
```
divhead, line 1: A little about consistency:
1. Maybe call it SUPPORTED_RESOLUTIONS like other constants?
2. It looks like some settings that can be changed, maybe put it in config/tradingView.ts?

src/domain/tradingview/utils.ts
```diff
@@ -0,0 +1,13 @@
...
+export const LAST_BAR_REFRESH_INTERVAL = 15000; // 15 seconds
+
+export function getPeriodFromResolutions(value, object = supportedResolutions) {
```
divhead, line 4: object -> resolutions?

value - what's value?) if it's resolution, maybe call it "resolution"?

src/domain/tradingview/utils.ts
```diff
@@ -0,0 +1,13 @@
...
+  return Object.keys(object).find((key) => object[key] === value);
+}
+
+export function formatTimeInBar(bar) {
```
divhead, line 8: It is not clear what the format of the bar is, it seems to me that we should add the types here

src/domain/tradingview/utils.ts
vipineth: Done!

src/domain/tradingview/utils.ts
vipineth: Renamed

src/domain/tradingview/utils.ts
vipineth: Type added

src/domain/tradingview/requests.ts
```diff
@@ -0,0 +1,40 @@
...
+  try {
+    prices = await getChartPricesFromStats(chainId, symbol, period);
+  } catch (ex) {
+    // eslint-disable-next-line no-console
+    console.warn(ex, "Switching to graph chainlink data");
+    try {
+      prices = await getChainlinkChartPricesFromGraph(symbol, period);
+    } catch (ex2) {
+      // eslint-disable-next-line no-console
+      console.warn("getChainlinkChartPricesFromGraph failed", ex2);
```
divhead, line 16: previous console logging ex and message in reverse format)

divhead (dismissed):

divhead (approved):

src/components/TVChartContainer/SaveLoadAdapter.ts
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    return Promise.reject();
+  }
+
+  saveChart(chartData) {
+    if (!chartData.id) {
+      chartData.id = Math.random().toString();
+    } else {
+      this.removeChart(chartData.id);
```
gdev8317, line 44: if `this.charts` is falsie then `removeChart` will be rejected. should `saveChart` be rejected in this case as well?

gdev8317: and is there a reason why chart is removed here? few lines later it will be pushed to `this.charts`

src/components/TVChartContainer/TVChartContainer.tsx
```diff
@@ -0,0 +1,172 @@
...
+  onSelectToken,
+  period,
+  setPeriod,
+}: Props) {
+  const chartContainerRef = useRef<HTMLDivElement | null>(null);
+  const tvWidgetRef = useRef<IChartingLibraryWidget | null>(null);
+  const [chartReady, setChartReady] = useState(false);
+  const [tvCharts, setTvCharts] = useLocalStorage<ChartData[] | undefined>(TV_SAVE_LOAD_CHARTS_KEY, []);
+  const { datafeed, resetCache } = useTVDatafeed();
+  const isMobile = useMedia("(max-width: 550px)");
```
gdev8317, line 42: we already define `isMobile` in PositionShare component, probably better to create hook like `useIsMobile` to not repeat this specific media query

src/components/TVChartContainer/TVChartContainer.tsx
```diff
@@ -0,0 +1,172 @@
...
+  const symbolRef = useRef(symbol);
+
+  const drawLineOnChart = useCallback(
+    (title: string, price: number) => {
+      if (chartReady && tvWidgetRef.current?.activeChart?.().dataReady()) {
```
gdev8317, line 47: is it possible that chartReady == true but the second condition is false?

1. if yes, them useCallback deps should contain something related to tvWidgetRef.current.(etc)?
2. if no only using `chartReady` inside condition should be sufficient?

src/components/TVChartContainer/TVChartContainer.tsx
gdev8317: not sure if i explained it correctly.

here `useEffect` hook uses `tvWidgetRef.current?.activeChart?.().symbol()` inside. however there is no `tvWidgetRef` inside hook dependencies. 

is it possible that `tvWidgetRef.current?.activeChart?.().symbol()` value changes, but hook would not call callback because dependencies didn't change?

for example in this case alert won't be called, right?:
```
const [a, setA] = useState(1);
const [b, setB] = useState(2);
setB(1);
useEffect(() => {
  if (a == b) { alert("equal"); }
}, [a]);
```

src/components/TVChartContainer/TVChartContainer.tsx
```diff
@@ -0,0 +1,172 @@
...
+      user_id: defaultChartProps.userId,
+      fullscreen: defaultChartProps.fullscreen,
+      autosize: defaultChartProps.autosize,
+      custom_css_url: defaultChartProps.custom_css_url,
+      overrides: defaultChartProps.overrides,
+      interval: getObjectKeyFromValue(period, SUPPORTED_RESOLUTIONS),
+      save_load_adapter: new SaveLoadAdapter(chainId, tvCharts, setTvCharts, onSelectToken),
+    };
+    tvWidgetRef.current = new window.TradingView.widget(widgetOptions);
+    tvWidgetRef.current?.onChartReady(function () {
```
gdev8317, line 143: `current` is always defined here, right? if so you can use `!` instead of `?` to tell typescript that value is defined here

src/domain/tradingview/TVRequests.ts
```diff
@@ -0,0 +1,100 @@
...
+  lastPeriod: string;
+
+  constructor() {
+    this.lastBar = null;
+    this.startTime = 0;
+    this.lastTicker = "";
+    this.lastPeriod = "";
+  }
+
+  async getLastBar(chainId, ticker, period) {
```
gdev8317, line 31: let's add types to parameters?

src/domain/tradingview/TVRequests.ts
```diff
@@ -0,0 +1,100 @@
...
+      // eslint-disable-next-line no-console
+      console.error(error);
+    }
+
+    if (!this.lastBar) return;
+
+    const currentPrice = await getCurrentPriceOfToken(chainId, ticker);
+    const averagePriceValue = parseFloat(formatAmount(currentPrice, USD_DECIMALS, 4));
+    if (this.lastBar.time && currentCandleTime === this.lastBar.time && ticker === this.lastBar.ticker) {
+      return {
```
gdev8317, line 80: just to confirm, in this case you don't need to update `this.lastBar`, right?

src/domain/tradingview/useTVDatafeed.ts
```diff
@@ -0,0 +1,97 @@
...
+  supported_resolutions: Object.keys(SUPPORTED_RESOLUTIONS),
+  supports_marks: false,
+  supports_timescale_marks: false,
+  supports_time: true,
+  reset_cache_timeout: 100,
+};
+
+export default function useTVDatafeed() {
+  const { chainId } = useChainId();
+  const intervalRef = useRef<ReturnType<typeof setInterval> | null>();
```
gdev8317, line 18: i think it would return undefined by default, not null. so correct type would be `ReturnType<typeof setInterval> | undefined>`

src/domain/tradingview/useTVDatafeed.ts
```diff
@@ -0,0 +1,97 @@
...
+  const resetCacheRef = useRef<() => void | undefined>();
+  const activeTicker = useRef<string | null>();
```
gdev8317, line 20: and null => undefined here?

gdev8317 (approved):

src/domain/tradingview/TVRequests.ts
vipineth: Yes, we dont need to update the lastBar in this case

src/components/TVChartContainer/TVChartContainer.tsx
vipineth: Yes, it's possible. SO `chartReady` means if the chart is initially rendered. When wen change the token on the rendered chart the value of dataReady can change and be false.

Regarding putting the value of `useRef` i.e tvWidgetRef to the dependency array does't makes sense because `useRef` stores a mutable value. Please let me know if I am missing something.

src/components/TVChartContainer/TVChartContainer.tsx
vipineth: Sure!

src/components/TVChartContainer/SaveLoadAdapter.ts
vipineth: Regarding this I took their sample implementation and used it. I didn't want change because it can have an edge case issue or something.

divhead (approved):

PR #373, opened 2022-12-01
Update translated files

vipineth:
[x] Update translated files
[x] Add French language

---
  - https://app.asana.com/0/0/1203265660786833


src/components/Footer/constants.ts
```diff
@@ -21,17 +22,21 @@ type SocialLink = {
...
-    { text: t`Referral Terms`, link: "/referral-terms" },
-    { text: t`Media Kit`, link: "https://gmxio.gitbook.io/gmx/media-kit", external: true },
+    { text: defineMessage({ message: "Terms and Conditions" }), link: "/terms-and-conditions" },
+    { text: defineMessage({ message: "Referral Terms" }), link: "/referral-terms" },
+    { text: defineMessage({ message: "Media Kit" }), link: "https://gmxio.gitbook.io/gmx/media-kit", external: true },
     // { text: "Jobs", link: "/jobs", isAppLink: true },
   ],
   app: [
-    { text: t`Media Kit`, link: "https://gmxio.gitbook.io/gmx/media-kit", external: true },
+    { text: defineMessage({ message: "Media Kit" }), link: "https://gmxio.gitbook.io/gmx/media-kit", external: true },
```
divhead, line 31: Sorry, why we need to change `t'...'` to defineMessage here? What's the difference?

divhead (dismissed):

src/components/Footer/constants.ts
vipineth: `defineMessage` is used for lazy translation which means the messages doesn't have to be defined at the same code location where they are displayed.

https://lingui.js.org/tutorials/react-patterns.html#lazy-translations

divhead (dismissed):

divhead (approved):

PR #374, opened 2022-12-02
update text for mark price on order list

vipineth:
[x] Update mark price tooltip for order list
[x] Removed `select` from translation 

---
  - https://app.asana.com/0/0/1203418687677798


src/components/Exchange/TradeHistory.js
```diff
@@ -373,16 +374,15 @@ export default function TradeHistory(props) {
           return defaultMsg;
         }
         const longShortDisplay = order.isLong ? t`Long` : t`Short`;
+        const orderTypeText = order.type === INCREASE ? t`Increase` : t`Decrease`;
```
divhead, line 377: Can there be a swap order?

divhead (dismissed):

src/components/Exchange/TradeHistory.js
vipineth: This section is only for `Increase` and `Decrease` orders.

divhead (approved):

PR #377, opened 2022-12-05
add tooltip to button

vipineth:
[x] Add tooltip to swapbox button for Pool Exceeded
[x] Add tooltip for invalid liquidation price.
[x] Refactor Tooltip and TooltipWithPortal to tsx

---
  - https://app.asana.com/0/0/1203406165446373
  - https://app.asana.com/0/0/1203406165446319
  - https://app.asana.com/0/0/1203539208079041


src/components/Exchange/PositionEditor.js
```diff
@@ -461,6 +463,35 @@ export default function PositionEditor(props) {
     [DEPOSIT]: t`Deposit`,
     [WITHDRAW]: t`Withdraw`,
   };
+  const ERROR_TOOLTIP_MSG = {
```
divhead, line 466: 1. Should we move this to ./constants too?
2. Can there be a case where there is an error code that is not in the config? Maybe we need to define a default message for that?

src/components/Exchange/constants.ts
```diff
@@ -0,0 +1,11 @@
+export enum ErrorCodes {
```
divhead, line 1: We recently discussed with G, enums should be called in the singular for consistency: ErrorCodes -> ErrorCode

You can also capitalize field names - ErrorCodes.Buffer

src/components/Exchange/constants.ts
```diff
@@ -0,0 +1,11 @@
...
+  poolExceeded = "POOL_EXCEEDED",
+  buffer = "BUFFER",
+  maxUSDG = "MAX_USDG",
+  invalidLiqPrice = "INVALID_LIQ_PRICE",
+}
+
+export enum ErrorTypes {
```
divhead, line 8: For me personally ErrorType sounds like the type of the error and not how we show it

I would call it ErrorDisplayType - but that's as you like it)

src/components/Tooltip/Tooltip.tsx
```diff
@@ -1,27 +1,38 @@
...
   const onMouseEnter = useCallback(() => {
     if (trigger !== "hover" || IS_TOUCH) return;
-
     if (intervalCloseRef.current) {
       clearInterval(intervalCloseRef.current);
       intervalCloseRef.current = null;
     }
     if (!intervalOpenRef.current) {
-      intervalOpenRef.current = setTimeout(() => {
+      intervalOpenRef.current = window.setTimeout(() => {
```
divhead, line 35: hmm why we need window here?

if typescript shows an error here it's strange because in other places we use just setTimeout (see lib/sleep.ts)

src/components/Tooltip/TooltipWithPortal.tsx
```diff
@@ -1,36 +1,58 @@
...
-export default function TooltipWithPortal(props) {
+type Props = {
+  handle: React.ReactNode;
+  renderContent: () => React.ReactNode;
+  position?: string;
+  trigger?: string;
+  className?: string;
+  disableHandleStyle?: boolean;
+  handleClassName?: string;
+  onDisabled?: boolean;
```
divhead, line 20: 1. mm, onDisabled sounds like an event handler callback

Not sure properly naming here, maybe isHandlerDisabled/disablePointerEvents?

2. I think comment should be here instead of Wrapper
// For onMouseLeave to work on disabled button https://github.com/react-component/tooltip/issues/18#issuecomment-411476678

3. I'm not sure we need a Wrapper component with so little functionality, Wrapper sounds like something global, not just a specific rare case handler. Maybe leaving a conditional div or className event is enough.

src/components/Tooltip/Tooltip.tsx
vipineth: The difference is assigning to a variable with type null | number. I will update type.

src/components/Exchange/PositionEditor.js
vipineth: 1. Other messages can need access to token or position info. Like in the case of SwapBox here (https://github.com/gmx-io/gmx-interface/blob/8db708a53d976708dcd1caccba605927ebbc245a/src/components/Exchange/SwapBox.js#L1818). So I think we can let it be here for now. LMK you thoughts.
2. I have updated the condition which is responsible for rendering the Tooltip

src/components/Exchange/constants.ts
vipineth: Thanks. Updated

src/components/Exchange/constants.ts
vipineth: Updated

src/components/Tooltip/TooltipWithPortal.tsx
vipineth: Sure!

divhead (approved):

PR #386, opened 2022-12-15
update and refactor icons

vipineth:
[x] Refactor reusable icons on multiple places
[x] Change icons on GLPSwap, BuyGMX, Ecosystem etc


src/config/icons.ts
```diff
@@ -0,0 +1,47 @@
...
+  [AVALANCHE]: {
+    network: avalanche,
+    gmx: gmxAvax,
+    glp: glpAvax,
+  },
+  gmx: gmxIcon,
+  glp: glpIcon,
+};
+
+export function getIcons(network, label) {
```
divhead, line 41: 1. i think `getIcon` not `getIcons` ? Because this function returns only one icon?

2. the config structure looks inconsistent because of common glp and gmx icon
3. in some cases we need to specify the size
4. There is CHAIN_ICONS in config/chains i think wee need to store it in one place
5. we don't need to define the type here, because  typescript should infer this automatically

about 2,3,4, 5 as a possible solution with type checking
```
const ICONS = {
  [ARBITRUM]: {
    network: {
      30: "ic_arbitrum_30.svg",
      16: "ic_arbitrum_16.svg",
    },
    glp: {
      30: "ic_arbitrum_glp_30.svg",
    }
  },
  [AVALANCHE]: {
    network: {
      30: "ic_avalanche_30.svg",
      16: "ic_avalanche_16.svg",
    },
    glp: {
      30: "ic_avalanche_glp_30.svg",
    }
  },
  common: {
    glp: {
      30: "ic_glp_30.svg",
    }
  },
};

type Icons = typeof ICONS;

function getIcon<
  TChain extends keyof Icons,
  TName extends keyof Icons[TChain],
  TSize extends keyof Icons[TChain][TName]
>(chainId: TChain, name: TName, size: TSize) {}

const arb = getIcon(ARBITRUM, "network", 30);
const glp = getIcon('common', "glp", 30);
```

divhead: Would wait for comments from @gdev8317

src/config/icons.ts
```diff
@@ -0,0 +1,47 @@
...
+    gmx: string;
+    glp: string;
+  };
+  [AVALANCHE]: {
+    network: string;
+    gmx: string;
+    glp: string;
+  };
+  gmx: string;
+  glp: string;
```
gdev8317, line 23: why do we need separate gmx/glp icons per network?

vipineth: This helps to differentiate between the active network. For better idea you can look at the icon below:

Screenshot 2023-01-10 at 2 38 02 PM

src/config/icons.ts
vipineth: I feel we don't need to have multiple sizes of the icons for `svg` files. Scalability is a huge advantage of using svg. We can have one icon and add the width attributes to the `img` element.

The issue with the sizes is it can't be standard for each icons and situations. Please let me know your thoughts on this.

divhead: Ah, yes, you're right about the SVG, the sizes aren't needed here then

divhead: Just now in the project for some reason we keep several sizes for svg icons, maybe we should make them all the same default size?

src/config/icons.ts
```diff
@@ -0,0 +1,49 @@
...
+    network: avalancheTestnet,
+  },
+  common: {
+    gmx: gmxIcon,
+    glp: glpIcon,
+  },
+};
+
+export function getIcon(chainId: number | "common", label: string) {
+  if (!chainId || !label) return;
```
divhead, line 37: chainId and label are required here, so, we don't need this check?

src/config/icons.ts
```diff
@@ -0,0 +1,49 @@
...
+  if (chainId in ICONS) {
+    if (label in ICONS[chainId]) {
+      return ICONS[chainId][label];
+    }
+  }
+}
+export function getCurrentIcons(chainId: number | "common") {
```
divhead, line 44: Maybe just 'getIcons'? The 'current' sounds a little redundant here for me

divhead (dismissed):

divhead (approved):

PR #388, opened 2022-12-19
Referrals update

vipineth:
[x] Add pagination to referrals page

---
  - https://app.asana.com/0/0/1203222378802875


src/components/Referrals/usePagination.ts
```diff
@@ -0,0 +1,46 @@
+import { useEffect, useState } from "react";
+
+type UsePagination = {
```
divhead, line 3: we don't need to define this type, typescript should infer it automatically
But, if you want, i think the name would be PaginationState

src/components/Buttons/TransparentButton.tsx
```diff
@@ -0,0 +1,5 @@
+function TransparentButton({ onClick, ...rest }) {
```
gdev8317, line 1: probably better to just extend `Button` component with e.g. optional prop `type` with value "transparent" or something like that? considering it differs by className only

src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,69 @@
+import TransparentButton from "components/Buttons/TransparentButton";
+import "./Pagination.css";
+
+type Props = {
+  page: number;
+  pageCount: number;
+  onPageChange: any;
+};
+
+function range(start, end) {
```
gdev8317, line 10: you can use `range` from lodash, no need to define custom one

src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,69 @@
...
+  let length = end - start + 1;
+  return Array.from({ length }, (_, idx) => idx + start);
+}
+
+function getPageNumbers(current, max = 1) {
+  if (max === 2) {
+    return [1, 2];
+  }
+  if (current <= 1) {
```
gdev8317, line 19: how `current` can be less then 1? if it's current page number then the lowest value is 1?

src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,69 @@
...
+    return range(current, current + 2);
```
gdev8317, line 20: what if max = 1? then function will return [1, 2]?

src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  if (current + 1 <= max) {
+    return range(current - 1, current + 1);
+  }
+
+  if (current <= max) {
+    return range(current - 2, current);
+  }
+  return [];
```
gdev8317, line 30: how can that case happen? `current <= max` is always true?

src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,69 @@
...
+}
```
gdev8317, line 31: if i understand correctly the goal is to display previous, current and next pages in UI. i think function can be simplified to:

```
const ret = [];
if (current > 1) {
  ret.push(current - 1};
}
ret.push(current);
if (current < max) {
  ret.push(current + 1);
}
return ret;
```

vipineth: No the requirements was a bit different. So we wanted to have three elements if possible to maintain the gas as you can see in the image below.

Screenshot 2023-01-12 at 11 48 02 AM

The function need to fulfil the following requirements:

if max is 1 return empty
if current is 1 and max is 2 return 1,2
if current is 2 and max is 2 return 1,2
if current is 1 and max is 4 return 1,2,3
if current is 2 and max is 4 return 1,2,3
if current is 3 and max is 4 return 2,3,4
if current is 4 and max is 4 return 2,3,4

The above solution will return 2 numbers in certain situations like current is 1 and max is 4 it will return 1 and 2. But it should return 1, 2, 3.

BTW I have refactored the code to this:

```js
function getPageNumbers(current, max = 1) {
  if (max === 1) return [];
  if (current === 1) {
    return max >= 3 ? [1, 2, 3] : [1, 2];
  } else if (current === max) {
    return max >= 3 ? [current - 2, current - 1, current] : [current - 1, current];
  } else {
    return [current - 1, current, current + 1];
  }
}
```

src/components/Buttons/TransparentButton.tsx
vipineth: We are using `Button` component only as a transparent button in our app as of now.

I have refactored the `Button` component and can add more type to it in future. Thanks 🙏

gdev8317: not sure i understand, i was proposing to remove TransparentButton, not sure it's worth adding separate component TransparentButton since it's only renders Button component with custom className?

what about having only this? 

```
components/Button
  ./Button.jsx
  ./Button.scss
```

or what is the reason to have "Common" and "Buttons" directories for buttons?

src/components/Pagination/Pagination.tsx
gdev8317: okay, got it

src/components/Common/Button.tsx
```diff
@@ -1,24 +1,35 @@
 import { ReactNode } from "react";
 import cx from "classnames";
-import "./Button.css";
+import "./Button.scss";
+
+export enum ButtonType {
+  Transparent,
+}
```
gdev8317, line 7: i think it's okay to have `type ButtonType = "transparent"` so we can use it as <Button type="transparent">...</Button>, importing enum from other places may be bit overhead

vipineth: Sure!

gdev8317 (approved):

PR #390, opened 2022-12-21
show chart position of only related positions

vipineth:
[x] Filter assets to show for chart positions

---
  - https://app.asana.com/0/0/1203592334171373


src/components/Exchange/ExchangeTVChart.js
```diff
@@ -292,9 +299,9 @@ export default function ExchangeTVChart(props) {
           );
         });
       }
-      if (positions && positions.length > 0) {
+      if (currentPositions && currentPositions.length > 0) {
```
divhead, line 302: mm, looks like currentPositions is always array because you return an empty array in useMemo?

divhead (approved):

PR #405, opened 2023-01-15
Make modals standard

vipineth:
[x] Refactor the fields on all the modals on trade page

---
  - https://app.asana.com/0/0/1202881629558494
  - https://app.asana.com/0/0/1203851354727830
  - https://app.asana.com/0/0/1203889330911721


src/components/Exchange/FeesTooltip.tsx
```diff
@@ -0,0 +1,72 @@
...
+      value: fundingFee,
+    });
+  }
+  if (isOpening && swapFee) {
+    feeRows.push({
+      label: t`Swap Fee`,
+      value: swapFee,
+    });
+  }
+  if (positionFee) {
```
divhead, line 25: 1. There seems to be a duplication of conditions?

```
if(!isOpening && swapFee) {...}

if(isOpening && swapFee) {...}
```
From this it is hard to see what the list of commissions will look like

Also, here are a couple of suggestions:
2. move the formatting also in this tooltip, now in PositionSeller/Editor/SwapBox
there is a duplicated formatting code like
```
${formatAmountFree(minExecutionFee, 18, 5)} ${nativeTokenSymbol}}
```
Thus, FeesTooltip component is not fully unified

3. Is it possible to remove the serial push in Rows array? I mean, if the only problem is the different order of opening / closing scenarios, can we just use "openingFee, closingFee" props instead of positionFee and just render them where needed?
Or, alternatively, if there is too much difference between the different Modals / SwapBoxes, it is easier to use this as
```
<FeesTooltip 
 rows={...}
/>
```

src/components/Exchange/FeesTooltip.tsx
```diff
@@ -0,0 +1,72 @@
...
+import ExternalLink from "components/ExternalLink/ExternalLink";
+import StatsTooltipRow from "components/StatsTooltip/StatsTooltipRow";
+import Tooltip from "components/Tooltip/Tooltip";
+
+type Row = {
+  label: string;
+  value: string;
+};
+
+function FeesTooltip({ totalFees, fundingFee, positionFee, swapFee, executionFee, isOpening = true }) {
```
divhead, line 11: It is worth typing Props)

src/components/Exchange/FeesTooltip.tsx
vipineth: I have updated the code based on your suggestions.

src/components/Exchange/FeesTooltip.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+  const feeRows: Row[] = [
+    { label: isOpening ? SWAP_FEE_LABEL : BORROW_FEE_LABEL, value: isOpening ? swapFee : fundingFee },
+    { label: isOpening ? t`Open Fee` : t`Close Fee`, value: positionFee },
+    { label: isOpening ? BORROW_FEE_LABEL : SWAP_FEE_LABEL, value: isOpening ? fundingFee : swapFee },
+    { label: t`Deposit Fee`, value: depositFee },
+    { label: t`Execution Fee`, value: getExecutionFeeStr(chainId, executionFee, executionFeeUSD) },
+  ]
+    .filter((row) => row.value)
+    .map(({ label, value }) => ({ label, value }));
```
divhead, line 61: Looks like useless .map?

src/components/Exchange/FeesTooltip.tsx
```diff
@@ -0,0 +1,85 @@
...
+  depositFee,
+  isOpening = true,
+}: Props) {
+  const { chainId } = useChainId();
+  const SWAP_FEE_LABEL = t`Swap Fee`;
+  const BORROW_FEE_LABEL = t`Borrow Fee`;
+  const executionFee = executionFees?.fee;
+  const executionFeeUSD = executionFees?.feeUSD;
+
+  const feeRows: Row[] = [
```
divhead, line 53: it's a bit confusing with a lot of conditions on isOpening
maybe just:

```
const openingFees = isOpening ? [...] : []
const defaultFees = [...]

const feesRows = [
   ...openingFees
]
```

or as you did before but less conditions:
```
let feesRows: Row[] = [];

if(isOpening) {
 feesRows.push(
   {},
   {},
   {},
)
} else {
 feesRows.push(
   {},
   {},
   {},
)
}

feeRows.push(
  {},
  {}
)

feesRows = feesRows.filter(...)
```

src/components/Exchange/FeesTooltip.tsx
```diff
@@ -0,0 +1,85 @@
...
+  totalFees: string;
+  fundingFee: string;
+  positionFee: string;
+  depositFee: string;
+  swapFee: string;
+  executionFees: { fee: BigNumberish; feeUSD: BigNumberish };
+  isOpening?: boolean;
+};
+
+function FeesTooltip({
```
divhead, line 38: mm, maybe FeesTooltip will accept BigNumbers instead of strings and format fee strings inside?

Then the component will be easier to use and there will be no duplicate logic in the SwapBox, PositionEditor, etc

src/components/Exchange/FeesTooltip.tsx
vipineth: Done!

src/components/Exchange/FeesTooltip.tsx
vipineth: Done!

src/components/Exchange/FeesTooltip.tsx
```diff
@@ -0,0 +1,122 @@
...
+import { BigNumber, BigNumberish } from "ethers";
+import { useChainId } from "lib/chains";
+import { USD_DECIMALS } from "lib/legacy";
+import { formatAmount, formatAmountFree } from "lib/numbers";
+
+type Fee = { label: string; value: string };
+type ExecutionFee = { fee?: BigNumberish; feeUSD?: BigNumberish };
+type FeeType = "open" | "close" | "swap" | "borrow" | "deposit" | "execution";
+
+const FEE_LABELS: Record<FeeType, MessageDescriptor> = {
```
divhead, line 16: not critial, but, i think, what if just wrap these messages to function like:
```
function getFeeLabel(type: FeeType) {
  const labels = {
     close: ...
     open: ...
   }
    
  return labels[type]
}
```
Then it will work when the language is changed? Just to avoid a bit complicated syntax with defineMessage and 18n._

src/components/Exchange/FeesTooltip.tsx
```diff
@@ -0,0 +1,122 @@
...
+  if (!nativeTokenSymbol || !executionFee || !executionFeeUsd) {
+    return "";
+  }
+
+  const formattedExecutionFee = formatAmountFree(executionFee, 18, 4);
+  const formattedExecutionFeeUsd = formatAmount(executionFeeUsd, USD_DECIMALS, 2);
+  return `${formattedExecutionFee} ${nativeTokenSymbol} ($${formattedExecutionFeeUsd})`;
+}
+
+function getFeesInStr(fees: BigNumber | undefined): string {
```
divhead, line 37: maybe just getFeesStr? as in getExecutionFeeStr

vipineth: Fixed and updated!

src/components/Exchange/FeesTooltip.tsx
vipineth: Updated!

divhead (approved):

PR #406, opened 2023-01-15
add scroll to inline tab

vipineth:
[x] Move Chart Positions to settings on smaller screen
[x] Make Tab inline able to scroll on x axis

---
  - https://app.asana.com/0/0/1202552600540639


src/App/App.scss
```diff
@@ -1205,6 +1205,10 @@ button.App-button {
   align-items: center;
 }
 
+.chart-positions-settings {
```
divhead, line 1208: Is there a rule that determines where we place the css classes?

I mean, here ```.chart-positions-settings```
Defined in App.css, although we also have Exchange.css (and I think chart-positions refers to the Exchange page?)

src/domain/common.ts
```diff
@@ -1,5 +1,17 @@
+import { ARBITRUM, AVALANCHE } from "config/chains";
+
 export function get1InchSwapUrl(chainId: number, from?: string, to?: string) {
   const rootUrl = `https://app.1inch.io/#/${chainId}/unified/swap`;
   if (!from && !to) return rootUrl;
   return `${rootUrl}/${from}/${to}`;
 }
+
+export function getLeaderboardLink(chainId) {
```
divhead, line 9: I'm afraid that "common" will turn into a new Helpers.js, because it's a very general name. Now some kind of link building utilities are being put together here, maybe we should call it (domain/routes/utils.ts)? Maybe even put it in config/routes.ts - because there's no special logic / types / hooks but just using parameters with hardcoded urls

src/App/App.scss
vipineth: We don't at this point of time. I generally try to keep the css class as close as possible to the place it is being used.

In this case we have a modal on app which will contain the setting to have orders visible on the chart or not. So I think it is a part of App more than the exchange. The setting is related to the exchange but visible on the app.

src/domain/common.ts
vipineth: Yeah makes sense to put in the config. Will update

divhead (approved):

PR #409, opened 2023-01-20
add error for minimum collateral deposit

vipineth:
[x] Show error when depositing to a position that is above 100x leverage

---
  - https://app.asana.com/0/0/1203668976600227
  - https://app.asana.com/0/0/1203801220464344


src/lib/numbers.ts
```diff
@@ -118,6 +118,20 @@ export const formatAmountFree = (amount: BigNumberish, tokenDecimals: number, di
...
+  displayDecimals?: number,
+  useCommas?: boolean
+) {
+  const amountStr = formatAmount(amount, tokenDecimals, displayDecimals, useCommas);
+  if (amount && bigNumberify(amount)?.gt(0)) {
+    return "$" + amountStr;
+  } else {
+    return "-$" + amountStr.slice(1);
+  }
+}
```
divhead, line 133: Oh, i've added some formstters in the synthetics branch, If you'd like,  you can review this and maybe copy it here to standartize?

https://github.com/gmx-io/gmx-interface/pull/397/files#diff-7fdf1a5171d1d1dbaacd892cb9454006354046c1f481ef39d068a04eed7afb25

src/components/Exchange/PositionEditor.js
```diff
@@ -465,6 +502,33 @@ export default function PositionEditor(props) {
...
+          label={t`Min. Net Value needed is`}
+          value={minimumCollateralRequired && formatAmount(minimumCollateralRequired, USD_DECIMALS, 2, true)}
+          showDollar={true}
+        />
+        {minimumCollateralToDeposit && minimumCollateralToDepositStr && (
+          <StatsTooltipRow
+            label={t`Deposit at least`}
+            value={[
+              formatAmountWithDollarSign(minimumCollateralToDeposit, USD_DECIMALS, 2, true),
+              minimumCollateralToDepositStr,
```
divhead, line 526: not sure, but probably would be better to keep string formatting in the jsx here instead of str. 209, just to keep business logic and layout separately

PR #412, opened 2023-01-28
update token selector search

vipineth:
[x] Added an icon for the search token
[x] Not to autofocus on smaller screen because the keyboard takes lots of space and most of the time user does't even want to search. More space makes it easy to select any token.

---
  - https://app.asana.com/0/0/1203004428193463


src/components/Exchange/TokenSelector.js
```diff
@@ -71,22 +73,27 @@ export default function TokenSelector(props) {
...
               type="text"
               placeholder={t`Search Token`}
               value={searchKeyword}
               onChange={(e) => onSearchKeywordChange(e)}
               onKeyDown={_handleKeyDown}
-              autoFocus
+              autoFocus={!isSmallerScreen}
+              className="Tokenselector-search-input"
+              style={{
+                background: `url(${searchIcon}) 12px center / 20px 20px no-repeat scroll`,
```
divhead, line 90: let's move the styles to css/scss and leave only background-image: url() here?

divhead: it would just be more readable and harder to break this place i think

divhead (approved):

PR #417, opened 2023-02-07
Update netvalue positionlist

vipineth:
[x] Update the default value of `Show PnL after fees` in settings
[x] Update the calculation of pnlafterfees and net value

---
  - https://app.asana.com/0/0/1203801209167772


src/components/Exchange/NetValueTooltip.tsx
```diff
@@ -0,0 +1,64 @@
+import { Trans, t } from "@lingui/macro";
+import StatsTooltipRow from "components/StatsTooltip/StatsTooltipRow";
+import Tooltip from "components/Tooltip/Tooltip";
+import { BigNumberish } from "ethers";
+import { USD_DECIMALS } from "lib/legacy";
+import { formatAmount } from "lib/numbers";
+
+type Props = {
+  position: {
```
gdev8317, line 9: might be better to define Position type somewhere and reuse it to avoid code duplication, do you think we can do it in this PR?

gdev8317: not critical though

gdev8317 (dismissed):

src/components/Exchange/NetValueTooltip.tsx
vipineth: I have added the type. Please have a look.

gdev8317 (approved):

PR #421, opened 2023-02-09
Refactor tradingview

vipineth:
[x] Implement Tradingview Full

---
  - https://app.asana.com/0/0/1201266503318581

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -10,7 +10,7 @@
 [[headers]]
   for = "/*"
   [headers.values]
-    X-Frame-Options = "DENY"
+    X-Frame-Options = "sameorigin"
     X-XSS-Protection = "1; mode=block"
     Referrer-Policy = "no-referrer"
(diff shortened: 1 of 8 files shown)


netlify.toml
```diff
@@ -10,7 +10,7 @@
 [[headers]]
   for = "/*"
   [headers.values]
-    X-Frame-Options = "DENY"
+    X-Frame-Options = "sameorigin"
```
gdev8317, line 13: copy-paste from chat just in case:

> from what i can see netlify.toml doesn't change headers in production actually 🤔 i forgot, what webhosting we're using currently? cloudflare proxies requests to Netlify, Fleek or something else? 

> we may need to edit headers in some other place (and there are some other security headers to add, but better to add them in separate task just in case)

divhead (approved):

divhead (approved):

PR #425, opened 2023-02-16
add button types

vipineth:
[x] Make a standard button component to have all the variants
[x] Refactor all the buttons to use the Button component

---
  - https://app.asana.com/0/0/1203876686973066


src/components/Button/Button.tsx
```diff
@@ -1,37 +1,74 @@
...
+  imgInfo?: {
+    src: string;
+    alt?: string;
+  };
 };
 
-export default function Button({ href, imgSrc, children, className, imgName, onClick, disabled = false }: Props) {
-  let classNames = cx("btn", className);
+export function getClassName(variant: ButtonVariant) {
+  return cx("button", {
```
divhead, line 23: it could be replaced with just cx('button', variant)

src/components/Button/Button.scss
```diff
@@ -1,34 +1,98 @@
...
-    display: inline-flex;
-    align-items: center;
-    justify-content: center;
-    max-height: 2.3rem;
-    max-width: 2.3rem;
+  &.primary-action {
+    background: linear-gradient(90deg, #2d42fc, #2e3dcd);
+    color: #fff;
+    border-radius: 0.3rem;
+    padding: 1.65rem 1.6rem;
```
divhead, line 38: Ideally, also standardize sizes by separate classes, e.g. ".s (small) .m (medium) .l (large) .xl etc.)

Because we can use one variant with different size i think

src/components/Button/Button.tsx
divhead: Also, what I usually encounter when developing UI kits, different alignment of texts and different "display" of the button. 

Usually, it is solved by props align: 'left' | 'center' and display: 'inline' | 'block' (block combines with:100% to take full width, like in SwapBox

So "variant" may only define the theme of the button, while the size and positioning are set by separate props

src/components/Button/Button.tsx
```diff
@@ -1,37 +1,74 @@
...
+  type?: "button" | "submit" | "reset";
+  imgInfo?: {
+    src: string;
+    alt?: string;
+  };
 };
 
-export default function Button({ href, imgSrc, children, className, imgName, onClick, disabled = false }: Props) {
-  let classNames = cx("btn", className);
+export function getClassName(variant: ButtonVariant) {
```
divhead, line 22: mm, do we need to export this function)?

src/styles/Shared.css
```diff
@@ -381,6 +385,10 @@ button.btn-link:hover .logout-icon {
   margin-top: 0 !important;
 }
 
+.mt-md {
```
divhead, line 388: I like this approach because we don't have a lot of "smart" components, css isolation and at the same time a lot of duplicated plain Html with not very clear classes like "Exchange-something".

I think we can make some kind of custom tailwind, with basic classes for flex boxes, indents, etc.

src/styles/Shared.css
```diff
@@ -424,3 +432,7 @@ button.btn-link:hover .logout-icon {
 .nowrap {
   white-space: nowrap;
 }
+
+.gap-right > *:not(:last-child) {
```
divhead, line 436: But this is the kind of thing I would avoid in a shared сss, such a selector as if for one particular case?

src/components/Button/Button.tsx
vipineth: Removed!

src/styles/Shared.css
vipineth: I think this is a common use-case to have margin right when mapping over items or something. As of now it is being used in two components, There are few more place where we have defined a custom class for the same purpose.

src/components/Button/Button.scss
vipineth: Yeah I agree. I think we can keep updating the styles based on the use-case.

src/components/Button/Button.tsx
```diff
@@ -1,37 +1,65 @@
...
+export default function Button({
+  variant,
+  disabled,
+  onClick,
+  children,
+  to,
+  className,
+  imgInfo,
+  type,
+  ...rest
```
divhead, line 31: looks like ...rest will contain nothing, since all the props are distracted?
maybe jsut remove it?
If you wanted to pass all the props for the html button
then the type should be

```
type ButtonProps =  React.HTMLProps<HTMLButtonElement> & {...}
```

vipineth: Thanks, this is a much better way 🙇‍♂️

src/components/Button/Button.tsx
```diff
@@ -1,37 +1,68 @@
+import "./Button.scss";
 import { ReactNode } from "react";
 import cx from "classnames";
-import "./Button.scss";
+import ButtonLink from "./ButtonLink";
+
+type ButtonVariant = "primary" | "primary-action" | "semi-clear" | "clear";
 
-type Props = {
+type ButtonProps = React.HTMLProps<HTMLButtonElement> & {
```
divhead, line 8: mm if you use React.HTMLProps<HTMLButtonElement>, logically it's need to pass all the props: <button {...rest}/> otherwise it's not obvious why some prop will not used by the button but existing in type ButtonProps

divhead (dismissed):

divhead (approved):

PR #433, opened 2023-03-02
Update tradingview ux

vipineth:
[x] Add loading indicator
[x] Cache requested data so it doesn't need to fetch multiple times
[x] Update few UI issues

---
  - https://app.asana.com/0/0/1203632258598534

### DIFF
--- a/public/tradingview-chart.css
+++ b/public/tradingview-chart.css
@@ -43,8 +43,13 @@ html.theme-dark .item-2IihgTnv:not(.isActive-2IihgTnv) {
   color: #b2b5be !important;
 }
 
-html.theme-dark .item-2IihgTnv:hover {
+html.theme-dark .item-2IihgTnv:not(.isActive-2IihgTnv):hover,
+html.theme-dark .container-3Ywm3-oo.selected-3Ywm3-oo {
   background-color: var(--dark-blue-bg) !important;
+}
+
+html.theme-dark .container-3Ywm3-oo.selected-3Ywm3-oo {
+  background-color: #2962ff !important;
 }
 
 html.theme-dark .dialog-2AogBbC7 {
@@ -136,7 +141,7 @@ html.theme-dark .summary-3UYGeClB {
 }
 html.theme-dark .summary-3UYGeClB:hover,
 html.theme-dark .itemRow-1SBAfF8E:not(.active-1SBAfF8E):hover,
-html.theme-dark .container-3Ywm3-oo:not(.disabled-3Ywm3-oo):hover {
+html.theme-dark .container-3Ywm3-oo:not(.disabled-3Ywm3-oo):not(.selected-3Ywm3-oo):hover {
   background-color: var(--dark-blue-bg) !important;
   cursor: pointer;
 }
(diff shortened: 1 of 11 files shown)


public/tradingview-chart.css
```diff
@@ -37,15 +37,21 @@ html.theme-dark .menuWrap-g78rwseC {
...
-html.theme-dark .item-2IihgTnv {
+html.theme-dark .item-2IihgTnv:not(.isActive-2IihgTnv) {
   cursor: pointer;
   background-color: #20243f !important;
   color: #b2b5be !important;
 }
-html.theme-dark .item-2IihgTnv:hover {
+
+html.theme-dark .item-2IihgTnv:not(.isActive-2IihgTnv):hover,
+html.theme-dark .container-3Ywm3-oo.selected-3Ywm3-oo {
```
divhead, line 47: hm, looks like generated classNames, probably will change if tv lib will be updated? Just wondering, not sure it can be fixed easily

src/domain/tradingview/TVDataProvider.ts
```diff
@@ -0,0 +1,183 @@
...
+    return bars.slice(bars.length - countBack, bars.length);
+  }
+
+  async getBars(
+    chainId: number,
+    ticker: string,
+    resolution: string,
+    isStable: boolean,
+    periodParams: PeriodParams,
+    shouldRefetchBars: { current: boolean }
```
divhead, line 103: i think it's better to keep TVDataProvider independent of React things, so it would be just bool variable instead of {current: boolean}

src/domain/tradingview/TVDataProvider.ts
```diff
@@ -0,0 +1,183 @@
...
+  ): Promise<Bar[]> {
+    const barsInfo = this.barsInfo;
+    if (
+      !barsInfo.data.length ||
+      barsInfo.ticker !== ticker ||
+      barsInfo.period !== period ||
+      shouldRefetchBars.current
+    ) {
+      try {
+        const bars = await getTokenChartPrice(chainId, ticker, period);
```
divhead, line 59: i think i will need to override getTokenChartPrice for synthetics here, maybe make it as a class method to?

src/domain/tradingview/TVDataProvider.ts
```diff
@@ -0,0 +1,183 @@
...
+    isStable: boolean,
+    periodParams: PeriodParams,
+    shouldRefetchBars: { current: boolean }
+  ) {
+    const period = SUPPORTED_RESOLUTIONS[resolution];
+    const { countBack } = periodParams;
+
+    try {
+      const bars = isStable
+        ? getStablePriceData(period, countBack)
```
divhead, line 110: not a priority, but G wants to get stable tokens prices for synthetics for from ChainLink instead of hardoced values, so in the future i will probably try to to make getTokenHistoryBars unified for stable and not stable tokens

public/tradingview-chart.css
vipineth: Yes, when the version will be updated it will change. We needed those changes and there is no other way to make that happen. Once tv is stable and will not require changes I will look into this again.

src/domain/tradingview/TVDataProvider.ts
vipineth: Fixed.

PR #448, opened 2023-03-15
fix 'autoRefreshOnNetworkChange' undefined

lokissue:
# Description
According to MetaMask docs, "autoRefreshOnNetworkChange" has been removed:
https://docs.metamask.io/guide/provider-migration.html#handling-the-removal-of-ethereum-autorefreshonnetworkchange

It will crash the app when Metamask is installed, with the error

> Uncaught TypeError: Cannot set properties of undefined (setting 'autoRefreshOnNetworkChange')

## Type of change
[x] Bug fix (non-breaking change which fixes an issue)

## Solution
Put a check if `ethereum.autoRefreshOnNetworkChange` is defined.

### DIFF
--- a/src/App/App.js
+++ b/src/App/App.js
@@ -99,7 +99,7 @@ import { useChainId } from "lib/chains";
 import ExternalLink from "components/ExternalLink/ExternalLink";
 import { isDevelopment } from "config/env";
 
-if ("ethereum" in window) {
+if ("ethereum" in window && typeof window.ethereum.autoRefreshOnNetworkChange !== "undefined") {
   window.ethereum.autoRefreshOnNetworkChange = false;
 }
 


src/App/App.js
```diff
@@ -99,7 +99,7 @@ import { useChainId } from "lib/chains";
 import ExternalLink from "components/ExternalLink/ExternalLink";
 import { isDevelopment } from "config/env";
 
-if ("ethereum" in window) {
+if ("ethereum" in window && typeof window.ethereum.autoRefreshOnNetworkChange !== "undefined") {
```
divhead, line 102: But window.etherium is not undefined, right?

so if I set `window.ethereum.autoRefreshOnNetworkChange = true` it won't crash?

Have you encountered this problem in production?

lokissue: I'm currently experiencing this issue in production which crashes the app, not sure if this issue is caused by upgrading MetaMask to version 10.26.x, 'autoRefreshOnNEtworkChange' has been removed.

So it will cause an uncaught TypeError as the following shown in the console:
> Uncaught TypeError: Cannot set properties of undefined (setting 'autoRefreshOnNetworkChange')

try put a check or try-catch to prevent the crash.

lokissue: > But window.etherium is not undefined, right?
> 
> so if I set `window.ethereum.autoRefreshOnNetworkChange = true` it won't crash?
> 
> Have you encountered this problem in production?

I'm currently experiencing this issue in production which crashes the app, not sure if this issue is caused by upgrading MetaMask to version 10.26.x, 'autoRefreshOnNEtworkChange' has been removed.

So it will cause an uncaught TypeError as the following shown in the console:
> Uncaught TypeError: Cannot set properties of undefined (setting 'autoRefreshOnNetworkChange')

try put a check or try-catch to prevent the crash.

divhead: > > But window.etherium is not undefined, right?
> > so if I set `window.ethereum.autoRefreshOnNetworkChange = true` it won't crash?
> > Have you encountered this problem in production?
> 
> I'm currently experiencing this issue in production which crashes the app, not sure if this issue is caused by upgrading MetaMask to version 10.26.x, 'autoRefreshOnNEtworkChange' has been removed.
> 
> So it will cause an uncaught TypeError as the following shown in the console:
> 
> > Uncaught TypeError: Cannot set properties of undefined (setting 'autoRefreshOnNetworkChange')
> 
> try put a check or try-catch to prevent the crash. 

Oh, thank you for the report in this case. Could you describe a scenario to reproduce it. Oh, thank you for the report in this case. Could you describe a scenario to reproduce it so we can test everything?

I just wanted to point out that this change should not fix the error mentioned
It follows from the error text that `window.etherium` is `undefined`

I think this is because checking `"ethereum" in window` will return `true` even if `window.ethereum === undefined`

So, the check should be a bit different, just `if (window.ethereum)`

lokissue: > Oh, thank you for the report in this case. Could you describe a scenario to reproduce it so we can test everything?

To reproduce it: upgrade MetaMask Extension to version 10.26.1 (I use the Chrome one), and run the app frontend, you may see the crash, sometimes the crash occurs after refreshing the page manually.

Note: Metamask extension version 10.25.0 works fine without the issue.

> Here you can check if your fix helped:
https://deploy-preview-448--superb-tarsier-e2aa29.netlify.app/#/trade

The fix helps.

> I just wanted to point out that this change should not fix the error mentioned
It follows from the error text that window.ethereum is undefined

In my case, `window.ethereum` is defined, `window.ethereum.autoRefreshOnNetworkChange` is `undefined`
So, I add a check for `autoRefreshOnNetworkChange` property

divhead: Just to clarify - do you have only Metamask installed or any other wallet?

divhead: Regarding this Error:
> Uncaught TypeError: Cannot set properties of undefined (setting 'autoRefreshOnNetworkChange')

It says that the object on which we are trying to set the property is undefined, not the specific property
in js you can do something like `window.foo = true` without any errors but `window.bra.foo = true` will throw the similar error as you mentioned

Thank you for highlight the issue

i’ve re-created PR with additional check for `window.ethereum` and gpg signature for the commit (it’s required by security settings to merge). Now trying to rerpoduce the issue and validate that this is fixed 

The issue is reproducing for you here?
https://deploy-preview-450--superb-tarsier-e2aa29.netlify.app/#/trade

lokissue: > Just to clarify - do you have only Metamask installed or any other wallet?

Well, I just found out that using MetaMask 10.26.1 together with Phantom wallet installed will cause the issue.
Just MetaMask 10.26.1 itself is OK.

Do you think is still worth putting a check if someone is in a mixed environment?

divhead: Thank you, yes, it reproduced with the Phantom wallet. 

> Do you think is still worth putting a check for someone who is in a mixed environment?
Yes, fix is needed as otherwise the service is not available for these users

It's already deployed to the prod, please report if you encounter anything else similar

I closing the PR since i was fixed https://github.com/gmx-io/gmx-interface/pull/450

PR #469, opened 2023-04-10
Rearrange positionlist mobile

vipineth:
[x] Re-arrange position list labels on smaller screens

---
  - https://app.asana.com/0/0/1203836410751133
  - https://app.asana.com/0/0/1204354288311140
  - https://app.asana.com/0/0/1204364376625308


src/App/App.scss
```diff
@@ -490,6 +490,10 @@ button.App-connect-wallet:hover {
   font-weight: bold;
 }
 
+.Page-title img {
```
divhead, line 493: maybe specify a class for this image, like Page-title-image?

I would avoid such general styles because it leads to non-obvious behavior

divhead (approved):

divhead (approved):

PR #471, opened 2023-04-17
convert allowed slippage to input

vipineth:
[x] Convert allowed slippage to input

### DIFF
--- a/src/components/Exchange/SlippageInput.tsx
+++ b/src/components/Exchange/SlippageInput.tsx
@@ -1,0 +1,35 @@
+import { formatAmount } from "lib/numbers";
+import "./SlippageInput.scss";
+import { useState } from "react";
+import cx from "classnames";
+
+const MAX_SLIPPAGE = 500;
+
+export default function SlippageInput({ placeholderValue, update }) {
+  const placeHolderValue = formatAmount(placeholderValue, 2, 2);
+  const [slippageInput, setSlippageInput] = useState("");
+  const [slippageError, setSlippageError] = useState("");
+  function handleChange(value: string) {
+    setSlippageInput(value);
+    setSlippageError("");
+    const parsed = Math.floor(Number.parseFloat(value) * 100);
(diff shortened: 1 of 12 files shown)


src/components/Exchange/SlippageInput.tsx
```diff
@@ -0,0 +1,91 @@
...
+function getSlippageText(value: number) {
+  return formatAmount(value, 2, 2).replace(/0+$/, "");
+}
+
+type Props = {
+  setAllowedSlippage: (value: number) => void;
+  defaultSlippage: number;
+};
+
+export default function SlippageInput({ setAllowedSlippage, defaultSlippage }: Props) {
```
divhead, line 20: i think we will use this component in V2 also, could it be moved to /compnents/SlippageInput?

src/components/Exchange/SlippageInput.tsx
```diff
@@ -0,0 +1,91 @@
...
+      setAllowedSlippage(defaultSlippage);
+      return;
+    }
+
+    const parsedValue = Math.round(Number.parseFloat(value) * 100);
+    if (Number.isNaN(parsedValue)) {
+      return;
+    }
+
+    if (parsedValue >= HIGH_SLIPPAGE) {
```
divhead, line 40: i think setState functions should be wrapped to useEffect to avoid endless updates just in case

src/components/Tooltip/Tooltip.css
```diff
@@ -135,6 +135,10 @@
   flex: 0 0 100%;
 }
 
+.high-z-index {
```
divhead, line 138: maybe rename it to "z-index-1001"? just to be more clear, i think we may have several z-index classes to handle layers

src/components/Exchange/SlippageInput.tsx
vipineth: Done!

src/components/Tooltip/Tooltip.css
vipineth: Done!

PR #484, opened 2023-04-30
add percentage to position close

vipineth:
[x] Add percentage to position close

---
  - https://app.asana.com/0/0/1204479712773919


src/config/ui.ts
```diff
@@ -2,3 +2,5 @@ export const TRIGGER_PREFIX_ABOVE = ">";
 export const TRIGGER_PREFIX_BELOW = "<";
 
 export const TOAST_AUTO_CLOSE_TIME = 7000;
+
+export const POSITION_CLOSE_SUGGESTION_LISTS = [10, 25, 50, 75];
```
divhead, line 6: A small thing - maybe rename it to PERCENTAGE_SUGGESTIONS as it not related to the close size directly and, probably we don't need to store it as a global config since it used only in one place

(not critial)

divhead:
I think we need the same logic in OrderEditor for trigger orders?

src/config/ui.ts
vipineth: Updated!

PR #501, opened 2023-05-26
Open interest to dashboard

vipineth:
[x] Add Open Interest and Unique Users to dashboard

---
  - https://app.asana.com/0/0/1204314306432024


src/lib/graphqlFetcher.ts
```diff
@@ -0,0 +1,23 @@
...
+      body: JSON.stringify({ query, variables }),
+      headers: { "Content-type": "application/json" },
+      method: "POST",
+    });
+
+    if (!response.ok) {
+      throw new Error(`HTTP error: ${response.status}`);
+    }
+
+    const { data } = await response.json();
```
divhead, line 17: It's a good idea to put it in a separate function, but I don't really understand why not to use gql client?

also, should this function throw the error and not just legitimize? otherwise we won't get an "error" in useSwr?

divhead (approved):

PR #503, opened 2023-05-30
add threshold numbers

vipineth:
[x] Add min and max threshold to number formats

---
  - https://app.asana.com/0/0/1204173577512419

### DIFF
--- a/src/lib/numbers.ts
+++ b/src/lib/numbers.ts
@@ -16,6 +16,30 @@ export function expandDecimals(n: BigNumberish, decimals: number): BigNumber {
   return bigNumberify(n).mul(bigNumberify(10).pow(decimals));
 }
 
+const MAX_EXCEEDING_THRESHOLD = "1000000000";
+const MIN_EXCEEDING_THRESHOLD = "0.01";
+
+function checkPriceThresholdExceeding(value: BigNumber, tokenDecimals: number) {
+  const maxExceeding = expandDecimals(MAX_EXCEEDING_THRESHOLD, tokenDecimals);
+  const minExceeding = ethers.utils.parseUnits(MIN_EXCEEDING_THRESHOLD, tokenDecimals);
+
+  if (!value || value.eq(0)) return;
+
+  if (value.gte(maxExceeding)) {
+    return {
+      symbol: ">",
+      value: maxExceeding,
+    };
+  }
+
+  if (value.lte(minExceeding)) {
+    return {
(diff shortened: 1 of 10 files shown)


src/lib/numbers.ts
```diff
@@ -197,12 +226,13 @@ export function formatTokenAmount(
       return undefined;
     }
   }
+  const exceedingInfo = checkPriceThresholdExceeding(amount, tokenDecimals);
 
   const formattedAmount = showAllSignificant
-    ? formatAmountFree(amount, tokenDecimals, tokenDecimals)
-    : formatAmount(amount, tokenDecimals, displayDecimals, useCommas);
+    ? formatAmountFree(exceedingInfo?.value ?? amount, tokenDecimals, tokenDecimals)
```
divhead, line 232: i think formatAmountFee should be unchanged, because it is usually used with the flag "showAllSignificant" to just show, for example, values like 0.00123222 in inputs

src/lib/numbers.ts
```diff
@@ -16,6 +16,30 @@ export function expandDecimals(n: BigNumberish, decimals: number): BigNumber {
   return bigNumberify(n).mul(bigNumberify(10).pow(decimals));
 }
 
+const MAX_EXCEEDING_THRESHOLD = "1000000000";
+const MIN_EXCEEDING_THRESHOLD = "0.01";
+
+function checkPriceThresholdExceeding(value: BigNumber, tokenDecimals: number) {
```
divhead, line 22: not sure about the naming - “check” looks like something that returns boolean

Maybe we could call it like “getLimitedDisplay” or something else

src/lib/numbers.ts
```diff
@@ -16,6 +16,30 @@ export function expandDecimals(n: BigNumberish, decimals: number): BigNumber {
...
+  const maxExceeding = expandDecimals(MAX_EXCEEDING_THRESHOLD, tokenDecimals);
+  const minExceeding = ethers.utils.parseUnits(MIN_EXCEEDING_THRESHOLD, tokenDecimals);
+
+  if (!value || value.eq(0)) return;
```
divhead, line 26: mm, maybe return original value and empty "" symbol if value doesn't exceed threshold? to avoid undefined and duplicating conditional using like exceedingInfo?.symbol ?? ""

src/lib/numbers.ts
vipineth: Updated!

PR #507, opened 2023-06-02
upgrade react scripts

vipineth:
[x] Upgrade react-scripts to v5

---
  - https://app.asana.com/0/0/1204742471949025


.env
```diff
@@ -0,0 +1 @@
+GENERATE_SOURCEMAP=false
```
gdev8317, line 1: does it mean local server and/or test stands won't have source maps?

src/index.d.ts
```diff
@@ -0,0 +1 @@
+// #
```
gdev8317, line 1: do we need this?

.env
vipineth: This is set for production. 

From docs:

When set to false, source maps are not generated for a production build. This solves out of memory (OOM) issues on some smaller machines.

https://create-react-app.dev/docs/advanced-configuration/

src/index.d.ts
vipineth: I was getting an error and it was suggested to have a file named index.d.ts. So I added that. I will look into the issue properly.

src/components/Exchange/FeesTooltip.tsx
```diff
@@ -22,8 +22,7 @@ function getFeeLabel(type: FeeType) {
     deposit: t`Deposit Fee`,
     execution: t`Execution Fee`,
   };
-
-  return i18n._(labels[type]);
+  return i18n._(/*i18n*/ labels[type]);
```
divhead, line 25: mm, is /*i18n*/ some kind of magic comment for lingui? Or can it be deleted?

Btw, probably this syntax is not needed as getFeeLabel is called inside component during render so language should update correctly with just 

```
return labels[type]?
```

src/components/Footer/constants.ts
```diff
@@ -22,13 +21,13 @@ type SocialLink = {
 
 export const FOOTER_LINKS: { home: Link[]; app: Link[] } = {
   home: [
-    { text: defineMessage({ message: "Terms and Conditions" }), link: "/terms-and-conditions" },
-    { text: defineMessage({ message: "Referral Terms" }), link: "/referral-terms" },
-    { text: defineMessage({ message: "Media Kit" }), link: "https://gmxio.gitbook.io/gmx/media-kit", external: true },
+    { text: t`Terms and Conditions`, link: "/terms-and-conditions" },
```
divhead, line 24: but here FOOTER_LINKS should be inside getFooterLinks function?
Otherwise they are only initialized once?

src/components/Exchange/FeesTooltip.tsx
vipineth: Yeah, I tested it and works well. I have removed it now.

src/components/Footer/constants.ts
vipineth: Moved it. Thanks for noticing.

divhead (approved):

PR #520, opened 2023-06-16
add warning for custom discount

vipineth:
---
  - https://app.asana.com/0/0/1204834612413163

### DIFF
--- a/src/components/Referrals/referralsHelper.js
+++ b/src/components/Referrals/referralsHelper.js
@@ -91,7 +91,6 @@ export function getDiscountSharePercentage(tierId, discountBps) {
 }
 
 export function getRebateSharePercentage(tierId, discountBps) {
-  console.log({ tierId, discountBps });
   if (!tierId) return;
   if (!discountBps || discountBps?.eq(0)) return tierRebateInfo[tierId];
   const decimals = 4;
--- a/src/components/Referrals/referralsHelper.js
+++ b/src/components/Referrals/referralsHelper.js
@@ -8,13 +8,18 @@ import {
 } from "lib/legacy";
 import { encodeReferralCode, getReferralCodeOwner } from "domain/referrals";
 import { ARBITRUM, AVALANCHE } from "config/chains";
-import { bigNumberify, formatAmount, trimZeroDecimals } from "lib/numbers";
+import { bigNumberify, formatAmount } from "lib/numbers";
 import { t } from "@lingui/macro";
 import { getRootUrl } from "lib/url";
 
 export const REFERRAL_CODE_REGEX = /^\w+$/; // only number, string and underscore is allowed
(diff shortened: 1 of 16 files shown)


src/components/Referrals/referralsHelper.js
```diff
@@ -13,6 +14,12 @@ import { getRootUrl } from "lib/url";
 
 export const REFERRAL_CODE_REGEX = /^\w+$/; // only number, string and underscore is allowed
 export const REGEX_VERIFY_BYTES32 = /^0x[0-9a-f]{64}$/;
+const MAX_DISCOUNT_SHARE = 10000; // 100%
+
+export function removeTrailingZeros(amount) {
+  const regex = /\.?0+$/;
+  return amount.replace(regex, "");
```
gdev8317, line 21: would `Number(amount)` work in this case? to make it a bit simpler

src/components/Referrals/referralsHelper.js
```diff
@@ -69,6 +76,39 @@ export const tierDiscountInfo = {
...
+  if (!discountShare || discountShare?.eq(0)) return tierRebateInfo[tierId];
+  const decimals = 4;
+
+  const totalDiscount = totalRebateInfo[tierId];
+  const discountPercentage = bigNumberify(totalDiscount)
+    .mul(MAX_DISCOUNT_SHARE - discountShare)
+    .mul(Math.pow(10, decimals))
+    .div(BASIS_POINTS_DIVISOR);
+  return removeTrailingZeros(formatAmount(discountPercentage, decimals, 3, true));
+}
```
gdev8317, line 110: these two functions are almost identical, the only difference is default value returned and first multiplier for `totalDiscount`). i think it would be better to refactor it into single function

src/components/Referrals/referralsHelper.js
```diff
@@ -13,6 +14,12 @@ import { getRootUrl } from "lib/url";
 
 export const REFERRAL_CODE_REGEX = /^\w+$/; // only number, string and underscore is allowed
 export const REGEX_VERIFY_BYTES32 = /^0x[0-9a-f]{64}$/;
+const MAX_DISCOUNT_SHARE = 10000; // 100%
```
gdev8317, line 17: i think it's okay to use BASIS_POINTS_DIVISOR? discount/rebates are measured in basis points

src/components/Referrals/referralsHelper.js
vipineth: Refactored!

src/components/Referrals/referralsHelper.js
vipineth: Refactored!

src/components/Referrals/referralsHelper.js
vipineth: Refactored!

PR #545, opened 2023-07-10
Synthetics leaderboard

ullin-oi:
---
  - https://app.asana.com/0/0/1202871855191662


package.json
```diff
@@ -5,6 +5,7 @@
   "dependencies": {
     "@apollo/client": "3.5.6",
     "@davatar/react": "1.8.1",
+    "@ensdomains/ens-avatar": "^0.3.3",
```
divhead, line 8: let's fix versions to avoid unexpected breaking changes?

src/components/AddressView/index.css
```diff
@@ -0,0 +1,22 @@
+.trader-account-label {
+  display: flex;
```
divhead, line 2: for now we are using structure like AddressView/AddressView.css

I understand some benefits of using index files, but, so far would probably better to keep code style as in the rest project

src/components/AddressView/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+  lengths,
+  maxLength,
+}: AddressViewProps) {
+  const trader = ensName || address;
+  const strLength = (breakpoint && lengths && lengths[breakpoint]) || maxLength;
+
+  return (
+    <Link className="trader-account-label" to={ `/actions/v2/${address}` }>
+      {
+        avatarUrl ? (
```
divhead, line 32: seems as you don't have prettier configured? formatting a bit different

src/components/GlobalLeaderboards/Context.tsx
```diff
@@ -0,0 +1,213 @@
...
+  pSorted?: boolean;
+  aSorted?: boolean;
+  pReturned?: boolean;
+  aReturned?: boolean;
+  reported?: boolean;
+}
+
+export const useLeaderboardContext = () => useContext(LeaderboardContext);
+
+export const LeaderboardContextProvider: FC<PropsWithChildren> = ({ children }) => {
```
divhead, line 50: why just not use function LeaderboardContextProvider({children}: PropsWithChildren) ?

i mean, "FC" type is required?

src/components/GlobalLeaderboards/Context.tsx
```diff
@@ -0,0 +1,213 @@
...
+  const [positionsOrderBy, setPositionsOrderBy] = useState<keyof TopPositionsRow>("unrealizedPnl");
+  const [positionsOrderDirection, setPositionsOrderDirection] = useState<number>(1);
+  const topPositionsUnordered = useTopPositions();
+  if (
+    topPositionsUnordered &&
+    topPositionsUnordered.data &&
+    topPositionsUnordered.data.length &&
+    p.current && !p.current.positions
+  ) {
+    p.current("topPositionsUnordered");
```
divhead, line 68: Just to clarify, is Profiler will removed at the prod?

src/components/GlobalLeaderboards/Context.tsx
```diff
@@ -0,0 +1,213 @@
...
+  if (
+    topAccountsUnordered &&
+    topAccountsUnordered.data &&
+    topAccountsUnordered.data.length &&
+    p.current && !p.current.accounts
+  ) {
+    p.current("topAccountsUnordered");
+    p.current.accounts = true;
+  }
+  const [topPositions, setTopPositions] = useState<Array<TopPositionsRow>>([]);
```
divhead, line 81: why just don't use TopPositionsRow[] ?

src/domain/synthetics/leaderboards/usePositionsInfo.ts
```diff
@@ -0,0 +1,276 @@
...
+export function usePositionsInfo(
+  chainId: number,
+  positionsHash: string,
+  positionKeys: string[],
+  marketPrices: ContractMarketPrices[],
+): PositionsResult {
+  const { minCollateralUsd } = usePositionsConstants(chainId);
+  const { marketsInfoData, tokensData, pricesUpdatedAt } = useMarketsInfo(chainId);
+  const [positions, setPositions] = useState<PositionsInfoData>();
+  const positionsData = useMulticall(chainId, "usePositionsData", {
```
divhead, line 228: we have different usePositionsData multicall hook. the name "usePositionsData" is a part of SWR key. it would be better to specify unique name to avoid possible collisions

src/domain/synthetics/leaderboards/usePositionsInfo.ts
```diff
@@ -0,0 +1,276 @@
...
+    }
+  // The dependencies below derive from the values used within the effect callback
+  // eslint-disable-next-line
+  }, [
+    positionsHash,
+    pricesUpdatedAt,
+    hasData,
+  ]);
+
+  const isLoading = !(positions && marketsInfoData && tokensData && minCollateralUsd);
```
divhead, line 273: Maybe it's my personal taste, but,  wouldn't the use of (||) more clear?

```
 !positions || !marketsInfoData || !tokensData ...
```

src/components/GlobalLeaderboards/Context.tsx
```diff
@@ -0,0 +1,213 @@
...
+  const topPositionsHeaderClick = useCallback((key: keyof TopPositionsRow) => () => {
+    if (key === positionsOrderBy) {
+      setPositionsOrderDirection((d: number) => -1 * d);
+    } else {
+      setPositionsOrderBy(key);
+      setPositionsOrderDirection(1);
+    }
+  }, [positionsOrderBy]);
+
+  const nextPositionsHash = utils.sha256(topPositionsUnordered.data ? topPositionsUnordered.data.map(a => (
```
divhead, line 106: 1. Could you please explain the purpose of using hash there?
2. What the perfomance of the hash function, probably we should use it inside useMemo?

src/components/GlobalLeaderboards/Context.tsx
```diff
@@ -0,0 +1,213 @@
...
+    a[positionsOrderBy]!.toString().split("").map((c: string) => c.charCodeAt(0))
+  )).flat() : []);
+
+  if (positionsHash !== nextPositionsHash) {
+    setPositionsHash(nextPositionsHash);
+  }
+
+  useEffect(() => {
```
divhead, line 114: For better readability, would be good to have names for useEffect's callback

e.g.
```
useEffect(function updateTopPositionsEff() {})
```

src/components/GlobalLeaderboards/Context.tsx
```diff
@@ -0,0 +1,213 @@
...
+      setPositionsOrderBy(key);
+      setPositionsOrderDirection(1);
+    }
+  }, [positionsOrderBy]);
+
+  const nextPositionsHash = utils.sha256(topPositionsUnordered.data ? topPositionsUnordered.data.map(a => (
+    a[positionsOrderBy]!.toString().split("").map((c: string) => c.charCodeAt(0))
+  )).flat() : []);
+
+  if (positionsHash !== nextPositionsHash) {
```
divhead, line 110: probably we should move it to useEffect it's just not usual when we have setState outside of callBacks / useEffects

src/components/GlobalLeaderboards/Context.tsx
```diff
@@ -0,0 +1,213 @@
...
+  }
+
+  useEffect(() => {
+    if (!topPositionsUnordered.data) {
+      return;
+    }
+
+    const data = [...topPositionsUnordered.data].sort((a, b) => {
+      const key = positionsOrderBy;
+      if (a[key] instanceof BigNumber && b[key] instanceof BigNumber) {
```
divhead, line 121: BigNumber has a method isBigNumber which also works as a type guard

src/components/GlobalLeaderboards/Context.tsx
```diff
@@ -0,0 +1,213 @@
...
+        return 1;
+      }
+    }).map((p, i) => ({ ...p, rank: i }));
+
+    setTopPositions(data);
+    if (p.current && !p.current.pSorted) {
+      p.current("sorted top positions");
+      p.current.pSorted = true;
+    }
+  // eslint-disable-next-line react-hooks/exhaustive-deps
```
divhead, line 136: honestly i would avoid disabling eslint in hook deps if possible, because there could be unobvious bugs after updating the code

src/components/GlobalLeaderboards/Context.tsx
```diff
@@ -0,0 +1,213 @@
...
+        return accountsOrderDirection * ((a[key] as BigNumber).gt(b[key] as BigNumber) ? -1 : 1);
+      } else {
+        // TODO: remove log
+        // eslint-disable-next-line no-console
+        console.warn("Not a BigNumber", {key, a, b});
+        return 1;
+      }
+    });
+
+    data.forEach((a, i) => a.rank = i);
```
divhead, line 165: looks stylish but a bit not clear that you modify some field here, actually entries in topAccountsUnordered will be modified which comes from other hook, if i'm not mistaken. 

i think .map would be safer

src/components/GlobalLeaderboards/Context.tsx
```diff
@@ -0,0 +1,213 @@
...
+  LeaderboardContextType,
+  useTopAccounts,
+  useTopPositions,
+  TopAccountsRow,
+  TopPositionsRow,
+  Profiler,
+} from "domain/synthetics/leaderboards";
+
+export const LeaderboardContext = createContext<LeaderboardContextType>({
+  chainId: 0,
```
divhead, line 25: we have simple useChainId hook, probably we don't need to store chainId in the context

src/components/GlobalLeaderboards/Context.tsx
```diff
@@ -0,0 +1,213 @@
...
+    p.current("returning top accounts");
+    p.current.pReturned = true;
+  }
+
+  if (topAccounts && topAccounts.length && topPositions && topPositions.length &&!p.current.reported) {
+    p.current("returning all data");
+    p.current.report();
+    p.current.reported = true;
+  }
+  const context = {
```
divhead, line 190: i this case all context consumers will re-render every time at every Provider state update

i think it more safer to wrap it into useMemo

src/components/GlobalLeaderboards/TopAccounts.tsx
```diff
@@ -0,0 +1,154 @@
...
+          () => (
+            <div>
+              <p>{ `${ t`RPnL` }: ${ formatUsd(s.rPnl) }` }</p>
+              <p>{ `${ t`UPnL` }: ${ formatUsd(s.uPnl) }` }</p>
+            </div>
+          )
+        }
+      />
+    ),
+    className: classnames(
```
divhead, line 50: why just not use cx shorthand?

src/components/GlobalLeaderboards/TopAccounts.tsx
```diff
@@ -0,0 +1,154 @@
...
+});
+
+export default function TopAccounts() {
+  const perPage = 15;
+  const [page, setPage] = useState(1);
+  const [search, setSearch] = useState("");
+  const term = useDebounce(search, 300);
+  const { topAccounts, topAccountsHeaderClick } = useLeaderboardContext();
+  const { isLoading, error } = topAccounts;
+  const filteredStats = topAccounts.data.filter(a => a.account.indexOf(term.toLowerCase()) >= 0);
```
divhead, line 96: hm, why you define isLoading and error as separated variables but not define data?)

src/domain/synthetics/leaderboards/usePositionsInfo.ts
gdev8317: agree it is probably a matter of taste, i think for me personally it's a bit easier to read `!positions || !marketsInfoData ...` too 🤔

src/components/TableFilterSearch/index.css
```diff
@@ -0,0 +1,50 @@
+.filter-search {
```
gdev8317, line 1: not sure if we strictly follow this convention, but we often use the same name for css class and component (at first glance we do in v2 components)

can we use `.TableFilterSearch` here?

src/components/GlobalLeaderboards/TopPositions.tsx
```diff
@@ -0,0 +1,141 @@
...
+import { getPriceDecimals } from "config/tokens";
+import { BigNumber } from "ethers";
+import { MarketInfo } from "domain/synthetics/markets";
+
+const formatPriceDecimals = (x: BigNumber, market: MarketInfo, chainId: number) => formatAmount(
+  x,
+  USD_DECIMALS,
+  getPriceDecimals(chainId, market.indexToken.symbol),
+  true,
+).replace(/^(-?)/, "$1$$") || ""
```
gdev8317, line 24: i think `formatUsd` does something similar, could it be used instead?

src/components/Table/index.tsx
```diff
@@ -0,0 +1,90 @@
...
+            )
+          )
+        }
+      </tbody>
+    </table>
+  )
+};
+
+const TableCell = ({ data, breakpoint }: TableCellProps) => {
+  const isObject = data && typeof data === "object";
```
gdev8317, line 80: would it be safer to use lodash's isObject? for example array and BigNumber are objects

gdev8317: i guess checking that `data` is truthy is not necessary, `type of data === "object"` is sufficient?

src/components/TableFilterSearch/index.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+export default function TableFilterSearch({value, onInput = () => {}, label, className = ""}: Props = {}) {
+  return (
+    <div className="filter-search">
+      <input
+        type="text"
+        placeholder={ label }
+        value={ value }
+        onChange={ (e) => onInput(e.target.value) }
+        onPaste={ (e) => onInput((e.target as HTMLInputElement).value) }
```
gdev8317, line 20: i think we can use `input` event instead of change/paste? or is there are reason to handle paste explicitly?

src/components/TableFilterSearch/index.tsx
```diff
@@ -0,0 +1,27 @@
...
+        onKeyDown={ () => undefined }
```
gdev8317, line 21: could you explain why is this needed?

src/domain/stats/useFeesSummary.js
```diff
@@ -1,9 +1,14 @@
 import useSWR from "swr";
-import { ARBITRUM, AVALANCHE, AVALANCHE_FUJI } from "config/chains";
+import { isDevelopment } from "config/env";
+import { ARBITRUM, ARBITRUM_GOERLI, AVALANCHE, AVALANCHE_FUJI } from "config/chains";
 import { getServerUrl } from "config/backend";
 import { arrayURLFetcher } from "lib/fetcher";
 const ACTIVE_CHAIN_IDS = [ARBITRUM, AVALANCHE, AVALANCHE_FUJI];
 
+if (isDevelopment()) {
+  ACTIVE_CHAIN_IDS.push(ARBITRUM_GOERLI);
```
gdev8317, line 9: should it also include AVALANCHE_FUJI?

src/domain/synthetics/leaderboards/index.ts
```diff
@@ -0,0 +1,10 @@
+export * from "./types";
+export * from "./useTopAccounts";
+export * from "./useTopPositions";
+export * from "./useAccountPerf";
+export * from "./useAccountOpenPositions";
+export * from "./useBatchProvider";
+export * from "./usePositionsInfo";
+// export * from "./useEnsBatchLookup";
```
gdev8317, line 8: is it okay to remove this line?

src/domain/synthetics/leaderboards/queries/queryAccountOpenPositions.gql
```diff
@@ -0,0 +1,29 @@
+query GetAccountOpenPositions(
```
gdev8317, line 1: maybe better "QueryAccountOpenPositions" to match filename?

src/domain/synthetics/leaderboards/useAccountOpenPositions.ts
```diff
@@ -0,0 +1,212 @@
...
+import { ContractMarketPrices, getContractMarketPrices, useMarketsInfo } from "../markets";
+import { convertToUsd } from "../tokens";
+// import { useChainId } from "lib/chains";
+import { usePositionsInfo } from "./usePositionsInfo";
+import { PositionsInfoData, getPositionKey } from "../positions";
+import { useEffect, useState } from "react";
+import { expandDecimals } from "lib/numbers";
+import { USD_DECIMALS } from "lib/legacy";
+import { AVALANCHE } from "config/chains";
+// import { useEnsBatchLookup } from "./useEnsBatchLookup";
```
gdev8317, line 16: commented lines can be removed?

src/domain/synthetics/leaderboards/useAccountOpenPositions.ts
```diff
@@ -0,0 +1,212 @@
...
+
+const fetchAccountOpenPositionsPage = async (
+  first: number,
+  skip: number,
+  orderBy: string = "sizeInUsd",
+  orderDirection: "asc" | "desc" = "desc",
+): Promise<Array<AccountOpenPositionJson>> => {
+  const res = await graph.query<{ accountOpenPositions: Array<AccountOpenPositionJson> }>({
```
gdev8317, line 24: i think we usually use `AccountOpenPositionJson[]` instead of `Array<AccountOpenPositionJson>`

gdev8317: just to confirm, does it load positions for specific account(s)? at first glance it loads all positions. i think using "account" in names here could be a bit confusing. for example we there are helpers in Reader contract like `getAccountPositions` that returns positions for account.

src/domain/synthetics/leaderboards/useAccountOpenPositions.ts
```diff
@@ -0,0 +1,212 @@
...
+  const [positionsHash, setPositionsHash] = useState<string>("");
+  const [keys, setKeys] = useState<string[]>([]);
+  const [prices, setPrices] = useState<ContractMarketPrices[]>([]);
+  const [data, setData] = useState<AccountOpenPosition[]>([]);
+
+  let positionKeys: number[] = [];
+
+  for (const { id } of positions.data || []) {
+    for (const c of id) {
+      positionKeys.push(c.charCodeAt(0));
```
gdev8317, line 152: could you explain this logic please?

src/domain/synthetics/leaderboards/useAccountOpenPositions.ts
```diff
@@ -0,0 +1,212 @@
...
+    }
+  }
+
+  const hash: string = utils.sha256(positionKeys);
+
+  if (hash !== positionsHash) {
+    setPositionsHash(hash);
```
gdev8317, line 159: should it be inside useEffect?

src/components/GlobalLeaderboards/Context.tsx
```diff
@@ -0,0 +1,213 @@
...
+
+export const LeaderboardContextProvider: FC<PropsWithChildren> = ({ children }) => {
+  const p = useRef<ProfilerRef>();
+  if (!p.current) {
+    p.current = Profiler();
+  }
+  const { chainId } = useChainId();
+  const [period, setPeriod] = useState<PerfPeriod>(PerfPeriod.TOTAL);
+  const [accountsOrderBy, setAccountsOrderBy] = useState<keyof TopAccountsRow>("absPnl");
+  const [accountsOrderDirection, setAccountsOrderDirection] = useState<number>(1);
```
gdev8317, line 58: 1 stands for asc or desc? would it be more clear to have boolean instead of number and variable like `accountsOrderDesc = true | false`?

src/components/GlobalLeaderboards/Context.tsx
```diff
@@ -0,0 +1,213 @@
...
+    p.current && !p.current.positions
+  ) {
+    p.current("topPositionsUnordered");
+    p.current.positions = true;
+  }
+  const topAccountsUnordered = useTopAccounts(period);
+  if (
+    topAccountsUnordered &&
+    topAccountsUnordered.data &&
+    topAccountsUnordered.data.length &&
```
gdev8317, line 75: i think it could be `topAccountsUnordered?.data?.length` for simplicity

src/components/GlobalLeaderboards/Context.tsx
```diff
@@ -0,0 +1,213 @@
...
+
+    setTopPositions(data);
+    if (p.current && !p.current.pSorted) {
+      p.current("sorted top positions");
+      p.current.pSorted = true;
+    }
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [positionsHash, positionsOrderBy, positionsOrderDirection]);
+
+  const nextAccountsHash = utils.sha256(topAccountsUnordered.data ? topAccountsUnordered.data.map(a => (
```
gdev8317, line 139: at first glance sha256 is an overhead to calculate hash for this specific case, i guess we don't need hash to be extra secure

if it takes few milliseconds to calculate sha256 then it's okay, but if it takes e.g. 50-100+ ms (and considering positions amount will grow) i think would be good to use some faster/simpler hash algorithm.

src/components/GlobalLeaderboards/Context.tsx
```diff
@@ -0,0 +1,213 @@
...
+  PerfPeriod,
+  LeaderboardContextType,
+  useTopAccounts,
+  useTopPositions,
+  TopAccountsRow,
+  TopPositionsRow,
+  Profiler,
+} from "domain/synthetics/leaderboards";
+
+export const LeaderboardContext = createContext<LeaderboardContextType>({
```
gdev8317, line 24: it also contains a lot of logic/data that at first glance should be part of views. for example it has handlers like `topPositionsHeaderClick`. so context kind of aware of specific view implementation, feels like it should be the opposite, context should not depend on view

also thinking of all sorting logic should be part of context. is it reused in multiple places? could it better to have sorting logic inside views?

src/components/GlobalLeaderboards/TopAccounts.tsx
```diff
@@ -0,0 +1,154 @@
...
+import { TableCell, TableHeader } from "components/Table/types";
+import { TopAccountsRow, formatDelta } from "domain/synthetics/leaderboards";
+import AddressView from "components/AddressView";
+import Tooltip from "components/Tooltip/Tooltip";
+
+const parseRow = (s: TopAccountsRow): Record<string, TableCell> => ({
+  id: s.id,
+  rank: s.rank + 1,
+  account: {
+    value: "",
```
gdev8317, line 21: if i understand correctly `value` won't be used because `render` is passed? maybe better to update types so value is not required? i think passed `value` with empty string might be a bit confusing 

another option: value could be either string or function (i.e. no need to have 2 separate keys)

what do you think?

src/components/GlobalLeaderboards/TopPositions.tsx
```diff
@@ -0,0 +1,141 @@
...
+import { USD_DECIMALS, importImage } from "lib/legacy";
+import AddressView from "components/AddressView";
+import { TopPositionsRow, formatDelta } from "domain/synthetics/leaderboards";
+import Tooltip from "components/Tooltip/Tooltip";
+import { getPriceDecimals } from "config/tokens";
+import { BigNumber } from "ethers";
+import { MarketInfo } from "domain/synthetics/markets";
+
+const formatPriceDecimals = (x: BigNumber, market: MarketInfo, chainId: number) => formatAmount(
+  x,
```
divhead, line 20: we trying to avoid using of "x", "a" etc. parameters and variables since we have a lot of different numbers and math and it could be unclear what x means here. 

also, to avoid confusion in V2 we use "usd" and "amount" endings for usd and token amount values. eg. feeUsd / feeAmount.

src/components/GlobalLeaderboards/Context.tsx
ullin-oi: Yeah, the profiler will be removed

src/components/GlobalLeaderboards/Context.tsx
ullin-oi: > honestly i would avoid disabling eslint in hook deps if possible, because there could be unobvious bugs after updating the code

@divhead I agree, but unfortunately in this case (and other similar cases) it doesn't seem possible.

`react-hooks/exhaustive-deps` expects me to provide `topPositionsUnordered.data` array as a dependency, but it makes no sense to use it here as react is not going to compare the arrays via `_.isEqual` method or similar, and will compare them by reference instead.

Because of that I cannot use the array as a dependency at all, because it always is a different object, which makes the hook run all the time and breaks the logic.

I need to re-run this hook every time any of the values we use to sort positions by changes, and to identify the change the most straightforward way is to use the values themselves.

However `react-hooks/exhaustive-deps` is not capable of figuring out the latter value is derived from the one it expects as a dependency.

Please correct me if I am wrong.

src/components/GlobalLeaderboards/Context.tsx
ullin-oi: AFAIU it will only re-render when React UI tree nodes change. In this case it only happens when the relevant data changes.

Again, please correct me if I am wrong.

src/components/GlobalLeaderboards/Context.tsx
divhead: but, the values changes when new data comes from a request? every 5 secs as other data. So, just for discussion, i thing as the whole site re-renders every 5 secs it's okay to run this hook and this is better than miss some variable in deps after changes because the rule is disabled

src/components/GlobalLeaderboards/Context.tsx
divhead: just tested in Code sandbox. 

it re-renders when "value" object is changed, in this case at every render of the context

React Context doesn't provide some smart way to pick only neccesary data from state as Redux does

src/components/GlobalLeaderboards/Context.tsx
ullin-oi: I am not sure the fields of the object in the list change that often. Definitely different values change with different frequency. Say if we sort by "Win/Loss" column the values in it change only when someone closes a position and we get to know that, whereas changes in other columns are not related to the ordering. This makes me think the objects in the array may change much more often than we want to re-run this effect and reorder the rows. IMO triggering this callback only on relevant data change improves the overall performance noticeably.

src/components/GlobalLeaderboards/Context.tsx
ullin-oi: @divhead, what do you mean by re-render? If you mean expensive DOM manipulations, they only occur when the html tree changes, which is not our case as we always render the same table with the same elements in all of its' rows. I believe changes in the "value" object, even if there are many of those, are cheap as long as they do not trigger significant DOM changes. In our case they only change html attributes that correspond to the currently displayed data. AFAIK this is a React feature that is available irregardless of whether we use Context or not.

src/components/GlobalLeaderboards/Context.tsx
gdev8317: >  I would not agree the context depends on views. IMO it does not. You may be confused by the name of the callback.

probably yes. i think function name `topAccountsHeaderClick` assumes the context knows that it is used in a view with some header that can be clicked and click on this header changes sorting. okay with me to just change the name

> Keeping the sate in the context here also has other advantages. For example, if you move all parsing and sorting hooks into the view, its' state will be destroyed when we switch between tabs, as the state lives only as long as the component is mounted. Whereas if we use the context the way it is used here, all the state is preserved which allows to switch between tables without any additional delay.

i see. i guess it could make sense. some first thoughts on this:
1. we don't do this on other pages. is it okay to have different logic on leaderboard page or maybe better to keep it consistent? (i don't have strong opinion on that)
2. this context exists on leaderboard page only, so it will be destroyed when you open different page, right?

src/components/SearchInput/SearchInput.tsx
```diff
@@ -8,11 +9,21 @@ type Props = {
...
   className?: string;
   placeholder?: string;
+  autoFocus?: boolean;
 };
 
-export default function SearchInput({ value, setValue, onKeyDown, className, placeholder }: Props) {
+export default function SearchInput({
+  value,
+  setValue,
+  onKeyDown,
```
gdev8317, line 21: if i'm checking correctly onKeyDown is not used currently, empty handler is passed to the SearchInput. is there a reason to have this handler?

src/components/SearchInput/SearchInput.tsx
```diff
@@ -8,11 +9,21 @@ type Props = {
   value: string;
   setValue: (e: any) => void;
   onKeyDown: (e: any) => void;
+  onInput?: (e: FormEvent<HTMLInputElement>) => void;
```
gdev8317, line 12: i guess i've previously suggested to use onInput instead of noChange. i've realized that in react onChange works the same as native "input" event. so onChange should be sufficient, sorry for confusion

src/components/GlobalLeaderboards/Context.tsx
ullin-oi: Yes, it will

src/components/SearchInput/SearchInput.tsx
ullin-oi: It turns out `onInput` is not exactly the same as `onChange`, as the latter one is not triggered when we paste a text into the input field. To solve this either `onPaste` callback should be used along with `onChange` (in this case the handler will be the same), or alternatively one can set `onInput` callback instead, which is triggered both on text change and on paste.

src/components/GlobalLeaderboards/Context.tsx
divhead: discussed in DM

ullin-oi:
List of changes to the existing code base:

1. [SearchInput component](https://github.com/gmx-io/gmx-interface/pull/545/files#diff-416b25495ca23338ed67650807f46a3deb964c10b4ef960b732ab902f05722c8R9)
  - `value` - argument is now optional, should not affect other pages.
  - Added an optional `onInput` callback argument, which is convenient to handle input changes including copy-pastes, should not affect other pages.
  - Added an optional `autoFocus` argument as previously `SearchInput` had been setting auto focus to true for large screens unconditionally. This part changes optionally, i.e. it is now possible to control the `autoFocus` property by passing an optional argument, so it should not affect other pages.

2. [Tooltip component](https://github.com/gmx-io/gmx-interface/pull/545/files#diff-861240fa85e92a6425d07ac1df273c214b9cf97d611c917be6bd835ffadaa617R18)
  - Added a way to prevent popup clicks propagation. This change is hidden under an optional argument `isPopupClickable`, but it optionally adds and additional click handler to the popup content. So it should not affect other pages, but maybe requires a bit more attention.

3. [useFeesSummary hook](https://github.com/gmx-io/gmx-interface/pull/545/files#diff-4a6168673da4584c6816fee2b3cb87cc1379bf83117b9710f9c0cb7a466c0ebdR5)
  - A new test chain (`Goerli`) is added to `ACTIVE_CHAIN_IDS` constant. Because it is a test chain I am not sure it affects the hook significantly, but as it is used in the `DashboardV2` component in multiple places, I would recommend to pay attention to it.

4. [shortenAddress helper](https://github.com/gmx-io/gmx-interface/pull/545/files#diff-70095d214311e96d216d138189aad0062b7d48cfe93f7db432b63691917385baR724)
  - I've added an option to change the way the address is shortened, as previously the left part of the shortened address was 1 character longer than the right one, and there was no way to change it. The change only takes place when we pass in the third optional argument, therefore it should not affect other pages.
  - I've added a condition that if the length we are trying to shorten the address to is greater than or equals to the address length itself, will return the unchanged address. Without this condition the address gets shortened incorrectly in the specified case. Should not be a noticeable change, but requires some attention.

5. [formatUsd helper](https://github.com/gmx-io/gmx-interface/pull/545/files#diff-7fdf1a5171d1d1dbaacd892cb9454006354046c1f481ef39d068a04eed7afb25R172)
  - I've added a space between exceeding symbol (`<` or `>`) and the sign of the formatted number, as the space is also added in other helpers we use to display limited numbers. This helper is used in a lot of places across the entire app, so it definitely should be tested everywhere we display USD values and exceeding numbers may appear. Also please let me know if the space was removed here intentionally and I should not have added it.

6. [formatDeltaUsd helper](https://github.com/gmx-io/gmx-interface/pull/545/files#diff-7fdf1a5171d1d1dbaacd892cb9454006354046c1f481ef39d068a04eed7afb25R183)
  - This helper how accepts all the same options `getLimitedDisplay` accepts, so now it also supports formatted number thresholds. Because of that I would recommend to test all the places where we display deltas amounts.
  - Beside the above options, I've also added a new one that allows to hide delta percentage (following the displayed delta amount) for exceeding numbers. I would recommend to also pay attention to this change. It should only affect the places where we show deltas along with delta percentage.

7. [formatPercentage helper](https://github.com/gmx-io/gmx-interface/pull/545/files#diff-7fdf1a5171d1d1dbaacd892cb9454006354046c1f481ef39d068a04eed7afb25R214)
  - Added commas for thousands unconditionally. Should affect all places where we show percentage values. If the change is undesired, it is much easier to revert it rather than re-test all the pages, I think.

8. [Shared stylesheet](https://github.com/gmx-io/gmx-interface/pull/545/files#diff-dfdb2aa9b019b7391f6cc7a3f01b17b6ddfc872a2f1c7aa5935720103392c57bR33)
  - Added a fallback font family `sans-serif` for all input elements, because the target font `RelativeNumber` is also sans-serif and the input looks really ugly before the font gets loaded. Not sure if it's worth testing, but if it is all input elements should be reviewed.

PR #549, opened 2023-07-20
add liquidation hours

vipineth:
[x] Add liquidation hours

---
  - https://app.asana.com/0/0/1204734254756599


src/domain/synthetics/positions/utils.ts
```diff
@@ -245,3 +246,47 @@ export function formatLeverage(leverage?: BigNumber) {
 
   return `${formatAmount(leverage, 4, 2)}x`;
 }
+
+export function getEstimatedLiquidationTime(
+  position: PositionInfo,
+  minCollateralUsd: BigNumber | undefined,
+  period = "1h"
```
divhead, line 253: If this function only counts hours, I don't think we need to add the "period" parameter

src/domain/synthetics/positions/utils.ts
```diff
@@ -245,3 +246,47 @@ export function formatLeverage(leverage?: BigNumber) {
...
+
+  if (isOpening || !minCollateralUsd) return 0;
+
+  let liquidationCollateralUsd = applyFactor(sizeInUsd, marketInfo.minCollateralFactor);
+  if (liquidationCollateralUsd.lt(minCollateralUsd)) {
+    liquidationCollateralUsd = minCollateralUsd;
+  }
+  const borrowFeePerHour = getBorrowingFeeRateUsd(marketInfo, isLong, sizeInUsd, CHART_PERIODS[period]);
+  const fundingFeePerHour = getFundingFeeRateUsd(marketInfo, isLong, sizeInUsd, CHART_PERIODS[period]);
+  const priceImpactDeltaUsd = getPriceImpactForPosition(marketInfo, sizeInUsd.mul(-1), isLong, {
```
divhead, line 265: we also need limit price impact with maxNegativePriceImpactForLiquidations

The same as in getLiquidationPrice, maybe we can create function line "getPriceImpactForLiquidation" to avoid duplication
```
 const maxNegativePriceImpactUsd = applyFactor(sizeInUsd, marketInfo.maxPositionImpactFactorForLiquidations).mul(-1);

  if (priceImpactDeltaUsd.lt(maxNegativePriceImpactUsd)) {
      priceImpactDeltaUsd = maxNegativePriceImpactUsd;
  }

 // Ignore positive price impact
 if (priceImpactDeltaUsd.gt(0)) {
      priceImpactDeltaUsd = BigNumber.from(0);
 }
```

src/domain/synthetics/positions/utils.ts
```diff
@@ -245,3 +246,47 @@ export function formatLeverage(leverage?: BigNumber) {
...
+
+  const hours = netValue
+    .add(priceImpactDeltaUsd.lt(0) ? priceImpactDeltaUsd : 0)
+    .sub(liquidationCollateralUsd)
+    .div(borrowFeePerHour.abs().add(fundingFeePerHour.abs()));
+
+  return hours.toNumber();
+}
+
+export function formatEstimatedLiquidationTime(hours: number) {
```
divhead, line 281: hours seems like required parameter, but in common, i think formatter function should handle optinal params to simplify jsx code, so you can just change param to  hours?: number )

src/domain/synthetics/positions/utils.ts
divhead: maybe rename function to getEstimatedLiquidationTimeInHours?

src/domain/synthetics/positions/utils.ts
vipineth: Fixed!

src/domain/synthetics/positions/utils.ts
vipineth: Fixed!

src/domain/synthetics/positions/utils.ts
vipineth: Done!

PR #555, opened 2023-07-23
Synths fixes UI

vipineth:
[x] Change the sequence of warning and available liquidity

---
  - https://app.asana.com/0/0/1205101390585908
  - https://app.asana.com/0/0/1205101390583823
  - https://app.asana.com/0/0/1205086281608774
  - https://app.asana.com/0/0/1205111467193545
  - https://app.asana.com/0/0/1205111169593201
  - https://app.asana.com/0/0/1205111468311930
  - https://app.asana.com/0/0/1205111468311926


src/components/Exchange/ConfirmationBox.js
```diff
@@ -138,6 +140,9 @@ export default function ConfirmationBox(props) {
 
   const [allowedSlippage, setAllowedSlippage] = useState(savedSlippageAmount);
 
+  const existingPositionPriceDecimal = getPriceDecimals(chainId, existingPosition?.indexToken.symbol);
```
divhead, line 143: existing position is optinal, but getPriceDecimals will throw an error if there is no token

divhead: also i think existingPosition.indexToken === marketInfo.indexToken here

so we don't need to use existingPosition

src/components/Exchange/ConfirmationBox.js
```diff
@@ -138,6 +140,9 @@ export default function ConfirmationBox(props) {
...
+  const toTokenPriceDecimal = getPriceDecimals(chainId, toToken.symbol);
```
divhead, line 144: and toToken should be === indexToken in case of position hm

src/components/Exchange/PositionEditor.js
```diff
@@ -76,6 +77,7 @@ export default function PositionEditor(props) {
   const [isSwapping, setIsSwapping] = useState(false);
   const prevIsVisible = usePrevious(isVisible);
   const longOrShortText = position?.isLong ? t`Long` : t`Short`;
+  const positionPriceDecimal = getPriceDecimals(chainId, position?.indexToken?.symbol);
```
divhead, line 80: same here, getPriceDecimals will throw an error

PR #570, opened 2023-08-08
Fix incorrect referral discount

vipineth:
[x] Fix incorrect referral discount

---
  - https://app.asana.com/0/0/1205227361086463
  - https://app.asana.com/0/0/1204834612413163
  - https://app.asana.com/0/0/1205276763751836


src/components/Referrals/referralsHelper.js
```diff
@@ -69,6 +76,26 @@ export const tierDiscountInfo = {
   2: 10,
 };
 
+export const totalRebateInfo = {
+  0: 10,
+  1: 20,
+  2: 25,
+};
```
divhead, line 83: hm, we don't need to hardcode this, beacause we have useTiers hook which requests ReferralStorage for actual totalTebate config

src/components/Referrals/referralsHelper.js
```diff
@@ -10,10 +10,17 @@ import { ARBITRUM, AVALANCHE } from "config/chains";
 import { bigNumberify, formatAmount } from "lib/numbers";
 import { t } from "@lingui/macro";
 import { getRootUrl } from "lib/url";
+import { BASIS_POINTS_DIVISOR } from "config/factors";
 
 export const REFERRAL_CODE_REGEX = /^\w+$/; // only number, string and underscore is allowed
 export const REGEX_VERIFY_BYTES32 = /^0x[0-9a-f]{64}$/;
 
+export function removeTrailingZeros(amount) {
```
divhead, line 18: probably it's better to move this function to lib/numbers

src/components/Referrals/referralsHelper.js
vipineth: updated!

src/components/Referrals/referralsHelper.js
vipineth: Updated

divhead (dismissed):

divhead (approved):

PR #590, opened 2023-08-17
replace web3react with wagmi + rainbowkit

vipineth:
[x] Add new wallets
[x] Shorten ens

---
  - https://app.asana.com/0/0/1202229483086810


src/App/App.js
```diff
@@ -810,26 +675,38 @@ function FullApp() {
...
   useScrollToTop();
   useEffect(() => {
     const defaultLanguage = localStorage.getItem(LANGUAGE_LOCALSTORAGE_KEY) || defaultLocale;
     dynamicActivate(defaultLanguage);
   }, []);
+
+  useEffect(() => {
+    const unwatch = watchNetwork(({ chain, chains }) => {
+      if (!chain || !chains) return;
+      const isValidChain = !!chains.find((c) => c.id === chain.id);
```
divhead, line 689: we can use .some to avoid Implicit cast to Boolean via !!

divhead (approved):

PR #595, opened 2023-08-19
Add v2 stats to overview

vipineth:
[x] Add v2 stats like volume, fees, open interests to dashboard

---
  - https://app.asana.com/0/0/1204819305030365


src/components/StatsTooltip/ChainsStatsTooltipRow.tsx
```diff
@@ -5,48 +5,45 @@ import "./StatsTooltip.css";
...
-  total,
-  avaxValue,
-  arbitrumValue,
+  entries,
   showDollar = true,
   decimalsForConversion = USD_DECIMALS,
   symbol,
   shouldFormat = true,
 }: Props) {
+  const validEntries = Object.entries(entries).filter(
```
divhead, line 22: why there entries is used if only values needed?
maybe change to Object.values?

src/lib/numbers.ts
```diff
@@ -331,6 +331,18 @@ export function roundToTwoDecimals(n) {
   return Math.round(n * 100) / 100;
 }
 
+export function addTwoBigNumbers(a, b) {
```
divhead, line 334: what if we will need to add more bigNumbers?
```
sumBigNumbers(...args) {
 return args.reduce(acc, value => acc.add(value || 0), BigNumber.from(0))
}
```

src/domain/synthetics/markets/useMarketsOverview.ts
```diff
@@ -0,0 +1,73 @@
...
+  totalShortPositionSizes: BigNumber;
+  openInterest: BigNumber;
+  dailyVolume: BigNumber;
+  totalVolume: BigNumber;
+  weeklyFees: BigNumber;
+  totalFees: BigNumber;
+  totalUsers: BigNumber;
+};
+
+export default function useMarketsOverview(chains: number[]) {
```
divhead, line 20: is it used somewhere?

it's a bit strange to depend on two chainIds

src/domain/synthetics/stats/useUsers.ts
```diff
@@ -0,0 +1,65 @@
...
+    }
+  }
+`;
+
+type UsersInfo = {
+  totalUsers: BigNumber;
+};
+
+export default function useUsers(chains: number[]) {
+  async function fetchUsersInfo(chain: number) {
```
divhead, line 19: "chainId" would be better as in other places

src/domain/synthetics/stats/useUsers.ts
```diff
@@ -0,0 +1,65 @@
...
+    } catch (error) {
+      // eslint-disable-next-line no-console
+      console.error("Error fetching feesInfo data:", error);
+      return {
+        totalUsers: BigNumber.from(0),
+      };
+    }
+  }
+
+  const { data: feesInfo } = useSWR("usersInfo", fetcher, {
```
divhead, line 60: swr key should depend on variable params, chains here is a parameter of the hook. if we will use this hook in different places with different chains - we will receive the same cached data from first hook

we need either remove chains from hook param and fix them to SUPPORTED_CHAIN_IDS or add it to swr key and fetcher params

src/domain/synthetics/stats/useUsers.ts
```diff
@@ -0,0 +1,65 @@
...
+      return {
+        totalUsers: BigNumber.from(0),
+      };
+    }
+  }
+
+  async function fetcher() {
+    try {
+      const promises = chains.map(fetchUsersInfo);
+      const results = await Promise.allSettled(promises);
```
divhead, line 42: why not use just Promise.all ? 
anyway if some chain request is broken then result value will be incorrect

src/domain/synthetics/stats/useVolumeInfo.ts
```diff
@@ -0,0 +1,78 @@
...
+        }, {});
+      return volumes;
+    } catch (error) {
+      // eslint-disable-next-line no-console
+      console.error("Error fetching volume data:", error);
+      return {};
+    }
+  }
+
+  const { data: volumes } = useSWR("v2VolumeInfos", fetcher, {
```
divhead, line 73: same here, need to fix swr key

src/domain/synthetics/stats/useVolumeInfo.ts
```diff
@@ -0,0 +1,78 @@
...
+    totalVolumeInfos: volumeInfos(where: { period: "total" }) {
+      volumeUsd
+      period
+      id
+    }
+  }
+`;
+
+export default function useVolumeInfo(chains: number[]) {
+  const lastPeriodFor24Hours = Math.floor(Date.now() / 1000 / 3600) * 3600 - 60 * 60 * 24;
```
divhead, line 26: the time here depends on hook first render time, then if swr key is not changed, it will not update the fetcher function and will refer to the initial time value

we need to move this into fetcher function

src/components/StatsTooltip/ChainsStatsTooltipRow.tsx
vipineth: Using entries I am creating an array of all the valid entries based on the value being truthy. So, if the value is falsy we are not adding the entry to the valid entries. We map over valid entries and display so, we will need the entries.

src/lib/numbers.ts
vipineth: Great idea!

src/components/StatsTooltip/ChainsStatsTooltipRow.tsx
divhead: i understand, but i don't see any usage of key

just wondering why we don't use Object.values here, not critial

src/domain/synthetics/stats/useFeesInfo.ts
```diff
@@ -0,0 +1,104 @@
...
+  const { data: feesSummaryByChain } = useFeesSummary();
+  const lastUpdatedAt = feesSummaryByChain[chainId]?.lastUpdatedAt;
+
+  async function fetchFeesInfo(chainId: number) {
+    try {
+      const client = getSyntheticsGraphClient(chainId);
+      const { data: weeklyFeesInfo } = await client!.query({
+        query: weeklyFeeQuery,
+        variables: {
+          lastTimestamp: lastUpdatedAt,
```
divhead, line 44: it still won't work properly because fetchFeesInfo don't accept lastUpdatedAt in params and get it from external scope at the render

src/domain/synthetics/stats/useFeesInfo.ts
```diff
@@ -0,0 +1,104 @@
...
+      // eslint-disable-next-line no-console
+      console.error(`Error fetching feesInfo data for chain ${chainId}:`, error);
+      return {
+        weeklyFees: BigNumber.from(0),
+        totalFees: BigNumber.from(0),
+      };
+    }
+  }
+
+  async function fetcher(timestamp: number) {
```
divhead, line 84: maybe rename timestamp to something more semantic? like lastFeesUpdatedTimestamp. btw, do we need to depend on feesSummaryByChain? because we don't use values from this entity maybe just create timestamp by Date.now() inside fetcher would be enough? or there is some issues?

src/domain/synthetics/stats/useFeesInfo.ts
```diff
@@ -0,0 +1,104 @@
...
+        totalFees,
+      };
+    } catch (error) {
+      // eslint-disable-next-line no-console
+      console.error("Error fetching feesInfo data:", error);
+      return {};
+    }
+  }
+
+  const { data: feesInfo } = useSWR(lastUpdatedAt, fetcher, {
```
divhead, line 99: now the key is danger in terms of cache collisions, i think it would be better to add some fixed string like "useFeesInfo" to the key

src/domain/synthetics/stats/useDashboardOverview.ts
```diff
@@ -0,0 +1,57 @@
+import { useMemo } from "react";
+import { BigNumber } from "ethers";
+import useVolumeInfo from "../stats/useVolumeInfo";
+import useFeesInfo from "../stats/useFeesInfo";
+import useUsers from "../stats/useUsers";
+import { useMarketsInfo } from "../markets";
+
+type MarketOverview = {
```
divhead, line 8: hm, maybe the name should be DashboardOverview?

src/domain/synthetics/stats/useFeesInfo.ts
vipineth: On dashboard we display fees after certain data (which gets updated every week). So for v2 I want to use the same date to show the fees after. So I am using the lastUpdatedAt from `feesSummaryByChain`. So I think we need to user it.

src/domain/synthetics/stats/useFeesInfo.ts
```diff
@@ -0,0 +1,104 @@
...
+        totalFees,
+      };
+    } catch (error) {
+      // eslint-disable-next-line no-console
+      console.error("Error fetching feesInfo data:", error);
+      return {};
+    }
+  }
+
+  const { data: feesInfo } = useSWR(["useFeesInfo", lastUpdatedAt], fetcher, {
```
divhead, line 99: maybe we don't need lastIpdatedAt since we have refreshInterval here?

divhead (dismissed):

divhead (dismissed):

divhead (approved):

PR #598, opened 2023-08-23
add trade params to v1 and v2

vipineth:
[x] Add route params

---
  - https://app.asana.com/0/0/1204202517415994


src/App/App.js
```diff
@@ -459,7 +459,7 @@ function FullApp() {
               <Route exact path="/buy">
                 <Buy savedSlippageAmount={settings.savedAllowedSlippage} setPendingTxns={setPendingTxns} />
               </Route>
-              <Route exact path="/pools">
+              <Route exact path="/pools/:market?/:operation?/:mode?">
```
divhead, line 462: we don't change page content based on the params righs? it just select some things in TradeBox. For that case usually query params is used instead of path params because path params is structured and if we will add more params then old links will be invalid

divhead: also with query params is more flexible, imagine that we want to post some link in twitter post to suggest users open Shorts and don't want to specify a market

src/components/Exchange/SwapBox.js
```diff
@@ -212,9 +209,8 @@ export default function SwapBox(props) {
 
   const [ordersToaOpen, setOrdersToaOpen] = useState(false);
 
-  let [orderOption, setOrderOption] = useLocalStorageSerializeKey([chainId, "Order-option"], MARKET);
   if (!flagOrdersEnabled) {
-    orderOption = MARKET;
+    setOrderOption(MARKET);
```
divhead, line 213: it's better to move it inside useEffect and check that option is not MARKET already

src/components/Glp/GlpSwap.js
```diff
@@ -422,10 +432,50 @@ export default function GlpSwap(props) {
     totalTokenWeights,
   ]);
 
+  useEffect(() => {
+    const { operation } = params;
```
divhead, line 436: same here, better to keep all params in "query"

src/components/Glp/GlpSwap.js
```diff
@@ -422,10 +432,50 @@ export default function GlpSwap(props) {
...
+    const pay = queryParams.get("pay");
+    const receive = queryParams.get("receive");
+
+    if (operation) {
+      setIsBuying(operation.toLowerCase() === "buy");
```
divhead, line 441: let's also create enums for param keys and possible values to better readability

src/config/tokens.ts
```diff
@@ -992,3 +992,16 @@ export function getPriceDecimals(chainId: number, tokenSymbol?: string) {
     return 2;
   }
 }
+
+export function getValidTokenBySymbol(chainId: number, symbol: string, version: "v1" | "v2") {
+  const ACTIVE_TOKENS = version === "v1" ? V1_TOKENS : V2_TOKENS;
```
divhead, line 997: it's better to use getV1Tokens / getV2Tokens

divhead: maybe jsut modify getTokenBySymbol? It's not clean what "Valid" token means

src/domain/synthetics/trade/useSelectedTradeOption.ts
```diff
@@ -33,6 +42,14 @@ export type SelectedTradeOption = {
   setMarketAddress: (marketAddress?: string) => void;
   setCollateralAddress: (tokenAddress?: string) => void;
   switchTokenAddresses: () => void;
+  setTradeOptions: ({
```
divhead, line 45: trade option should be in the singular here i think, roughly speaking it's trade config

src/pages/Exchange/Exchange.js
```diff
@@ -487,6 +523,107 @@ export const Exchange = forwardRef((props, ref) => {
...
+      setSwapOption,
+      setOrderOption,
+      setFromAndToTokenAddress,
+      setFromTokenAddress,
+      setToTokenAddress,
+      setShortCollateralAddress,
+    ]
+  );
+
+  const tradeOptions = useMemo(() => {
```
divhead, line 560: maybe move it to separate hook and and other hooks too?

src/locales/fr/messages.po
```diff
@@ -4806,7 +4806,7 @@ msgstr "TOKEN"
 
 #: src/components/Synthetics/GmList/GmList.tsx
 msgid "TOTAL SUPPLY"
-msgstr ""
+msgstr "<<<<<<< HEAD======="
```
onemicky, line 4809: Looks like a merge conflict

onemicky (changes requested):

onemicky (dismissed):

onemicky (approved):

onemicky (approved):

onemicky (approved):

PR #604, opened 2023-08-29
add bridging info

vipineth:
[x] Add bridging info for tokens

---
  - https://app.asana.com/0/0/1205276763751738


src/domain/synthetics/markets/getBridigingInfo.tsx
```diff
@@ -0,0 +1,26 @@
+import { Trans } from "@lingui/macro";
+import ExternalLink from "components/ExternalLink/ExternalLink";
+import getBridgingOptionsForToken from "config/bridging";
+import { getChainName } from "config/chains";
+
+export default function getBridgingInfo(chainId: number, tokenSymbol?: string) {
```
divhead, line 6: not sure that this code relatred to markets. Also there is jsx and not some maths.  Probably better to create component BridgingInfo

vipineth: Thanks, for the feedback. Updated

src/config/bridging.ts
```diff
@@ -0,0 +1,33 @@
...
+    },
+  ],
+};
+
+export function getBridgingOptionsForToken(tokenSymbol?: string): BridgingOptionsReturn {
+  if (!tokenSymbol) return;
+  return BRIDGING_OPTIONS[tokenSymbol];
+}
+
+export function isBridgingInfoAvailableForToken(tokenSymbol?: string): boolean {
```
divhead, line 31: mm, probably we don't need this func as we could just use getBridgingOptionsForToken ?

src/config/bridging.ts
```diff
@@ -0,0 +1,33 @@
+import { ARBITRUM, AVALANCHE } from "./chains";
+
+const SOLANA_BRIDGING_IDS = {
+  [ARBITRUM]: 23,
+  [AVALANCHE]: 6,
+};
+
+type BridgingOptionsReturn =
```
divhead, line 8: probaby just BridgingOptions?

src/config/bridging.ts
```diff
@@ -0,0 +1,33 @@
...
+  | [
+      {
+        name: string;
+        generateLink: (chainId: number) => string;
+      }
+    ]
+  | undefined;
+
+const BRIDGING_OPTIONS = {
```
divhead, line 17: would be good to specify type here to increase intellisense

BRIDGING_OPTIONS: {[symbol: string]: BridgingOptions[]}

divhead (approved):

PR #605, opened 2023-08-31
Add v2 charttokenselector

vipineth:
[x] Add token selector dropdown for v2

---
  - https://app.asana.com/0/0/1205283385387744


src/components/Synthetics/ChartTokenSelector/ChartTokenSelector.tsx
```diff
@@ -0,0 +1,246 @@
...
+    return {
+      maxLongLiquidityPool,
+      maxShortLiquidityPool,
+    };
+  }
+
+  return (
+    <Popover className="Synths-ChartTokenSelector">
+      {({ open, close }) => {
+        if (!open) setSearchKeyword("");
```
divhead, line 139: looks dangerous to set state only by "open" field, probably we could also chech that search.length > 0

vipineth: Updated!

divhead (approved):

PR #614, opened 2023-09-11
add sharing popup to v2 positions

gmxer:
added `<PositionShare .. />` to PositionList

Discussed with @divhead that there's no need to split sharing popup. So I just unified its props so that both versions work fine with it now. Hope that makes sense.

task https://app.asana.com/0/1204313444805313/1205268602258545/f


vipineth (changes requested):

vipineth:
Sorry for not noticing it earlier. I was able to click on the Tweet even when the image was not uploaded. Resulting into a message like. I think it's happening because tweet is an Anchor tag and disabled doesn't work on Anchor. It's an existing issue but would be awesome if that can be fixed.


Screenshot 2023-09-20 at 12 48 19 PM

Also after making the changes please change the base from `master` branch to `release-1`.

src/components/Exchange/PositionShare.tsx
```diff
@@ -34,22 +31,44 @@ function getShareURL(imageInfo, ref) {
...
+  indexToken,
+  isLong,
+  leverage,
+  markPrice,
+  pnlAfterFeesPercentage,
+  setIsPositionShareModalOpen,
+  isPositionShareModalOpen,
+  account,
+  chainId,
+}: {
```
divhead, line 45: let's move prop types to type Props?

src/components/Synthetics/PositionList/PositionList.tsx
```diff
@@ -109,10 +123,29 @@ export function PositionList(p: Props) {
                 currentTradeType={p.currentTradeType}
                 openSettings={p.openSettings}
                 hideActions={p.hideActions}
+                onShareClick={() => {
+                  setPositionToShareKey(position.key);
```
divhead, line 127: maybe create a function to reuse the code?

src/components/Exchange/PositionShare.tsx
gmxer: I'm ok with it in general but I don't really see a reason to create `p` variable for props. It just feels redundant

PR #617, opened 2023-09-13
update wallet list

vipineth:
[x] Update wallet list
[x] Add core custom connector
[x] Add bitKeep custom connector

---
  - https://app.asana.com/0/0/1205414635528502


src/lib/wallets/connecters/bitGet/bitGetWallet.ts
```diff
@@ -0,0 +1,148 @@
...
+    window.bitkeep.ethereum !== undefined &&
+    // @ts-expect-error
+    window.bitkeep.ethereum.isBitKeep === true;
+
+  const shouldUseWalletConnect = !isBitKeepInjected;
+
+  return {
+    id: "bitKeep",
+    name: "Bitget Wallet",
+    iconUrl: async () => (await import("./bitGetWallet.svg")).default,
```
divhead, line 54: why we don't use importImage here?

divhead (approved):

src/lib/wallets/connecters/bitGet/bitGetWallet.ts
vipineth: I am just trying to follow the Rainbowkit adapter modal so it will be easier to update and fix.

PR #623, opened 2023-09-19
Trade Log Improvements for V2

gmxer:
I introduced  `formatPositionMessage` for `PositionTradeAction`. It returns:
```
{
    text: string;
    tooltipProps?: StatsTooltipRowProps[];
    tooltipTitle?: string;
    ...
 }[]
```
So that we can just render text or also tooltip with StatsTooltipRows on top of text.
As you can see logic is covered with tests, hope that makes sense

LiquidationTooltip is elliminated, as tooltipProps + tooltipTitle are now responsible for all tooltips.

I didn't touch anything related to `SwapTradeAction`, thus only `PositionTradeAction` is affected.

also introduced `museNeverExist`, hope we gonna use it more often on enum-like types

---
  - https://app.asana.com/0/0/1205172761500164


src/components/Synthetics/TradeHistoryRow/TradeHistoryRow.tsx
```diff
@@ -1,62 +1,29 @@
...
-import "./TradeHistoryRow.scss";
-import { getTriggerThresholdType } from "domain/synthetics/trade";
-import { useSettings } from "context/SettingsContext/SettingsContextProvider";
+import { formatTokenAmount } from "lib/numbers";
+import { useMemo } from "react";
 import { Link } from "react-router-dom";
-import { formatAcceptablePrice } from "domain/synthetics/positions";
-import { MarketInfo, getMarketIndexName, getMarketPoolName } from "domain/synthetics/markets";
+import "./TradeHistoryRow.scss";
+import { formatPositionMessage, getOrderActionText } from "./helpers";
```
divhead, line 17: usually we use "utils" naming instead of helpers

src/components/Synthetics/TradeHistoryRow/helpers.ts
```diff
@@ -0,0 +1,382 @@
...
+import { PositionTradeAction, TradeAction, TradeActionType } from "domain/synthetics/tradeHistory";
+import { BigNumber, ethers } from "ethers";
+import { PRECISION } from "lib/legacy";
+import { formatDeltaUsd, formatTokenAmount, formatTokenAmountWithUsd, formatUsd } from "lib/numbers";
+import { museNeverExist } from "lib/types";
+import { trimStart } from "lodash";
+
+type FormatPositionMessageChunk = {
+  text: string;
+  textRed?: boolean;
```
divhead, line 18: i'm not sure about textRed naming

1. textRed sounds as a string field
2. is it field to indicate that PositionMessage has some error right? maybe just rename it to isError?

src/components/Synthetics/TradeHistoryRow/helpers.ts
```diff
@@ -0,0 +1,382 @@
...
+              t`Execution Price: ${formatUsd(executionPrice, { displayDecimals: priceDecimals })}`,
+            ]
+              .filter(Boolean)
+              .join(", "),
+            tooltipRows: getExecutionPriceTooltipRows(tradeAction),
+          },
+        ];
+      }
+
+      const isFailed = tradeAction.eventName === TradeActionType.OrderFrozen;
```
divhead, line 98: probably it's better to keep isFrozen naming to avoid confusion with cancelled orders

src/components/Synthetics/TradeHistoryRow/helpers.ts
```diff
@@ -0,0 +1,382 @@
+import CustomErrors from "abis/CustomErrors.json";
```
divhead, line 1: there is also better to use utils.ts naming

src/components/Synthetics/TradeHistoryRow/TradeHistoryRow.tsx
```diff
@@ -111,104 +80,53 @@ function getSwapOrderMessage(tradeAction: SwapTradeAction) {
...
+          <Tooltip
+            position="left-top"
+            handle={textElement}
+            className={message.textRed ? "Tooltip-error" : undefined}
+            renderContent={() => (
+              <>
+                {message.tooltipTitle ? (
+                  <span className={message.tooltipTitleRed ? "text-red" : undefined}>
+                    {message.tooltipTitle}
+                    {hasSmthAfterTitle && <br />}
```
divhead, line 106: it could be just 
```
{hasSmthAfterTitle && (<><br/><br/></>)}
```

or maybe we can do it in css

divhead (approved):

PR #624, opened 2023-09-21
add market token dropdown

vipineth:
[x] Add dropdown to pools page

---
  - https://app.asana.com/0/0/1205214480183554


src/components/Synthetics/MarketTokenSelector/MarketTokenSelector.tsx
```diff
@@ -0,0 +1,179 @@
...
+                        <tr>
+                          <th>MARKET</th>
+                          <th>MINTABLE</th>
+                          <th>SELLABLE</th>
+                          <th>APR</th>
+                        </tr>
+                      </thead>
+                    )}
+                    <tbody>
+                      {filteredTokens.map((market) => {
```
divhead, line 117: i think it's better to move the logic outside JSX

src/components/Synthetics/MarketTokenSelector/MarketTokenSelector.tsx
```diff
@@ -0,0 +1,179 @@
...
+                        if (!market) return null;
+                        const marketInfoData = getByKey(marketsInfoData, market?.address)!;
+                        const mintableInfo =
```
divhead, line 120: it could be simplified to one if

```
if(!marketInfo || market || ...) {
return null
}
```

vipineth: Updated

src/components/Synthetics/MarketTokenSelector/MarketTokenSelector.tsx
vipineth: Updated

src/components/MarketSelector/PoolSelector.tsx
```diff
@@ -54,36 +57,59 @@ export function PoolSelector({
...
           balance: gmBalance || BigNumber.from(0),
           balanceUsd: gmBalanceUsd || BigNumber.from(0),
           state,
         };
       });
-  }, [getMarketState, marketTokensData, markets, selectedIndexName]);
+    const marketsWithBalance: MarketOption[] = [];
+    const marketsWithoutBalance: MarketOption[] = [];
+
+    for (let market of allMarkets) {
```
gdev8317, line 90: iirc we use `const` when possible?

src/components/MarketSelector/PoolSelector.tsx
```diff
@@ -54,36 +57,59 @@ export function PoolSelector({
...
+      if (market.balance.gt(0)) {
+        marketsWithBalance.push(market);
+      } else {
+        marketsWithoutBalance.push(market);
+      }
+    }
+
+    const sortedMartketsWithBalance = marketsWithBalance.sort((a, b) => {
+      return b.balanceUsd?.sub(a.balanceUsd || 0).gt(0) ? 1 : -1;
```
gdev8317, line 99: just to confirm, `b.balanceUsd?.gt(a.balanceUsd || 0) ? 1 : -1` would give the same result? looks a bit cleaner/simpler

src/components/MarketSelector/PoolSelector.tsx
```diff
@@ -54,36 +57,59 @@ export function PoolSelector({
...
+  }, [getMarketState, marketTokensData, markets, selectedIndexName, showAllPools]);
 
   const marketInfo = marketsOptions.find(
     (option) => option.marketInfo.marketTokenAddress === selectedMarketAddress
   )?.marketInfo;
 
   const filteredOptions = marketsOptions.filter((option) => {
-    return option.poolName.toLowerCase().indexOf(searchKeyword.toLowerCase()) > -1;
+    const name = option.name.toLowerCase();
+    const search = searchKeyword.toLowerCase();
```
gdev8317, line 111: just i thing to consider for future: may be better to define such vars outside of the loop since it just does the same each time (`searchKeyword` is the same for each iteration)

src/components/MarketSelector/PoolSelector.tsx
```diff
@@ -54,36 +57,59 @@ export function PoolSelector({
...
+    return name.indexOf(search) > -1;
```
gdev8317, line 112: not critical, might be a bit cleaner to use `includes` for such things in future, works for both strings and arrays

gdev8317 (approved):

PR #645, opened 2023-10-10
Slippage validation

vipineth:
[x] Add slippage validation

---
  - https://app.asana.com/0/0/1204911431110707


src/App/App.js
```diff
@@ -606,6 +580,14 @@ function FullApp() {
             />
             <div className="App-slippage-tolerance-input-percent">%</div>
           </div>
+          {parseFloat(slippageAmount) < (DEFAULT_ALLOWED_SLIPPAGE_BPS / BASIS_POINTS_DIVISOR) * 100 && (
+            <div className="warning settings-modal-error">
+              <Trans>
+                Allowed Slippage below {(DEFAULT_ALLOWED_SLIPPAGE_BPS / BASIS_POINTS_DIVISOR) * 100}% may result in
```
gmxer, line 586: I know `formatPercentage` works with BigNumbers but wouldn't it still better to use it? To keep consistency in formatting

gmxer: Also I'm not sure I fully get the idea of BASIS_POINTS_DIVISOR here.
Why not just make DEFAULT_ALLOWED_SLIPPAGE = 0.3?

gmxer (changes requested):
1 question/comment from me

src/App/App.js
vipineth: Using `formatPercentage` here would make the code unnecessarily complicated and adds overhead, especially given the limitations with decimals and `BigNumber`.

Regarding the use of `BASIS_POINTS_DIVISOR`, it ensures that our internal calculations remain consistent throughout the app. We store `DEFAULT_ALLOWED_SLIPPAGE` in BPS, which stands for "basis points." This is a commonly used term in finance to signify changes in value. Also helps to store integers in local storage.

gmxer: ok, makes sense

gmxer (approved):

PR #651, opened 2023-10-12
fix loop issue

vipineth:
[x] Fix infinite loop issue

### DIFF
--- a/src/App/App.js
+++ b/src/App/App.js
@@ -271,8 +271,14 @@ function FullApp() {
     setSelectedToPage(to);
   };
 
+  const isFirstRender = useRef(false);
+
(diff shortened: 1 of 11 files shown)


src/App/App.js
```diff
@@ -271,8 +271,14 @@ function FullApp() {
     setSelectedToPage(to);
   };
 
+  const isFirstRender = useRef(false);
```
gmxer, line 274: weird to see that isFirstRender=false for the first render and isFirstRender=true for the all other renders, it should have opposite name/value

vipineth: Sorry, I was still refactoring. It's updated now.

divhead (approved):

PR #660, opened 2023-10-23
add adaptive funding rate text to markets

vipineth:
[x] Add adaptive funding rate text to events and on the market card on trade page

---
  - https://app.asana.com/0/0/1205743945727419

### DIFF
--- a/src/components/Synthetics/MarketCard/MarketCard.tsx
+++ b/src/components/Synthetics/MarketCard/MarketCard.tsx
@@ -63,6 +63,7 @@ export function MarketCard({ marketInfo, allowedSlippage, isLong, isIncrease }: 
 
   const renderFundingFeeTooltipContent = useCallback(() => {
     if (!fundingRateLong || !fundingRateShort) return [];
+    const isMarketWithAdaptiveFundingRate = marketInfo?.fundingIncreaseFactorPerSecond.gt(0);
 
     const isLongPositive = fundingRateLong?.gt(0);
     const long = (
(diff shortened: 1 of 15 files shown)


src/components/Synthetics/MarketCard/MarketCard.tsx
```diff
@@ -96,9 +97,22 @@ export function MarketCard({ marketInfo, allowedSlippage, isLong, isIncrease }:
...
+          <>
+            <br />
+            <br />
+            <Trans>
+              This market uses an Adaptive Funding Rate. The Funding Rate will adjust over time depending on the ratio
+              of longs and shorts.
+              <br />
+              <br />
+              <ExternalLink href="https://docs.gmx.io/docs/trading/v2/#adaptive-funding">Read more</ExternalLink>.
+            </Trans>
```
gmxer, line 110: Will "Read more" be translated by `<Trans>` that way?
feels like it should have separate `<Trans>` in children

gmxer (changes requested):
left 1 comment

src/components/Synthetics/MarketCard/MarketCard.tsx
vipineth: It will be translated, Trans takes care of all the text element inside.

The message will look like this `This market uses an Adaptive Funding Rate. The Funding Rate will adjust over time depending on the ratio of longs and shorts.<0/><1/><2>Read more</2>.`

gmxer (approved):

PR #674, opened 2023-11-03
Add incentive info

vipineth:
[x] Add GLP to GM migration info

---
  - https://app.asana.com/0/0/1205829386512416

### DIFF
--- a/src/components/Glp/GlpSwap.js
+++ b/src/components/Glp/GlpSwap.js
@@ -73,10 +73,25 @@ import PageTitle from "components/PageTitle/PageTitle";
 import useIsMetamaskMobile from "lib/wallets/useIsMetamaskMobile";
 import { MAX_METAMASK_MOBILE_DECIMALS } from "config/ui";
 import { getFeeItem } from "domain/synthetics/fees";
+import { intervalToDuration, nextWednesday } from "date-fns";
 
 const { AddressZero } = ethers.constants;
 
 const MAX_REBATE_BPS = 25;
+
+function getTimeLeftToNextWednesday() {
+  const now = new Date();
+  const nextWed = nextWednesday(now);
+
+  const duration = intervalToDuration({
+    start: now,
+    end: Date.UTC(nextWed.getUTCFullYear(), nextWed.getUTCMonth(), nextWed.getUTCDate()),
+  });
+  const days = duration.days ? `${duration.days}d ` : "";
+  const hours = duration.hours ? `${duration.hours}h ` : "";
+  const minutes = duration.minutes ? `${duration.minutes}m` : "";
+  return `${days}${hours}${minutes}`.trim();
+}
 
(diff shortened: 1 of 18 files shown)


src/components/Glp/GlpSwap.js
```diff
@@ -973,10 +987,43 @@ export default function GlpSwap(props) {
...
+                          <>
+                            <StatsTooltipRow
+                              label="Base Fee"
+                              value={formatDeltaUsd(feeItem.deltaUsd, feeItem.bps)}
+                              showDollar={false}
+                              className="text-red"
+                            />
+                            <StatsTooltipRow
+                              label="Max Bonus Rebate"
+                              value={`+${formatUsd(maxRebateUsd.abs())}`}
```
gdev8317, line 999: using `abs` looks a bit inconsistent

here https://github.com/gmx-io/gmx-interface/pull/674/files#diff-cd0741d6fc8d0943f840d44461016a3ff764a177a6d6da1ab44f38859d4d65d3R976 `swapUsdMin.mul(-1)` is used

and here https://github.com/gmx-io/gmx-interface/pull/674/files#diff-cd0741d6fc8d0943f840d44461016a3ff764a177a6d6da1ab44f38859d4d65d3R979 just `swapUsdMin`. i guess it's better to use `swapUsdMin.mul(-1)` in both cases for simplicity?

vipineth: `formatUsd(maxRebateUsd.abs())` is used to display the rebate reward amount. So I used `abs` to make sure it will always be positive.

I have updated the use of `swapUsdMin` to be consistent now.

src/lib/numbers.ts
```diff
@@ -320,7 +320,14 @@ export function applyFactor(value: BigNumber, factor: BigNumber) {
 }
 
 export function getBasisPoints(numerator: BigNumber, denominator: BigNumber) {
-  return numerator.mul(BASIS_POINTS_DIVISOR).div(denominator);
+  const result = numerator.mul(BASIS_POINTS_DIVISOR).div(denominator);
+  const remainder = numerator.mul(BASIS_POINTS_DIVISOR).mod(denominator);
+
+  if (!remainder.isZero()) {
+    return result.isNegative() ? result.sub(1) : result.add(1);
```
gdev8317, line 327: is it always desirable behavior? we're using `getBasisPoints` in many places, i'm not sure if it is expected to update rounding logic for all of them

gdev8317 (dismissed):

src/lib/numbers.ts
vipineth: Updated!

gdev8317 (dismissed):

gdev8317 (dismissed):

gmxer (dismissed):

gdev8317 (approved):

PR #675, opened 2023-11-07
Add max open interest

vipineth:
[x] Add MAX_OPEN_INTEREST

---
  - https://app.asana.com/0/0/1205877859617801

### DIFF
--- a/src/domain/synthetics/markets/utils.ts
+++ b/src/domain/synthetics/markets/utils.ts
@@ -110,7 +110,7 @@ export function getReservedUsd(marketInfo: MarketInfo, isLong: boolean) {
   }
 }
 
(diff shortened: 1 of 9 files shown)


src/domain/synthetics/markets/utils.ts
```diff
@@ -119,11 +119,15 @@ export function getMaxReservedUsd(marketInfo: MarketInfo, isLong: boolean) {
     ? marketInfo.openInterestReserveFactorLong
```
gdev8317, line 119: thinking if it makes sense to rename `getMaxReservedUsd`, it may not be clear that `maxOpenInterest` key is also checked inside 🤔

gdev8317 (approved):

PR #681, opened 2023-11-08
Separate incentives apr

gmxer:


src/domain/synthetics/common/useIncentiveStats.ts
```diff
@@ -1,22 +1,16 @@
...
-  useEffect(() => {
-    if (!oracleKeeperFetcher) {
-      return;
-    }
-    async function load() {
-      const res = await oracleKeeperFetcher.fetchIncentivesRewards();
-      if (res) {
-        setIncentiveStats(res);
+  return (
+    useSWR<RawIncentivesStats | null>(`incentiveStats-${chainId}`, async () => {
```
gdev8317, line 8: iirc for swr key we usually use array if it consists of multiple components

src/domain/synthetics/common/useIncentiveStats.ts
```diff
@@ -1,22 +1,16 @@
+import useSWR from "swr";
 import { RawIncentivesStats, useOracleKeeperFetcher } from "../tokens";
-import { useEffect, useState } from "react";
 
 export default function useIncentiveStats(chainId: number) {
```
gdev8317, line 4: does it make sense to make chainId optional?

gmxer: Let's do. in all other places we provide chainId. But it might make sense to have overriding arg instead everywhere at some point

gdev8317 (approved):

PR #689, opened 2023-11-13
Fix max liq

vipineth:
[x] Add support for max open interest on market card and utilization

---
  - https://app.asana.com/0/0/1205933447410189


src/components/Synthetics/MarketCard/MarketCard.tsx
```diff
@@ -5,9 +5,11 @@ import Tooltip from "components/Tooltip/Tooltip";
 import {
   MarketInfo,
   getAvailableUsdLiquidityForPosition,
+  getCurrentOpenInterestUsd,
   getMarketIndexName,
   getMarketPoolName,
-  getMaxOpenInterest,
+  getMaxAllowedReservedUsd,
```
gdev8317, line 11: small thing, can we use `getMaxReservedUsd` and `getOpenInterestUsd` to keep it more consistent with contracts code. this is not necessary, but similar names often help to find UI logic that corresponds to contracts logic

vipineth: Updated!

gdev8317 (approved):

gdev8317 (approved):

src/domain/synthetics/markets/utils.ts
```diff
@@ -162,6 +176,31 @@ export function getAvailableUsdLiquidityForCollateral(marketInfo: MarketInfo, is
...
+    };
+  }
+
+  const reservedUsd = getReservedUsd(marketInfo, isLong);
+  const maxReservedUsd = getMaxReservedUsd(marketInfo, isLong);
+
+  const openInterestUsd = getOpenInterestUsd(marketInfo, isLong);
+  const maxOpenInterestUsd = getMaxOpenInterestUsd(marketInfo, isLong);
+
+  const [currentMaxReserve, currentReserve] = getArrayWithSmallerDifference(
```
gdev8317, line 193: `getArrayWithSmallerDifference` is used only in this specific place, i think it's simpler to keep calculation logic here instead of moving to separate function?

src/domain/synthetics/markets/utils.ts
```diff
@@ -162,6 +176,31 @@ export function getAvailableUsdLiquidityForCollateral(marketInfo: MarketInfo, is
...
+    [maxReservedUsd, reservedUsd],
+    [maxOpenInterestUsd, openInterestUsd]
+  );
+
+  return {
+    reserveUsd: currentReserve,
```
gdev8317, line 199: i think name "reserve" can be confusing since it's either "reserve" or "open interest" in reality

gdev8317 (approved):

PR #700, opened 2023-11-17
Update glp fee

vipineth:
[x] Update fee structure for GLP

---
  - https://app.asana.com/0/0/1205902686460234
  - https://app.asana.com/0/0/1205957707764531
  - https://app.asana.com/0/0/1205276763751769


src/components/Banner/BannerButton.tsx
```diff
@@ -0,0 +1,27 @@
+import "./BannerButton.scss";
+import { FiChevronRight } from "react-icons/fi";
+import cx from "classnames";
+
+type Props = {
+  label: string;
```
gmxer, line 6: let's use `children` instead?

gmxer (approved):

gmxer (changes requested):
some quick to fix questions

src/components/Banner/BannerButton.tsx
vipineth: I didn't use children because we need so the sparkle next to the last word. So I just want to make sure the label is a string. Children can also be a ReactNode or Element

gmxer (approved):

PR #780, opened 2024-01-31
Bonus gmx UI fixes

onemicky:
---
  - https://app.asana.com/0/0/1206349481659089

### DIFF
--- a/src/components/ModalInfo/ModalInfo.scss
+++ b/src/components/ModalInfo/ModalInfo.scss
@@ -9,6 +9,10 @@
   &-icon {
     padding-right: 0.6rem;
     padding-top: 0.2rem;
+
+    img {
+      display: block;
+    }
   }
 
   &-text {
--- a/src/components/ModalInfo/ModalInfo.scss
+++ b/src/components/ModalInfo/ModalInfo.scss
@@ -1,0 +1,17 @@
+.ModalInfo.standalone {
+  margin-bottom: 0.8rem;
+}
+
+.ModalInfo {
+  display: flex;
+  margin-bottom: 1.6rem;
+
+  &-icon {
+    padding-right: 0.6rem;
+    padding-top: 0.2rem;
+  }
+
+  &-text {
+    font-size: 1.3rem;
+  }
+}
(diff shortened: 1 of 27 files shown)


src/components/ModalInfo/ModalInfo.scss
```diff
@@ -0,0 +1,21 @@
...
+
+    img {
+      display: block;
+    }
+  }
+
+  &-text {
+    font-size: 1.3rem;
+  }
+}
```
vipineth, line 21: On branch `release-13` there is a component named `AlertInfo` which is coming form the SLTP modal PR. I think `ModalInfo` component can be replaced by the `AlertInfo` component if that's okay.

src/lib/numbers.ts
```diff
@@ -7,6 +7,8 @@ import { getPlusOrMinusSymbol } from "./utils";
 const MAX_EXCEEDING_THRESHOLD = "1000000000";
 const MIN_EXCEEDING_THRESHOLD = "0.01";
 
+export const BIG_ZERO = BigNumber.from(0);
```
vipineth, line 10: I think a better name would be `BN_ZERO` or you can use `ethers.constants.Zero`

vipineth (changes requested):

src/lib/numbers.ts
onemicky: Good thought! Switched to BN_ZERO as the idea was to have a short constant name

src/components/ModalInfo/ModalInfo.scss
onemicky: Hm, there really is a good component with a same functionality. By the way this task is going to be merge directly to master and goes off-release so I don't know a good way to take this component from release-13 don't you?
May we merge this PR as is and i will replace ModalInfo separately after release-13 merging?

vipineth (approved):

src/lib/numbers.ts
gmxer: fyi, we use ZERO in subgraph

image

maybe we can do similar in UI, I'm not sure really though

PR #781, opened 2024-01-31
Introducing selectors with global state

gmxer:
Hi, 
I've prepared some explanations on what's happenning here. Hope it would make sense. I also expect to receive some feedback but would be nice to merge it in ~release-13 to not accumulate diff conflicts.

PR is not that big actually but I've just combined it with eslint changes. It shouldn't break code review, perhaps merging though, but that's covered below.

# Preamble

* We are not working on optimizing website at the moment. Though we should follow all the guidelines and write code optimally but not fixing any not optimal parts at the moment. At some point lack of optimization might be an issue and we should have an ability to fix it.
* Project is growing and we need to have consensus of how we doing new features optimally and new components reusably. And here I introduce global state + selectors for it.
* I believe the only proper way to avoid duplications in frontend proj and also have project working optimally and keep it maintanable is to have a (centralized) state manager. (change_my_mind_meme.jpg)
* Ideally we shouldn't use props as long as we can store them in state for logical components (like Tradebox, TradeFeesRow, PositionLists, PositionItem) except some unique identifier (`positionKey` for `<PositionItem />`) or `place` for some shared warning component. Anyways that's not covered in this PR to the extent I wanted.

# Summary

Changes in this PR:
* More eslint rules
* Introduced selectors (with reselect lib) and "global state" using one big context.
inb4: it basically shares similar concept to what I've done before with `SubaccountContext` recently, using `use-context-selector` lib. So `reselect` lib was just added on top of already existing concept.
* Renamed most of current contract request hooks like `useMarketsInfo` -> `useMarketsInfoRequest` to not get confused with selector-hooks

This is one of several steps to more optimal and organized UI. Not everything was made perfectly from my side. Nether everything was considered properly initially that's why there're some dirty temporary workarounds. But it anyway works and shows the proof of concept.

I'd recommend to review this PR in the following way:
* get familiar with reselect lib (simplest one among many)
* Start with src/context/SyntheticsStateContext/ where all the logic is stored
* Check how it's used in Tradebox, SyntheticsPage and some other giants.
* Get familiar with other eslint changes (probably better check dedicated commits)

# Eslint rules

I didn't introduce any codestyle changes, only react(-perf) recommendations.

Also
```
      "react-perf/jsx-no-new-array-as-prop": "warn",
      "react-perf/jsx-no-new-object-as-prop": "warn",
      "react-perf/jsx-no-new-function-as-prop": "off",
```
You may see I've introduced these three lines. Only two of them are on. That's just because memoizing functions would require much more effort than I expected. I'll try to do it sometime. But as long as we don't care about optimizations at the moment it's not that crucial to fix it asap. Btw that's also one big step to having optimal UI

# Selectors

Idea of selectors is:
* Split logic from layout (jsx is just for html and relevant hooks)
* Reuseable logic of selectors 
* Avoid using many props in a component (simple plain structure of the App and its children)
* Global state that is reused across whole app the same way (no re-implementing/re-retrieving stuff)

There's temprorary workaround though. As not all the components (primarily `PositionEditor`, `PositionSeller`) share they state globally I had to do a little hack with introducing arguments for selectors. That's temporary and not temporary thing at the same time.
Currently all selectors should not need arguments. But they have. e.x. `makeSelectIncreasePositionAmounts`, etc have many arguments because all those props are not in the global state yet.

However in future there anyway would be selectors that unavoidable have arguments. Random example:
imagine `<PositionItem/>` inside `<PositionList/>`). We an't just do `useSelector(selectPositionTotalFees)`. We would need to construct the selector first:
```
// positionHooks.ts
function usePositionTotalFees(positionKey) {
    const selector = useMemo(() => makeSelectPositionTotalFees(position.key), [position.key]);
    return useSelector(selector);
}

// PositionItem.tsx
const fees = usePositionTotalFees(position.key);
```

`makeSelectPositionTotalFees` should be memoizeble by id ofc. I intentionally didn't introduce any memoization for selector factories here as I focused mostly on selectors themselves in this PR.

Hope that makes sense, but I'm very open to discuss it further.

# Noticeable changes / Advantages of the approach

### Commonly used contract calls are just oneliners now:
example
image
currently supported wrapped calls implementations:
image

### Ton of props were elliminated in favor of hook-selectors.
image
image

### Less repeatable code
Before we fetched everything manually (or had to provide it explicitly) https://github.com/gmx-io/gmx-interface/pull/759/files#diff-769dcbf0ef4db331ffc76a453a5bd819540a8d665116cff20ca68cceeed39d60R298
now such call is way smaller.
Ideally it should be just a one liner but I just didn't have enough time, only `Tradebox.tsx` have oneline-complex-selectors for now.

Idea of selectors is that you don't need to write the logic twice. You just reuse selectors that you have already like I do with  `makeSelectNextPositionValues` in this refactored example .

# Next steps (1-2 iterations further, not in the nearest time)
* Use `rereselect` instead of `reselect` (muuch more optimal, more readable code) 
* Get rid of undefineds and `?.`. More strict typecheckers (e.x. `getByKey` -> `getByKeyOrThrow`) and eslint rules. 

I'll let you know in chat in case I have smth to add here. But I hope I've overviewed it enough to get familiar with it.
Thanks for reading that far 🤠 

---
  - https://app.asana.com/0/0/1206463084602534


src/components/Synthetics/GmList/GmList.tsx
```diff
@@ -345,7 +345,7 @@ export function GmList({
                         />
                       </div>
                       <div>
-                        {renderMintableAmount({
+                        {MintableAmount({
```
vipineth, line 348: We can use JSX here instead of function call, helps with the readability.

src/components/Synthetics/TradeBox/TradeBox.tsx
```diff
@@ -150,43 +113,23 @@ const tradeTypeIcons = {
   [TradeType.Swap]: swapImg,
 };
 
+const emptyArray = [];
```
vipineth, line 116: We can use the one defined inside `lib/objects`

vipineth (approved):

src/domain/synthetics/markets/marketsSelector.ts
```diff
@@ -0,0 +1,3 @@
+export function hello() {
```
onemicky, line 1: Left by chance after the esLint test?

onemicky (approved):

onemicky (approved):

onemicky (approved):

PR #784, opened 2024-02-01
fix slide keeps dragging after dropped on iframe

vipineth:
[x] Fix slide when dropped on iframe (chart) it keeps dragging

---
  - https://app.asana.com/0/0/1204545702859373

### DIFF
--- a/src/components/LeverageSlider/LeverageSlider.tsx
+++ b/src/components/LeverageSlider/LeverageSlider.tsx
@@ -83,7 +83,11 @@ function LeverageSliderHandle({ value, dragging, index, keyValueMap, ...restProp
   const displayValue = keyValueMap[value || 0] ?? DEFAULT_LEVERAGE_KEY;
 
   useEffect(() => {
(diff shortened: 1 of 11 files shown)


src/components/LeverageSlider/LeverageSlider.tsx
```diff
@@ -81,6 +81,11 @@ export function LeverageSlider(p: Props) {
 
 function LeverageSliderHandle({ value, dragging, index, keyValueMap, ...restProps }: HandleProps) {
   const displayValue = keyValueMap[value || 0] ?? DEFAULT_LEVERAGE_KEY;
+
+  useEffect(() => {
+    document.body.classList.toggle("dragging", dragging);
```
onemicky, line 86: I worry about what might happen if the component were initially rendered with the dragging={false} prop. There could be desynchronization between the prop value and the class. What if we better add and remove the class name explicitly?

onemicky (dismissed):

onemicky (approved):

PR #794, opened 2024-02-06
High network fee

gmxer:
Not super ellegant I guess as I do both rendering and logic in [useHighExecutionFeeAcknowledgement.tsx](https://github.com/gmx-io/gmx-interface/compare/high-network-fee?expand=1#diff-1707110ffac98e372ef96a7c9609e02cf7e11250d63ca45505ea48e673fd2d58) but I think it's better to keep the logic not separated there

---
  - https://app.asana.com/0/0/1206188451069039


src/domain/synthetics/trade/useHighExecutionFeeAcknowledgement.tsx
```diff
@@ -0,0 +1,31 @@
...
+      <div>
+        <Checkbox asRow isChecked={isHighExecutionFeeAccepted} setIsChecked={setIsHighExecutionFeeAccepted}>
+          <span className="text-warning font-sm">
+            <Trans>Acknowledge very high network Fees</Trans>
+          </span>
+        </Checkbox>
+      </div>
+    ) : null,
+  };
+}
```
vipineth, line 31: Few suggestions (only if you think it will improve the readability of the hook)

- Using the name `useHighExecutionFeeConsent` (a bit smaller and conveys the exact same meaning)
- The name`highExecutionFeeNotAcceptedError` signifies it will return some text but it will return boolean so maybe call it something like `isHighFeeConsentPending` or `isConsentRequiredAndPending` (if I understand it correctly)

vipineth (changes requested):

src/domain/synthetics/trade/useHighExecutionFeeAcknowledgement.tsx
gmxer: 1. fixed
2. changed to isHighFeeConsentError

vipineth (approved):

PR #806, opened 2024-02-08
One click trading part 2

onemicky:


src/components/Synthetics/SubaccountModal/SubaccountModal.tsx
```diff
@@ -181,13 +191,13 @@ const MainView = memo(({ setPendingTxns }: { setPendingTxns: (txns: any) => void
...
     value: maxAutoTopUpAmount,
-  } = useBigNumberInput(null, nativeToken.decimals, 4);
+  } = useBigNumberInput(null, wrappedToken.decimals, 4);
   const {
     displayValue: wntForAutoTopUpsString,
     setDisplayValue: setWntForAutoTopUpsString,
     setValue: setWntForAutoTopUps,
     value: wntForAutoTopUps,
-  } = useBigNumberInput(null, wrappedToken.decimals, 4);
+  } = useBigNumberInput(null, nativeToken.decimals, 4);
```
gmxer, line 200: this is most likely won't change behavior as wnt and nt have same decimals in arbitrum and avalanche but I'd revert it as it confuses. Or am I missing smth?

gmxer (approved):

src/components/Synthetics/SubaccountModal/SubaccountModal.tsx
onemicky: This change is related to this subtask: https://app.asana.com/0/1204313444805313/1206626863771251
Xhiroz discovered that the proper token should be ETH or AVAX instead of WETH or WAVAX here. So I replaced the wrappedToken to a nativeToken across all related places. However i understand that this particular change wouldn't has any effect, it looks more consistent according to other rows.

gmxer: ok thanks

PR #817, opened 2024-02-13
Fix tooltip position

vipineth:
[x] Add support to auto flip the tooltip placement based on the available space.
[x] Refactor the tooltip position names and make it a bit more standard.

Supported Positions:
```ts
type TooltipPosition =
  | "top"
  | "top-start"
  | "top-end"
  | "right"
  | "right-start"
  | "right-end"
  | "bottom"
  | "bottom-start"
  | "bottom-end"
  | "left"
  | "left-start"
  | "left-end";
```
[x] Updated old position names to new names

[x] Add support for `maxAllowedWidth`

In this PR, we've added support for prop named `maxAllowedWidth` to the `TooltipWithPortal` and `Tooltip` component, allowing tooltips to adjust their width dynamically within the available space, maintaining a `10px` padding from screen edges. This ensures tooltips adapt to different screen sizes without overflowing, enhancing UI responsiveness.

---
  - https://app.asana.com/0/0/1206207862538937


src/components/Tooltip/Tooltip.scss
midas-myth, line 1: Huge work done, super cool.

One thing is as we are starting to use floating-ui, why do we need this styles for positioning, floating-ui has its own styles, logic, for example useFloating uses prop, position that automagically places the item.

midas-myth (approved):

PR #849, opened 2024-03-07
Refactor redirectPopupTimestamp serialization and deserialization

midas-myth:
---
  - https://app.asana.com/0/0/1206777951596280

### DIFF
--- a/src/App/App.js
+++ b/src/App/App.js
@@ -156,7 +156,21 @@ function FullApp({ pendingTxns, setPendingTxns }) {
   );
   const [redirectModalVisible, setRedirectModalVisible] = useState(false);
   const [shouldHideRedirectModal, setShouldHideRedirectModal] = useState(false);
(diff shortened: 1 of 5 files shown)


src/App/App.js
```diff
@@ -156,7 +156,21 @@ function FullApp({ pendingTxns, setPendingTxns }) {
...
+      const num = parseInt(val);
+
+      if (Number.isNaN(num)) {
+        return undefined;
+      }
+
+      return num;
+    },
+    serializer: (val) => (val ? val.toString() : undefined),
+  });
```
gmxer, line 173: but what happened? is there some NaN got saved?)

gmxer (approved):

src/App/App.js
midas-myth: initially there would be no value, but it sstill would be treated as number, i think it worked only by luck and js type coersion

PR #856, opened 2024-03-11
Refactor wallet-related code and remove unused files

midas-myth:
---
  - https://app.asana.com/0/0/1206399503398381


gmxer (changes requested):
I don't have any suggestions here, all seems good
Besides 1 typo and bug I've mentioned in chat

PR #875, opened 2024-03-19
Fix pool selector hidden chevron

midas-myth:
telegram-cloud-photo-size-2-5460740185026319899-y

now chevron hides only when nessessary

---
  - https://app.asana.com/0/0/1206433988899249

### DIFF
--- a/src/components/Modal/Modal.tsx
+++ b/src/components/Modal/Modal.tsx
@@ -54,7 +54,7 @@ export default function Modal(props: ModalProps) {
 
   const style = useMemo(() => ({ zIndex }), [zIndex]);
 
-  const stopPropagation = useCallback((e: React.MouseEvent | React.KeyboardEvent) => {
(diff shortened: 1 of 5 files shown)


src/components/Modal/Modal.tsx
```diff
@@ -54,6 +54,10 @@ export default function Modal(props: ModalProps) {
 
   const style = useMemo(() => ({ zIndex }), [zIndex]);
 
+  const stopPropagation = useCallback((e: React.MouseEvent | React.KeyboardEvent) => {
```
gmxer, line 57: in your case it's just a mouse event

gmxer (approved):

src/components/Modal/Modal.tsx
midas-myth: fixed

gmxer (approved):

PR #883, opened 2024-03-22
Fix data TradingView data feed provider

midas-myth:
---
  - https://app.asana.com/0/0/1205118053791993

### DIFF
--- a/src/domain/prices.ts
+++ b/src/domain/prices.ts
@@ -9,13 +9,14 @@ import { chainlinkClient } from "lib/subgraph/clients";
 import { sleep } from "lib/sleep";
 import { formatAmount } from "lib/numbers";
 import { getNativeToken, getNormalizedTokenSymbol, isChartAvailabeForToken } from "config/tokens";
+import type { Bar, FromOldToNewArray } from "./tradingview/types";
 
 const BigNumber = ethers.BigNumber;
 
 // Ethereum network, Chainlink Aggregator contracts
 const FEED_ID_MAP = {
   BTC_USD: "0xae74faa92cb67a95ebcab07358bc222e33a34da7",
-  ETH_USD: "0x37bc7498f4ff12c19678ee8fe19d713b87f6a9e6",
+  ETH_USD: "0xE62B71cf983019BFf55bC83B48601ce8419650CC".toLowerCase(),
   BNB_USD: "0xc45ebd0f901ba6b2b8c7e70b717778f055ef5e6d",
   LINK_USD: "0xdfd03bfc3465107ce570a0397b247f546a42d0fa",
   UNI_USD: "0x68577f915131087199fe48913d8b416b3984fd38",
@@ -25,12 +26,20 @@ const FEED_ID_MAP = {
   YFI_USD: "0x8a4d74003870064d41d4f84940550911fbfccf04",
   SPELL_USD: "0x8640b23468815902e011948f3ab173e1e83f9879",
 };
-export const timezoneOffset = -new Date().getTimezoneOffset() * 60;
+export const TIMEZONE_OFFSET_SEC = -new Date().getTimezoneOffset() * 60;
 
-function formatBarInfo(bar) {
-  const { t, o: open, c: close, h: high, l: low } = bar;
+type CompactBar = {
+  t: number;
+  o: number;
+  c: number;
+  h: number;
+  l: number;
+};
+
+function formatBarInfo(compactBar: CompactBar): Bar {
+  const { t, o: open, c: close, h: high, l: low } = compactBar;
   return {
-    time: t + timezoneOffset,
+    time: t + TIMEZONE_OFFSET_SEC,
     open,
     close,
     high,
@@ -68,7 +77,12 @@ export function fillGaps(prices, periodSeconds) {
   return newPrices;
 }
 
-export async function getLimitChartPricesFromStats(chainId, symbol, period, limit = 1) {
+export async function getLimitChartPricesFromStats(
+  chainId: number,
+  symbol: string,
+  period: string,
+  limit = 1
+): Promise<FromOldToNewArray<Bar>> {
   symbol = getNormalizedTokenSymbol(symbol);
 
   if (!isChartAvailabeForToken(chainId, symbol)) {
@@ -91,7 +105,11 @@ export async function getLimitChartPricesFromStats(chainId, symbol, period, limi
   }
 }
 
-export async function getChartPricesFromStats(chainId, symbol, period) {
+export async function getChartPricesFromStats(
+  chainId: number,
+  symbol: string,
+  period: string
+): Promise<FromOldToNewArray<Bar>> {
   symbol = getNormalizedTokenSymbol(symbol);
 
   const timeDiff = CHART_PERIODS[period] * 3000;
@@ -123,10 +141,10 @@ export async function getChartPricesFromStats(chainId, symbol, period) {
   if (!res.ok) {
     throw new Error(`request failed ${res.status} ${res.statusText}`);
   }
-  const json = await res.json();
-  let prices = json?.prices;
-  if (!prices || prices.length < 1) {
-    throw new Error(`not enough prices data: ${prices?.length}`);
+  const json: { prices: CompactBar[]; updatedAt: number; period: string } = await res.json();
+  let compactPrices = json?.prices;
+  if (!compactPrices || compactPrices.length < 1) {
+    throw new Error(`not enough prices data: ${compactPrices?.length}`);
   }
 
   const OBSOLETE_THRESHOLD = Date.now() / 1000 - 60 * 30; // 30 min ago
@@ -140,7 +158,7 @@ export async function getChartPricesFromStats(chainId, symbol, period) {
     );
   }
 
-  prices = prices.map(formatBarInfo);
+  const prices = compactPrices.map(formatBarInfo);
 
   return prices;
 }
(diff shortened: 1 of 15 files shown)


src/domain/prices.ts
```diff
@@ -9,13 +9,14 @@ import { chainlinkClient } from "lib/subgraph/clients";
...
 import { getNativeToken, getNormalizedTokenSymbol, isChartAvailabeForToken } from "config/tokens";
+import type { Bar, FromOldToNewArray } from "./tradingview/types";
 
 const BigNumber = ethers.BigNumber;
 
 // Ethereum network, Chainlink Aggregator contracts
 const FEED_ID_MAP = {
   BTC_USD: "0xae74faa92cb67a95ebcab07358bc222e33a34da7",
-  ETH_USD: "0x37bc7498f4ff12c19678ee8fe19d713b87f6a9e6",
+  ETH_USD: "0xE62B71cf983019BFf55bC83B48601ce8419650CC".toLowerCase(),
```
gmxer, line 19: let's just lowercase it?

src/domain/tradingview/TVDataProvider.ts
```diff
@@ -115,22 +120,33 @@ export class TVDataProvider {
...
     const { from, to } = periodParams;
 
     try {
-      const bars = isStable
+      const bars: FromOldToNewArray<Bar> = isStable
         ? getStableCoinPrice(period, from, to)
         : await this.getTokenHistoryBars(chainId, ticker, period, periodParams);
 
-      return bars.map(formatTimeInBarToMs);
+      return bars.map(formatTimeInBarToMs) as FromOldToNewArray<Bar>;
```
gmxer, line 138: why do you need `as` here if `bars` already has this type?

src/domain/synthetics/tradingview/SyntheticsTVDataProvider.ts
```diff
@@ -1,43 +1,50 @@
...
 
-  override async getLimitBars(chainId: number, ticker: string, period: string, limit: number) {
-    return this.oracleKeeperFetcher.fetchOracleCandles(ticker, period, limit);
+  override async getLimitBars(
+    _chainId: number,
+    ticker: string,
+    period: string,
+    limit: number
+  ): Promise<FromOldToNewArray<Bar>> {
+    const limitBars = (await this.oracleKeeperFetcher.fetchOracleCandles(ticker, period, limit)).reverse();
```
gmxer, line 46: was `reverse` an issue?
just thinking if we could sort by time here? maybe even having a helper function that receives an array and returns `AscOrderArray`

src/domain/tradingview/useTVDatafeed.ts
```diff
@@ -1,20 +1,29 @@
...
-function getConfigurationData(supportedResolutions) {
+function getConfigurationData(supportedResolutions): DatafeedConfiguration {
   return {
     supported_resolutions: Object.keys(supportedResolutions),
     supports_marks: false,
     supports_timescale_marks: false,
     supports_time: true,
     reset_cache_timeout: 100,
-  };
+  } as unknown as DatafeedConfiguration;
```
gmxer, line 26: let's do rather smth like
image

also `ResolutionString` looks weird type, thinking if we could just use `string` instead?
And reset_cache_timeout is missing in `DatafeedConfiguration`
image

src/domain/tradingview/useTVDatafeed.ts
```diff
@@ -86,108 +95,195 @@ export default function useTVDatafeed({ dataProvider }: Props) {
...
-          }
-        },
-        async subscribeBars(
-          symbolInfo: SymbolInfo,
-          resolution: ResolutionString,
-          onRealtimeCallback: SubscribeBarsCallback
-        ) {
-          const period = supportedResolutions[resolution];
-          const { ticker, isStable } = symbolInfo;
+export function buildFeeder({
```
gmxer, line 111: no need for `export`

src/domain/tradingview/useTVDatafeed.ts
```diff
@@ -86,108 +95,195 @@ export default function useTVDatafeed({ dataProvider }: Props) {
...
+
+  if (intervalRef.current !== undefined) {
+    clearInterval(intervalRef.current);
+    intervalRef.current = undefined;
+  }
+
+  const handleInterval = () => {
+    if (missingBarsInfoRef.current.isFetching || !feedDataRef.current) return;
+    if (missingBarsInfoRef.current.bars?.length > 0) {
+      let processedBarTimes: FromOldToNewArray<number> = [];
```
gmxer, line 259: const

gmxer: also to speed up a bit, we could use `Set<number>`. You would perhaps need `latestBarTime` var to replace `processedBarTimes.at(-1)`.

But not worth it if `missingBarsInfoRef.current.bars.length` isn't that long

src/domain/tradingview/useTVDatafeed.ts
```diff
@@ -86,108 +95,195 @@ export default function useTVDatafeed({ dataProvider }: Props) {
...
+      missingBarsInfoRef.current.bars.forEach((bar: Bar, index) => {
+        if (processedBarTimes.includes(bar.time)) return;
+        if (index !== 0 && bar.time < processedBarTimes.at(-1)!) {
```
gmxer, line 262: could you explain?
so `missingBarsInfoRef.current.bars` should go in descending order? can't we just sort them?

src/domain/tradingview/useTVDatafeed.ts
```diff
@@ -86,108 +95,195 @@ export default function useTVDatafeed({ dataProvider }: Props) {
...
+            "missingBarsInfoRef bars order is violating TradingView api schema. See https://www.tradingview.com/charting-library-docs/latest/connecting_data/Datafeed-API#subscribebars"
+          );
+        }
+
+        onRealtimeCallback(formatTimeInBarToMs(bar));
+        processedBarTimes.push(bar.time);
+      });
+      missingBarsInfoRef.current.bars = [];
+    } else {
+      tvDataProviderRef.current?.getLiveBar(chainId, ticker, period).then((bar) => {
```
gmxer, line 273: we call this line by 500ms interval. Can't it be that `getLiveBar` responds in more than 500ms and thus would mess up bars?

src/domain/tradingview/types.ts
```diff
@@ -8,3 +8,26 @@ export type Bar = BarType & {
...
+}
+type Flavor<T, FlavorT> = T & Flavoring<FlavorT>;
+
+/**
+ * This type ensures type safety when dealing with ordered arrays.
+ * @see {@link FromNewToOldArray}
+ */
+export type FromOldToNewArray<T> = Flavor<Omit<T[], "reverse">, "FromOldToNewArray"> & {
+  reverse(): FromNewToOldArray<T>;
+};
```
gmxer, line 23: maybe `AscOrderArray`/`DescOrderArray`?

from the first glance it looks like a helper type

gmxer (changes requested):

src/domain/tradingview/TVDataProvider.ts
midas-myth: map removes the flavored type back to T[] instead of F<T[]>

src/domain/synthetics/tradingview/SyntheticsTVDataProvider.ts
midas-myth: most probably yes, reverse was an issue.
Here reversing == sorting by time

src/domain/prices.ts
midas-myth: fixed

src/domain/tradingview/useTVDatafeed.ts
midas-myth: 1) fixed
2) added to ts

src/domain/tradingview/useTVDatafeed.ts
midas-myth: fixed

src/domain/tradingview/useTVDatafeed.ts
midas-myth: 1) fixed
2) agree about length

src/domain/tradingview/useTVDatafeed.ts
midas-myth: 1) We sort them multiple times throughout the pipeline. Would not want to add another sorting. The problem was in the order the `onRealtimeCallback` was called
2) ascending.

src/domain/tradingview/useTVDatafeed.ts
midas-myth: but it later checks that 
```
          bar.ticker === tvDataProviderRef.current?.currentTicker &&
          bar.period === tvDataProviderRef.current?.currentPeriod
```
it seems to take that into acc?

midas-myth: I will add
```
bar.time >= lastBarTimeRef.current
```
but im not sure it will do anything

src/domain/tradingview/types.ts
midas-myth: for me its just not that obviuos that asc = from old to new. I still have to spend a second or two to decode it to what it really means

gmxer (approved):

src/domain/tradingview/types.ts
gmxer: ok

src/domain/tradingview/useTVDatafeed.ts
gmxer: try intentionally slowdown getLiveBar to ~1 second resp time. If it doesn't affect badly than it's fine like it's

gmxer (approved):

PR #922, opened 2024-04-22
fix infinite trade redirect

gmxer:
tradePageVersion now only depends on a browser url, not on an internal state, i.e. there's only 1 point of truth.

---
  - https://app.asana.com/0/0/1207092385792627


src/context/GlobalContext/GlobalContextProvider.tsx
```diff
@@ -1,14 +1,25 @@
...
 import { useChainId } from "lib/chains";
 import { useLocalStorageSerializeKey } from "lib/localStorage";
 import { PendingTransaction, SetPendingTransactions } from "domain/legacy";
 import { useLocalStorage } from "react-use";
+import { matchPath, useHistory, useLocation } from "react-router-dom";
 
 type GlobalContextType = null | {
-  tradePageVersion: number;
-  setTradePageVersion: Dispatch<SetStateAction<number | undefined>>;
+  tradePageVersion: any;
```
onemicky, line 21: let's fix types here

onemicky (changes requested):

src/context/GlobalContext/GlobalContextProvider.tsx
gmxer: fixed

onemicky (dismissed):

onemicky (approved):

PR #938, opened 2024-05-07
update dao wording

onemicky:


src/App/App.tsx
```diff
@@ -406,7 +406,6 @@ function FullApp() {
         closeOnClick={false}
         draggable={false}
         pauseOnHover
-        icon={false}
```
onemicky, line 409: catched a linter error like "prop icon doesn't exist"

gmxer (approved):

PR #970, opened 2024-05-29
max lev 100x announcement

gmxer:


gmxer:
will be deployed together with gmxusdc-market branch

PR #1020, opened 2024-06-12
Order positions by available markets, sort descending by Size within the same market

hubert-de-lalye:
1. I attempted to break down usage of [useAvailableTokenOptions](https://github.com/gmx-io/gmx-interface/blob/master/src/domain/synthetics/trade/useAvailableTokenOptions.ts) to separate selectors, but found that it widely used [in building of trade state](https://github.com/gmx-io/gmx-interface/blob/master/src/domain/synthetics/trade/useTradeboxState.ts#L79). Ideal solution seems to in in removing redundant useSomthings which are not making any requests but just mapping the data for building state and replace them by clear selectors, but it requires a massive refactor, so I retreated.

2. Tried to move all positions-specific selectors to a separate file, but it will create a triangle of circular deps, so I didn't change much, just add new file for complex selectors. I opened to discuss and change it.

Before:
Screenshot 2024-06-12 at 16 28 42

After:
Screenshot 2024-06-12 at 16 26 17

---
  - https://app.asana.com/0/0/1207057109124491

### DIFF
--- a/src/context/SyntheticsStateContext/selectors/positionsSelectors.ts
+++ b/src/context/SyntheticsStateContext/selectors/positionsSelectors.ts
@@ -1,11 +1,11 @@
 import { createSelector } from "../utils";
 import { selectPositionsInfoData } from "./globalSelectors";
-import { selectTradeboxMarketsOrder } from "./tradeboxSelectors";
+import { selectTradeboxMarketsOrderMap } from "./tradeboxSelectors";
 import { bigintToNumber } from "lib/numbers";
 
 export const selectPositionsInfoDataSortedByMarket = createSelector((q) => {
   const positionsInfoData = q(selectPositionsInfoData);
-  const marketsOrder = q(selectTradeboxMarketsOrder);
+  const marketsOrder = q(selectTradeboxMarketsOrderMap);
 
   const positions = Object.values(positionsInfoData || {});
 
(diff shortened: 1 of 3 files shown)


src/context/SyntheticsStateContext/selectors/positionsSelectors.ts
```diff
@@ -0,0 +1,23 @@
...
+  const marketsOrder = q(selectTradeboxMarketsOrderMap);
+
+  const positions = Object.values(positionsInfoData || {});
+
+  const sortedPositions = positions.sort((a, b) => {
+    const bMarketIdx = marketsOrder[a.indexToken.symbol];
+    const aMarketIdx = marketsOrder[b.indexToken.symbol];
+
+    if (aMarketIdx === bMarketIdx) {
+      return bigintToNumber(b.sizeInUsd - a.sizeInUsd, 1);
```
midas-myth, line 17: hmmmm,
are not we afraid that this might overflow ?
maybe just A - B > 0n ? 1 : -1 ?

src/context/SyntheticsStateContext/selectors/tradeboxSelectors/index.ts
```diff
@@ -774,3 +774,12 @@ export const selectTradeboxLeverageSliderMarks = createSelector((q) => {
   const maxAllowedLeverage = q(selectTradeboxMaxLeverage);
   return getTradeboxLeverageSliderMarks(maxAllowedLeverage);
 });
+
+export const selectTradeboxMarketsOrderMap = createSelector((q) => {
+  const { sortedAllMarkets } = q(selectTradeboxAvailableTokensOptions);
+
+  return sortedAllMarkets.reduce((acc, market, idx) => {
```
midas-myth, line 781: I am not sure but lets try to keep keys as addresses where possible

beware that there are native tokens and non-native tokens so be careful when using a.indexToken and indexTokenAddress

hubert-de-lalye: agree, fixed

midas-myth (approved):

PR #1045, opened 2024-06-20
fix subaccount withdrawal

onemicky:
---
  - https://app.asana.com/0/0/1207589010234688

### DIFF
--- a/src/context/SubaccountContext/SubaccountContext.tsx
+++ b/src/context/SubaccountContext/SubaccountContext.tsx
@@ -294,7 +294,7 @@ export function useSubaccountDefaultExecutionFee() {
   return useSubaccountSelector((s) => s.defaultExecutionFee) ?? 0n;
 }
 
-function useSubaccountDefaultNetworkFee() {
+export function useSubaccountDefaultNetworkFee() {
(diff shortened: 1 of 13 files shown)


src/context/SubaccountContext/SubaccountContext.tsx
```diff
@@ -367,7 +368,7 @@ export function useSubaccount(requiredBalance: bigint | null, requiredActions =
   ]);
 }
 
-export function useSubaccountInsufficientFunds(requiredBalance: bigint | undefined | null) {
+export function useSubaccountInsufficientFunds(requiredBalance?: bigint | null) {
```
gmxer, line 371: my intention was to always provide argument to not forget pass balance where possible

but maybe that's a silly way, and it's fine to omit it

gmxer (dismissed):

src/context/SubaccountContext/SubaccountContext.tsx
onemicky: Oh got it. Then I guess it may be better to force passing the required Balance even if it is default. In such case it should be manually received it outside of the function.

gdev8317 (approved):

PR #1078, opened 2024-07-02
execute order optimisation

onemicky:
---
  - https://app.asana.com/0/0/1206754384322333


src/domain/synthetics/orders/simulateExecuteOrderTxn.tsx
```diff
@@ -58,7 +69,7 @@ export async function simulateExecuteOrderTxn(chainId: number, p: SimulateExecut
         maxTimestamp: priceTimestamp,
       },
       ...(p.extraArgs ?? []),
-    ]),
+    ] as any),
```
hubert-de-lalye, line 72: Not quite sure about this `any`s, as I'm a bit out of context with typechain. Is it possible to fix, or at least mitigate possible following issues if any?

onemicky: Good point, thanks. Made types more strict

onemicky:
Also added handy getters for all contracts we use:
```
  const dataStore = getDataStoreContract(chainId, provider);
  const multicall = getMulticallContract(chainId, provider);
  const exchangeRouter = getExchangeRouterContract(chainId, provider);
  ...
```

hubert-de-lalye (dismissed):

src/config/contracts.ts
```diff
@@ -314,3 +317,15 @@ export function getContract(chainId: number, name: string): string {
 
   return CONTRACTS[chainId][name];
 }
+
+function makeGetContract<T extends unknown & { abi: InterfaceAbi; connect: (address: string) => unknown }>(
```
hubert-de-lalye, line 321:
```suggestion
function makeGetContract<T extends { abi: InterfaceAbi; connect: (address: string) => unknown }>(
```

`unknown & {...}` is equals to `{...}` as unknown a supertype of any type, so there's no necessity to use unknown

hubert-de-lalye (dismissed):

hubert-de-lalye (approved):

PR #1104, opened 2024-07-17
Batch multicalls

midas-myth:
---
  - https://app.asana.com/0/0/1207723133359558
  - https://app.asana.com/0/0/1208009601938866


src/domain/synthetics/markets/useMarketsInfoRequest.ts
```diff
@@ -61,26 +61,275 @@ export type MarketsInfoResult = {
...
+    | "claimableFundingAmountLong"
+    | "claimableFundingAmountShort"
+    | "borrowingFactorPerSecondForLongs"
+    | "borrowingFactorPerSecondForShorts"
+    | "fundingFactorPerSecond"
+    | "longsPayShorts"
+    | "virtualPoolAmountForLongToken"
+    | "virtualPoolAmountForShortToken"
+    | "virtualInventoryForPositions"
+    | "netPnlMin"
```
gdev8317, line 97: it is already defined on line 87

midas-myth: fixed

src/domain/tokens/useInfoTokens.ts
```diff
@@ -47,7 +47,7 @@ export function useInfoTokens(
   const { data: indexPrices } = useSWR(indexPricesUrl, {
     // @ts-ignore spread args incorrect type
     fetcher: (url) => fetch(url).then((res) => res.json()),
-    refreshInterval: 500,
+    refreshInterval: 1000,
```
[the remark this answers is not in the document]
midas-myth: Do we want to replace all occurrences of 1000 / 5000 / 60000 milliseconds with such constants?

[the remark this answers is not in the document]
midas-myth: Agree, done

gmxer (dismissed):

src/domain/synthetics/markets/useMarketsInfoRequest.ts
```diff
@@ -61,26 +62,281 @@ export type MarketsInfoResult = {
...
+  tokensData: TokensData | undefined;
+}) {
+  const dataStoreAddress = getContract(chainId, "DataStore");
 
-  const { data } = useMulticall(chainId, "useMarketsInfo", {
-    inWorker: true,
-    key: !isDepencenciesLoading &&
-      marketsAddresses.length > 0 && [marketsAddresses.join("-"), dataStoreAddress, account, pricesUpdatedAt],
+  const marketsValuesQuery = useMulticall(chainId, "useMarketsValuesRequest", {
+    key: !isDependenciesLoading && marketsAddresses!.length > 0 && [marketsAddresses, account],
```
gmxer, line 331: not sure if it's a critical or not but let's do `marketsAddresses.join('-')` just in case

src/lib/contracts/contractFetcher.ts
```diff
@@ -81,29 +109,93 @@ export const contractFetcher =
...
+
+      let timeout = 2000;
+
+      if (isThroughMulticall) {
+        if (isUrgent) {
+          const workerStartupTimeout = 1000;
+          timeout = timeout + workerStartupTimeout;
+        } else {
+          const workerTimeout = 5000;
+          const mainThreadTimeout = 5000;
```
gmxer, line 123: logic makes sense to me
just thinking 
- what if we do all urgent calls in main thread. that way we would save this startup timeout time. Not a priority for this iteration but rather a point to think
- code is self explaining but should we put numbers into constants? maybe not, not sure really.

src/lib/multicall/executeMulticall.ts
```diff
@@ -0,0 +1,274 @@
...
+      if (!call) {
+        continue;
+      }
+
+      // To reduce duplicate calls, we hash the call data and use it as a unique identifier.
+      // There are two main reasons for this:
+      // 1. Single token backed pools have many pairs with the same method signatures
+      // 2. The majority of pools have USDC as the short token, which means they all have some common calls
+
+      const callId = stableHash([callGroup.contractAddress, call.methodName, call.params]);
```
gmxer, line 211: just in case, have you measured this call?

src/lib/multicall/executeMulticallWorker.ts
```diff
@@ -44,20 +51,25 @@ export async function executeMulticallWorker(chainId: number, request: Multicall
     PRODUCTION_PREVIEW_KEY: localStorage.getItem(PRODUCTION_PREVIEW_KEY),
   });
 
-  const { promise, resolve, reject } = promiseWithResolvers();
+  const { promise, resolve, reject } = promiseWithResolvers<MulticallResult<any> | undefined>();
   promises[id] = { resolve, reject };
 
-  const escapePromise = sleep(2_000).then(() => "timeout");
+  const internalMulticallTimeout = MAX_TIMEOUT;
+  const bufferTimeout = 500;
```
gmxer, line 58: is it FREQUENT_UPDATE_INTERVAL - FREQUENT_MULTICALL_REFRESH_INTERVAL
or where it come from?

gmxer (changes requested):

src/domain/synthetics/markets/useMarketsInfoRequest.ts
midas-myth: for swr stableHash library is used under the hood. so i dont see the reason to keep serializing the values in keys. whats your concert?

src/lib/contracts/contractFetcher.ts
midas-myth: I think this place should be changed in part 2 as we plan to change the fallback logic

src/lib/multicall/executeMulticall.ts
midas-myth: no i did not.
but there is no need.
use swr uses it. In reality its not a hash. Just a concatenated and serialized view. concatenating 3-8 primitives is fast

src/lib/multicall/executeMulticallWorker.ts
midas-myth: its just a small gap to ensure that worker (that has MAX_TIMEOUT timeout inside) finishes calling public rpc. For example if worker did not start because of being busy by synchronous tasks for a little.

src/lib/contracts/contractFetcher.ts
midas-myth: If we do urgent all urgen calls in main thread we will encouter more stuttering during page load. as the first render causes a lot of simultanious urgent calls

src/lib/multicall/executeMulticall.ts
midas-myth: image
image

src/lib/contracts/contractFetcher.ts
midas-myth: moved to constant. but the archietcture will change i hope

src/lib/multicall/executeMulticall.ts
midas-myth: https://github.com/vercel/swr/blob/1585a3e37d90ad0df8097b099db38f1afb43c95d/src/_internal/utils/serialize.ts

https://github.com/vercel/swr/blob/1585a3e37d90ad0df8097b099db38f1afb43c95d/src/core/use-swr.ts#L118

gmxer (approved):

PR #1111, opened 2024-07-19
fix promise with resolvers

gmxer:


src/lib/utils.ts
```diff
@@ -16,3 +16,14 @@ export function getPlusOrMinusSymbol(value?: bigint, opts: { showPlusForZero?: b
   const { showPlusForZero = false } = opts;
   return value === 0n ? (showPlusForZero ? "+" : "") : value < 0n ? "-" : "+";
 }
+
+export function promiseWithResolvers() {
```
hubert-de-lalye, line 20: if it's long-term solution then let's make it generic to not loose resolving type

_skip if it's just a patch for hotfix_

hubert-de-lalye (approved):

PR #1113, opened 2024-07-22
Tradebox switch logic update v2

onemicky:
---
  - https://app.asana.com/0/0/1207300395314351

### DIFF
--- a/src/domain/synthetics/trade/useTradeboxState.ts
+++ b/src/domain/synthetics/trade/useTradeboxState.ts
@@ -9,7 +9,7 @@ import {
   getLeverageKey,
   getSyntheticsTradeOptionsKey,
 } from "config/localStorage";
-import { getTokenBySymbolSafe, getToken, isSimilarToken } from "config/tokens";
+import { getToken, isSimilarToken } from "config/tokens";
 import { createTradeFlags } from "context/SyntheticsStateContext/selectors/tradeSelectors";
 import { createGetMaxLongShortLiquidityPool } from "context/SyntheticsStateContext/selectors/tradeboxSelectors";
 import { getIsUnwrap, getIsWrap } from "domain/tokens";
@@ -707,14 +707,15 @@ function fallbackPositionTokens(
     if (nextPayTokenAddress) {
       const desirablePayToken = getToken(chainId, nextPayTokenAddress);
 
-      if (desirablePayToken && desirablePayToken.symbol) {
-        const nonSyntheticToken = getTokenBySymbolSafe(chainId, desirablePayToken.symbol, {
-          isSynthetic: false,
-          symbolType: "baseSymbol",
+      if (desirablePayToken) {
+        const similarPayTokenAddress = Array.from(allowedPayTokensSet).find((m) => {
+          const indexToken = getToken(chainId, m);
+
+          return isSimilarToken(indexToken, desirablePayToken);
         });
 
-        if (nonSyntheticToken && allowedPayTokensSet.has(nonSyntheticToken.address)) {
-          fallbackPayTokenAddress = nonSyntheticToken.address;
+        if (similarPayTokenAddress) {
+          fallbackPayTokenAddress = similarPayTokenAddress;
         }
       }
     }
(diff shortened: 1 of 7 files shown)


src/domain/synthetics/trade/useTradeboxState.ts
```diff
@@ -705,10 +801,10 @@ function fallbackCollateralTokens(newState: StoredTradeOptions, marketsInfoData:
       collateralAddress = marketInfo.shortTokenAddress;
     }
 
-    set(newState, ["collaterals", marketAddress, longOrShort], collateralAddress);
+    return produce(newState, (draft) => {
+      set(draft, ["collaterals", marketAddress, longOrShort], collateralAddress);
+    });
```
midas-myth, line 806: i dont really understand the benefit of using this whole creating proxy setter with immer to just set one field

src/domain/synthetics/trade/useTradeboxState.ts
```diff
@@ -621,62 +648,131 @@ function setToTokenAddressUpdaterBuilder(
...
+
+  const isNextPayTokenValid = nextPayTokenAddress && allowedPayTokensSet.has(nextPayTokenAddress);
+  const isNextIndexTokenValid = nextIndexTokenAddress && allowedIndexTokens.has(nextIndexTokenAddress);
+  const isNextMarketTokenValid =
+    nextMarketTokenAdress && marketsMap[nextMarketTokenAdress]?.marketTokenAddress === nextMarketTokenAdress;
+
+  const fallbackPayToken = (fallbackPayToken?: string) => {
+    return produce(nextState, (draft) => {
+      draft.tokens.fromTokenAddress = fallbackPayToken ?? prevState.tokens.fromTokenAddress;
+    });
```
midas-myth, line 676: I dont really undestand the benefit of of immer when its used to set a single field

src/domain/synthetics/trade/useTradeboxState.ts
```diff
@@ -156,9 +162,9 @@ export function useTradeboxState(
       setStoredOptionsOnChain({
         ...INITIAL_SYNTHETICS_TRADE_OPTIONS_STATE,
         markets: {
-          [market.marketTokenAddress]: {
-            long: market.longTokenAddress,
-            short: market.shortTokenAddress,
+          [market.indexTokenAddress]: {
+            long: market.marketTokenAddress,
+            short: market.marketTokenAddress,
```
midas-myth, line 167: This seems strange
could you please explain why long and short otken are now the same market token address

src/domain/synthetics/trade/useTradeboxState.ts
gmxer: but it's nested single field, I think it's fine

src/domain/synthetics/trade/useTradeboxState.ts
gmxer: imagine doing dis without immer

src/domain/synthetics/trade/useTradeboxState.ts
onemicky: It all looks like we have a mistake in the prod codebase here. Since this is a part of the tradebox state, it does not store long and short tokens for the market. In fact, these are: "selected market for the Long tab" and "selected market for the Short tab."

If you look at how the markets property is used in other places, it becomes clear that the data structure there is:
```
{
  [indexTokenAdress]: {
    long: marketTokenAddressForLongs,
    short: marketTokenAddressForShorts,
  }
}
```
and only here it was different, which caused the bug.


Indeed, it's a confusing place. Thanks for pointing it out.

src/domain/synthetics/trade/useTradeboxState.ts
onemicky: yep... it would be like:
```
return {
    ... nextState,
    collaterals: {
        ... nextState.collaterals,
        [marketAddress]: {
            ... nextState.collaterals[marketAddress],
            [longOrShort]: collateralAddress,
        },
    },
};
```

And we also need to take care of the fact that these structures might not exist (which is why we use `set`). As for me, immer greatly enhances readability.

midas-myth: i undestandd the benefit of "set". I am the one who wrote it. How does immer specifically increase readability

src/domain/synthetics/trade/useTradeboxState.ts
midas-myth: ok

midas-myth (approved):

src/domain/synthetics/trade/useTradeboxState.ts
onemicky: discussed in im

PR #1121, opened 2024-07-25
Implement interview notification

divhead:
---
  - https://app.asana.com/0/0/1207802401321392


src/components/InterviewModal/InterviewModal.scss
```diff
@@ -0,0 +1,29 @@
...
+  display: flex;
+  align-items: center;
+  gap: 1.2rem;
+}
+
+.InterviewModal-bullet-point {
+  flex-shrink: 0;
+  height: 1.2rem;
+  width: 1.2rem;
+}
```
gmxer, line 29: let's please use tailwind instead
seems like all of it can be done within it

src/components/InterviewModal/InterviewModal.tsx
```diff
@@ -0,0 +1,40 @@
...
+          <div className="InterviewModal-bullet">
+            <ArrowBulletIcon />
+            <ExternalLink className="InterviewModal-bullet-text" href="https://t.me/GMXPartners">
+              <Trans>Official Telegram account</Trans>
+            </ExternalLink>
+          </div>
+          <div className="InterviewModal-bullet">
+            <ArrowBulletIcon />
+            <div className="text-gray-300 underline">
+              <Trans>Official Discord account: @gmx_feedback</Trans>
```
gmxer, line 33: not a code related one but, perhaps we could share a discord link here?

src/config/localStorage.ts
```diff
@@ -25,6 +25,7 @@ export const KEEP_LEVERAGE_FOR_DECREASE_KEY = "Exchange-keep-leverage";
 export const TRADE_LINK_KEY = "trade-link";
 export const SHOW_DEBUG_VALUES_KEY = "show-debug-values";
 export const ORACLE_KEEPER_INSTANCES_CONFIG_KEY = "oracle-keeper-instances-config";
+export const DEBUG_ACCOUNT_ADDRESS = "debug-account-address";
```
gmxer, line 28: hm, we could use it in `useWallet` then it could be a legit flag

but we somewhere use `useAccount` directly, so won't work in 100% cases

src/context/SyntheticsStateContext/SyntheticsStateContextProvider.tsx
```diff
@@ -139,6 +143,16 @@ export function SyntheticsStateContextProvider({
 
   const orderEditor = useOrderEditorState(ordersInfo.ordersInfoData);
 
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  const lastWeekPeriod = useMemo(() => getTimePeriodsInSeconds().week, [formatDate(Date.now())]);
```
gmxer, line 147: it's not worth to format date per each render to check if week has changed
```
const lastWeekPeriod = useMemo(() => getTimePeriodsInSeconds().week, []);
```
should be enough

src/domain/synthetics/accountStats/usePeriodAccountStats.ts
```diff
@@ -0,0 +1,47 @@
...
+    if (!stats) {
+      return undefined;
+    }
+
+    return {
+      volume: BigInt(stats.volume),
+    };
+  }, [res.data]);
+
+  return { data: result, error: res.error, loading: res.loading };
```
gmxer, line 46: we probably miss it somewhere else, but this line needs useMemo to not trigger SyntheticsContext useMemo

src/domain/synthetics/accountStats/usePnlSummaryData.ts
```diff
@@ -0,0 +1,136 @@
...
+        startUnrealizedPnlUsd: BigInt(row.startUnrealizedPnlUsd),
+        volume: BigInt(row.volume),
+        wins: row.wins,
+        winsLossesRatioBps: row.winsLossesRatioBps ? BigInt(row.winsLossesRatioBps) : undefined,
+        usedCapitalUsd: BigInt(row.usedCapitalUsd),
+      };
+    });
+  }, [res.data?.accountPnlSummaryStats, showDebugValues]);
+
+  return { data: transformedData, error: res.error, loading: res.loading };
```
gmxer, line 135: useMemo or just return transformedData

gmxer (changes requested):
looks good, left a few notes

src/components/InterviewModal/InterviewModal.tsx
divhead: It seems that Discord doesn't have links to accounts

src/config/localStorage.ts
divhead: Tried, it breaks requests in different places

gmxer (dismissed):

src/domain/synthetics/accountStats/usePeriodAccountStats.ts
```diff
@@ -0,0 +1,47 @@
+import { gql, useQuery as useGqlQuery } from "@apollo/client";
+import { DEBUG_ACCOUNT_ADDRESS } from "config/localStorage";
+import { getSubsquidGraphClient } from "lib/subgraph";
+import { useMemo } from "react";
+
+export type PeriodAccountStats = {
+  volume: bigint;
+};
+
+export function usePeriodAccountStats(
```
gmxer, line 10: usePeriodAccount**Volume**Stats ?

src/components/InterviewToast/InterviewToast.tsx
```diff
@@ -0,0 +1,30 @@
...
+      <br />
+      <br />
+      <div onClick={onButtonClick} className="text-xl cursor-pointer text-gray-300 underline">
+        <Trans>Give us your feedback on GMX.</Trans>
+      </div>
+      <br />
+      <Trans>
+        As a token of our appreciation, you'll receive a reward of{" "}
+        <span className="font-400 text-green-500">
+          100 <TokenSymbolWithIcon symbol={getNormalizedTokenSymbol("USDC")} />
```
gmxer, line 24: btw there's no need in getNormalizedTokenSymbol call, same as in events.ts
you can fix it if you would have more comments to fix

gmxer (approved):

PR #1141, opened 2024-08-05
TV chart candles fix

hubert-de-lalye:
Changed: 

On initialize `TVDataProvider` will start an interval to update live bars every `LAST_BAR_FETCH_INTERVAL` and save the last bar in `liveBars` array. If the last bar is the same as the previous one, it will replace it. Otherwise, it will add the new bar to the array. 

Meanwhile, on any update of chartTokenInfo it will update the last bar with the new price.

On reading live bars from feed it will ensure that `open` of new bar is the same as the `close` of the previous bar and drop `liveBars` to array of length = 1. This required to remove possible inaccuracy in prices between `/candles` and `/tickers` sources.

So general picture is: 
 - every second pull new candle
 - update current price on every /tickers response
 - every 500ms update candles on TV chart

---
  - https://app.asana.com/0/0/1207592170196250


gmxer (changes requested):
I think `lastBar` should be called `latestBar`/`newestBar`

Also I don't get why you need an array for `liveBars`. Maybe just a reference to a `currentBar` is enough?

clearing `updateInterval` should be made inside provider, maybe introduce some method `.finalize` and call it in the useEffect cleaning function.

hubert-de-lalye:
Agree with `.finalize` approach, updated that

>Also I don't get why you need an array for liveBars. Maybe just a reference to a currentBar is enough?

Array is required cuz it's easier to handle from outside of provider. There's some points when we need to work with 2 bars, like: 

```
[1] [1] [1] [1] [1] | [1, 2] [2] [2] [2] [2]
                    |   ^
                    | Here we update 1 and write new 2
                    ^
            Here we get new candle
```


It's convenient from outside to grab all required bars for update ([look here](https://github.com/gmx-io/gmx-interface/pull/1141/files#diff-8913689e2fd03cda36bf49401636c119f216c86ba8f122e3b5fa560f1d1ab216R289)). I can store instead of array 2 different fields, but anyway at some point I will need to joggle them as I do in getLiveBars, I decided that array is a bit easier.

>I think lastBar should be called latestBar/newestBar

Renamed to latestBar

gmxer (approved):

gmxer:
thank you for explaining

gmxer (approved):

PR #1142, opened 2024-08-05
Price impact collateral

onemicky:
---
  - https://app.asana.com/0/0/1207880747331350


src/domain/synthetics/trade/usePriceImpactWarningState.ts
```diff
@@ -10,12 +10,12 @@ import { TradeFlags } from "./types";
 export type PriceImpactWarningState = ReturnType<typeof usePriceImpactWarningState>;
 
 export function usePriceImpactWarningState({
-  positionPriceImpact,
+  positionCollateralPriceImpact,
```
hubert-de-lalye, line 13: Why do we renaming this option but didn't change naming of returned fields and internal variabled like `isHighPositionImpact` or `isHighPositionImpactAccepted`?

hubert-de-lalye (dismissed):

src/domain/synthetics/trade/usePriceImpactWarningState.ts
onemicky: Good point, changed back the variable naming. I think it's better to keep the usePriceImpactWarningState hook more abstract. So it doesn't matter which kind of priceImpact it receive.

hubert-de-lalye (dismissed):

hubert-de-lalye (dismissed):

gmxer (dismissed):

PR #1155, opened 2024-08-09
add stake lido apr

onemicky:


src/domain/stake/useLidoStakeApr.ts
```diff
@@ -0,0 +1,19 @@
+import useSWR from "swr";
+import { expandDecimals } from "lib/numbers";
+
+const LIDO_STAKE_APR_URL = "https://eth-api.lido.fi/v1/protocol/steth/apr/sma";
```
divhead, line 4: i think it should be in config?

src/domain/stake/useLidoStakeApr.ts
```diff
@@ -0,0 +1,19 @@
...
+export function useLidoStakeApr() {
+  const { data: lidoAprResult } = useSWR(["lido-stake-apr"], {
+    fetcher: () => fetch(LIDO_STAKE_APR_URL).then((res) => res.json()),
+  });
+
+  if (!lidoAprResult?.data.smaApr) {
+    return undefined;
+  }
+
+  return BigInt(Math.ceil(lidoAprResult.data.smaApr * Number(LIDO_APR_DIVISOR)));
```
divhead, line 18: It's not a big problem, but we can do it in the fetcher instead of every render

onemicky: Good point, fixed

divhead (dismissed):

divhead (approved):

PR #1157, opened 2024-08-11
Prevent store trade options with wrong chainId + reset invalid stored data in case of tokens mismatch

hubert-de-lalye:
---
  - https://app.asana.com/0/0/1207972476109456

### DIFF
--- a/src/domain/synthetics/trade/useTradeboxState.ts
+++ b/src/domain/synthetics/trade/useTradeboxState.ts
@@ -1,5 +1,10 @@
 import get from "lodash/get";
+import isEqual from "lodash/isEqual";
+import keyBy from "lodash/keyBy";
+import set from "lodash/set";
+import values from "lodash/values";
 import mapValues from "lodash/mapValues";
+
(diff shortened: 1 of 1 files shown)


src/domain/synthetics/trade/useTradeboxState.ts
```diff
@@ -598,13 +621,21 @@ export function useTradeboxState(
   );
 
   useEffect(() => {
+    /**
+     * This check required to avoid redundant calls caused by chainId override in context/SynteticsStateContextProvider.txs:104
+     * Overriding chainId causes setStoredOptionsOnChain change and this effect is triggered overriding stored options with wrong chainId
+     */
+    if (isEqual(storedOptions.advanced, advancedOptions)) {
```
hubert-de-lalye, line 628: P.S. I didn't prefer useLatest cuz can't predict possible issues in UI, if reviewer can approve that it would be safe the useLatest also possible solution. But I think that this whole hook (useTradeboxState) should be refactored considering chainId override (or  refuse chainId overriding and solve required things in other maner)

PR #1166, opened 2024-08-14
vite migration

onemicky:
---
  - https://app.asana.com/0/0/1207940040392632


index.html
```diff
@@ -5,17 +5,6 @@
...
-    <meta property="og:site_name" content="GMX" />
-    <meta property="og:image" content="https://gmx.io/og.png" />
-    <meta
-      name="og:description"
-      content="Trade spot or perpetual BTC, ETH, AVAX and other top cryptocurrencies with up to 100x leverage directly from your wallet on Arbitrum and Avalanche."
-    />
-    <meta
-      name="description"
-      content="Trade spot or perpetual BTC, ETH, AVAX and other top cryptocurrencies with up to 100x leverage directly from your wallet on Arbitrum and Avalanche."
-    />
```
gmxer, line 18: why removing this?

src/components/DownloadAsCsv/DownloadAsCsv.tsx
```diff
@@ -3,7 +3,7 @@ import { useCallback } from "react";
 
 import { downloadAsCsv } from "lib/csv";
 
-import DownloadIcon from "img/ic_download.svg";
+import downloadIcon from "img/ic_download.svg";
```
gmxer, line 6: don't we need `?react` here?

src/components/Synthetics/TradeBox/TradeBoxRows/OneClickTrading.tsx
```diff
@@ -28,7 +28,7 @@ import {
 } from "context/SyntheticsStateContext/selectors/tradeboxSelectors";
 import { useSelector } from "context/SyntheticsStateContext/utils";
 import { SUBACCOUNT_DOCS_URL } from "domain/synthetics/subaccount/constants";
-import CloseIcon from "img/navbutton-close.svg";
+import closeIcon from "img/navbutton-close.svg";
```
gmxer, line 31: don't we need `?react` here?

src/lib/multicall/hashData/hashDataAsync.ts
```diff
@@ -1,12 +1,11 @@
...
+import uniqueId from "lodash/uniqueId";
 
 import { sleep } from "lib/sleep";
 
-import HashDataWorker from "./hashData.worker";
 import { hashDataMap } from "./hashDataMap";
 import { promiseWithResolvers } from "lib/utils";
 
-export const hashDataWorker: Worker = new HashDataWorker();
+export const hashDataWorker: Worker = new Worker(new URL("./hashData.worker", import.meta.url), { type: "module" });
```
gmxer, line 8: this is dangerous change, need to find some library for initiating workers

gmxer (changes requested):

index.html
onemicky: Hmm, on one hand, this is a duplication of the SEO component: https://github.com/gmx-io/gmx-interface/blob/vite-build/src/components/Common/SEO.tsx.

On the other hand, it may be useful to have some meta tags even in the static HTML.

Another thought is that all modern crawlers should execute JavaScript.

What do you think?

src/components/DownloadAsCsv/DownloadAsCsv.tsx
onemicky: That's an intended fix: downloadIcon is used as a url, not as a component in the further code

src/components/Synthetics/TradeBox/TradeBoxRows/OneClickTrading.tsx
onemicky: Same thing. Previously, we had a universal import syntax whether we needed a URL or a React-wrapped svg. Now we should use:
```
import CloseIcon from "img/navbutton-close.svg?react";
```
if we want a React component, or
```
import closeIcon from "img/navbutton-close.svg";
```
if we need a URL.

index.html
onemicky: Discussed and decided to bring them back to keep seo stable

src/lib/multicall/hashData/hashDataAsync.ts
onemicky: Discussed in IM: I performed some research, and it seems like Vite handles the new Worker() syntax in a specific way to ensure the reliable operation of workers. Also it's a recommended way to import workers according to vite documentation

gmxer (approved):

gmxer (dismissed):

gmxer (approved):

PR #1170, opened 2024-08-16
Fix workers AB

midas-myth:
---
  - https://app.asana.com/0/0/1208083051719312


src/lib/multicall/Multicall.ts
```diff
@@ -172,7 +179,20 @@ export class Multicall {
       sleep(maxTimeout).then(() => Promise.reject(new Error("multicall timeout"))),
     ]).catch((_viemError) => {
       const e = new Error(_viemError.message.slice(0, 150));
-
+      globalThis.dispatchEvent(
+        new CustomEvent("metrics-mark", {
+          detail: {
+            event: "multicall-timeout",
```
divhead, line 185: event name differs from types

src/lib/multicall/Multicall.ts
```diff
@@ -172,7 +179,20 @@ export class Multicall {
...
+            isError: true,
+            message: _viemError.message.slice(0, 150),
+            data: {
+              metricType: "multicall.timeout",
+              isInMainThread: !isWebWorker,
+            } satisfies MulticallMetricData,
+            time: Date.now(),
```
divhead, line 192: time uses for time diffs, not just “time” 
you can send data.timestamp if needed

src/lib/multicall/Multicall.ts
midas-myth: fixed and typed

src/lib/multicall/Multicall.ts
midas-myth: removed

divhead (approved):

divhead (approved):

PR #1188, opened 2024-08-22
Leverage num input

midas-myth:
---
  - https://app.asana.com/0/0/1206513187657165

### DIFF
--- a/src/components/LeverageSlider/LeverageSlider.tsx
+++ b/src/components/LeverageSlider/LeverageSlider.tsx
@@ -42,12 +42,14 @@ export function LeverageSlider(p: Props) {
     return { marksLabel, keyValueMap, valueKeyMap };
   }, [finalMarks]);
 
-  const sliderValue = valueKeyMap[value ?? 0] ?? sliderValueToSliderKey(keyValueMap, value ?? 0);
+  const sliderValue =
+    value === undefined ? valueKeyMap[finalMarks[0]] : valueKeyMap[value] ?? sliderValueToSliderKey(keyValueMap, value);
+
   const max = (finalMarks.length - 1) * 10;
 
   const handleChange = useCallback(
     (value: number) => {
-      const truncatedValue = Math.trunc(value ?? 0);
+      const truncatedValue = Math.trunc(value ?? DEFAULT_LEVERAGE_KEY);
       onChange(keyValueMap[truncatedValue] ?? DEFAULT_LEVERAGE_KEY);
     },
     [onChange, keyValueMap]
@@ -144,7 +146,7 @@ function sliderValueToSliderKey(keyValueMap: { [key: number]: number }, value: n
 
   const leftValue = sortedValues[leftIndex - 1];
   if (leftValue === undefined) {
-    return DEFAULT_LEVERAGE_KEY;
+    return sortedValues[0];
   }
 
   let rightIndex = sortedValues.findLastIndex((val) => val <= value);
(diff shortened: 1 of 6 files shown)


src/components/LeverageSlider/LeverageSlider.tsx
```diff
@@ -42,12 +42,13 @@ export function LeverageSlider(p: Props) {
     return { marksLabel, keyValueMap, valueKeyMap };
   }, [finalMarks]);
 
-  const sliderValue = valueKeyMap[value ?? 0] ?? DEFAULT_LEVERAGE_KEY;
+  const sliderValue = valueKeyMap[value ?? 0] ?? lerpValue2SliderKey(keyValueMap, value ?? 0);
```
gmxer, line 45: not sure about `?? 0` should not it be `x2`? I think it's not the very first key

src/components/LeverageSlider/LeverageSlider.tsx
```diff
@@ -127,3 +128,40 @@ function generateKeyValueMap(marks: number[]) {
 
   return { keyValueMap, valueKeyMap };
 }
+
+function lerpValue2SliderKey(keyValueMap: { [key: number]: number }, value: number): number {
+  const sortedValues = Object.values(keyValueMap);
```
gmxer, line 133: let's explicitly sort values, order may vary I think

src/domain/synthetics/trade/useTradeboxState.ts
```diff
@@ -637,6 +673,24 @@ export function useTradeboxState(
...
+  //         return;
+  //       }
+
+  //       setLeverageOption(newValue);
+  //     } else {
+  //       setLeverageInputValue(leverageOption?.toString() ?? "");
+  //     }
+  //   },
+  //   [leverageFocusedInput, leverageInputValue, leverageOption, setLeverageOption]
+  // );
```
gmxer, line 692: still needed

src/components/LeverageSlider/LeverageSlider.tsx
```diff
@@ -127,3 +128,40 @@ function generateKeyValueMap(marks: number[]) {
 
   return { keyValueMap, valueKeyMap };
 }
+
+function lerpValue2SliderKey(keyValueMap: { [key: number]: number }, value: number): number {
```
gmxer, line 132: lerp confuses here, it in fact doesn't matter what kind of value is passed here, single value can't be "lerped"

I'd call it valueToSliderKey / sliderValueToSliderKey

gmxer (changes requested):

src/components/LeverageSlider/LeverageSlider.tsx
midas-myth: done

src/components/LeverageSlider/LeverageSlider.tsx
midas-myth: done

src/domain/synthetics/trade/useTradeboxState.ts
midas-myth: done

src/components/LeverageSlider/LeverageSlider.tsx
midas-myth: Agree, made `valueKeyMap[finalMarks[0]]`

gmxer (dismissed):

PR #1191, opened 2024-08-26
add rpc fallback mode

onemicky:
---
  - https://app.asana.com/0/0/1208114241573681


src/lib/oracleKeeperFetcher.ts
```diff
@@ -0,0 +1,286 @@
...
+      body: JSON.stringify({
+        ...body,
+        isDev: isDevelopment(),
+        host: window.location.host,
+        version: APP_VERSION,
+      }),
+    });
+  }
+
+  fetchPostEventReport(body: { event: string }): Promise<Response> {
```
gdev8317, line 237: probably `fetchPostCounterReport`?

src/lib/oracleKeeperFetcher.ts
```diff
@@ -0,0 +1,286 @@
...
+    return fetch(buildUrl(this.url!, "/report/ui/counter"), {
+      method: "POST",
+      headers: {
+        "Content-Type": "application/json",
+      },
+      body: JSON.stringify({
+        ...body,
+        isDev: isDevelopment(),
+        host: window.location.host,
+        version: APP_VERSION,
```
gdev8317, line 247: i have just realized we also need to send `abFlags` to this endpoint, let's add it to the request body

otherwise we won't be able to compare data from different groups in datadog

gdev8317: the endpoint accepts abFlags as Record<string, boolean>

src/lib/metrics/Metrics.ts
```diff
@@ -181,8 +181,8 @@ export class Metrics {
           errorData: txErrorData,
         },
         env: {
-          VITE_APP_IS_HOME_SITE: import.meta.env.VITE_APP_IS_HOME_SITE ?? null,
-          VITE_APP_VERSION: import.meta.env.VITE_APP_VERSION ?? null,
+          IS_HOME_SITE: IS_HOME_SITE ?? null,
+          VERSION: APP_VERSION ?? null,
```
gmxer, line 185: let's use APP_VERSION just in case?
I think there might be some other version types (builder version, git version, etc) that can be confused with just "version"

src/lib/multicall/Multicall.ts
```diff
@@ -116,13 +120,39 @@ export class Multicall {
...
+    },
+  });
+
+  getClient: (options?: { forceFallback?: boolean }) => ReturnType<typeof Multicall.getViemClient>;
 
   constructor(
     public chainId: number,
-    public rpcUrl: string
+    public rpcUrl: string,
+    public fallbackRpcUrl?: string
```
gmxer, line 144: I don't see why fallbackRpcUrl can be undefined, you always provide it.
let's then not support undefined here and fallbackClient variable?

gmxer (changes requested):

gmxer (dismissed):

gmxer (approved):

gmxer (approved):

PR #1204, opened 2024-09-03
prebuild markets

onemicky:
---
  - https://app.asana.com/0/0/1208114241573675
  - https://app.asana.com/0/0/1208114241573678


src/config/contracts.ts
```diff
@@ -332,3 +333,4 @@ function makeGetContract<T extends { abi: InterfaceAbi; connect: (address: strin
 export const getDataStoreContract = makeGetContract("DataStore", DataStore__factory);
 export const getMulticallContract = makeGetContract("Multicall", Multicall__factory);
 export const getExchangeRouterContract = makeGetContract("ExchangeRouter", ExchangeRouter__factory);
+export const getSyntheticsReaderContract = makeGetContract("SyntheticsReader", SyntheticsReader__factory);
```
gmxer, line 336: no need in this?

scripts/prebuild/index.ts
```diff
@@ -0,0 +1,186 @@
...
+
+      return marketsAcc;
+    }, {});
+
+    chainsAcc[chainId] = chainMarkets;
+
+    return chainsAcc;
+  }, {});
+
+  fs.writeFile(resolve(OUTPUT_DIR, 'marketInfoHashedKeys.json'), JSON.stringify(chainMarketKeys, null, 2), (err) => {
```
gmxer, line 177: writeFileSync

scripts/prebuild/index.ts
```diff
@@ -0,0 +1,186 @@
...
+  },
+  [AVALANCHE]: {
+    DataStore: "0x2F0b22339414ADeD7D5F06f9D604c7fF5b2fe3f6",
+    SyntheticsReader: "0xBAD04dDcc5CC284A86493aFA75D2BEb970C72216",
+  },
+  [AVALANCHE_FUJI]: {
+    DataStore: "0xEA1BFb4Ea9A412dCCd63454AbC127431eBB0F0d4",
+    SyntheticsReader: "0xD52216D3A57F7eb1126498f00A4771553c737AE4",
+  },
+};
```
gmxer, line 37: you can take it from `config/contracts`

scripts/prebuild/index.ts
```diff
@@ -0,0 +1,186 @@
...
+  POSITION_IMPACT_POOL_AMOUNT_KEY,
+  SWAP_IMPACT_POOL_AMOUNT_KEY,
+} from "config/dataStore";
+
+import { DataStore__factory } from "typechain-types/factories/DataStore__factory";
+import { SyntheticsReader__factory } from "typechain-types/factories/SyntheticsReader__factory";
+
+const ARBITRUM = 42161;
+const AVALANCHE = 43114;
+const AVALANCHE_FUJI = 43113;
```
gmxer, line 22: `config/chains`

gmxer (changes requested):

gmxer:
1. regarding markets.json let's get rid of it, including building script. Let's extend config/markets ENABLED_MARKETS so that it would be just `const MARKETS`, with `Record<chainId, <Record<marketAddress, MarketUiConfig>>` type:

```
type MarketUiConfig = {
  enabled: boolean;
  listingDate: Date;
  
  marketToken: string;
  indexToken: string;
  longToken: string;
  shortToken: string;
};
```
let's do that for fuji as well.


2. there's no AB yet, PR isn't ready for review

3. you're hashing market values, but not market settings. Let's do both

4. [hashData.worker.ts](https://github.com/gmx-io/gmx-interface/blob/cc822512a968d2880c82c755e49aa254a91dfbb6/src/lib/multicall/hashData/hashData.worker.ts#L4) with `hashDataMapAsync` should have been elliminated already in favor of single worker that does both in it's side?

src/prebuild/index.ts
```diff
@@ -0,0 +1,44 @@
...
+
+import marketsJson from "./markets.json";
+import marketInfoHashedKeysJson from "./marketInfoHashedKeys.json";
+
+type PrebuildMarket = Pick<
+  Market,
+  "marketTokenAddress" | "indexTokenAddress" | "longTokenAddress" | "shortTokenAddress"
+>;
+
+type PrebuildMarketInfo = {
```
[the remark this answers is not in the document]
onemicky: Will use `HASHED_MARKET_VALUES_KEYS` and `HASHED_MARKET_CONFIG_KEYS` if you are ok with it?

src/prebuild/index.ts
```diff
@@ -0,0 +1,44 @@
...
+  };
+} = marketsJson;
+
+const prebuildMarketInfoHashedKeys: {
+  [chainId: number]: {
+    [marketToken: string]: PrebuildMarketInfo;
+  };
+} = marketInfoHashedKeysJson;
+
+export { prebuildMarkets, prebuildMarketInfoHashedKeys };
```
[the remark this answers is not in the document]
onemicky: Was managed to create `src/config/static/markets.ts` to avoid problems with client-only code in node

src/domain/synthetics/markets/useMarketsInfoRequest.ts
```diff
@@ -393,72 +395,99 @@ function useMarketsValuesRequest({
...
-          shortInterestInTokensUsingLongToken: [
-            ["bytes32", "address", "address", "bool"],
-            [OPEN_INTEREST_IN_TOKENS_KEY, marketAddress, market.longTokenAddress, false],
-          ],
-          shortInterestInTokensUsingShortToken: [
-            ["bytes32", "address", "address", "bool"],
-            [OPEN_INTEREST_IN_TOKENS_KEY, marketAddress, market.shortTokenAddress, false],
-          ],
-        });
+        const prebuildHashedKeys = prebuildMarketInfoHashedKeys[chainId]?.[marketAddress];
```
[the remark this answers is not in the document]
onemicky: This file was gone finally in [pr 1206](https://github.com/gmx-io/gmx-interface/pull/1206)

scripts/prebuild/index.ts
onemicky: good point, thanks 👍

scripts/prebuild/index.ts
onemicky: No need more for contract fetching here

scripts/prebuild/index.ts
onemicky: Had to move it into `src/config/static/chains.ts`

onemicky:
> 1. regarding markets.json let's get rid of it, including building script. Let's extend config/markets ENABLED_MARKETS so that it would be just `const MARKETS`, with `Record<chainId, <Record<marketAddress, MarketUiConfig>>` type:
> 
> ```
> type MarketUiConfig = {
>   enabled: boolean;
>   listingDate: Date;
>   
>   marketToken: string;
>   indexToken: string;
>   longToken: string;
>   shortToken: string;
> };
> ```
> 
> let's do that for fuji as well.

Did it this way if you have no objections:
```
type MarketUiConfig = {
  listingDate: Date;
  tokens: {
    marketTokenAddress: string;
    indexTokenAddress: string;
    longTokenAddress: string;
    shortTokenAddress: string;
  };
};
```

> 2. there's no AB yet, PR isn't ready for review

Added AB test `testPrebuiltMarkets`

> 3. you're hashing market values, but not market settings. Let's do both

Now both market values and market config keys are hashing

> 4. [hashData.worker.ts](https://github.com/gmx-io/gmx-interface/blob/cc822512a968d2880c82c755e49aa254a91dfbb6/src/lib/multicall/hashData/hashData.worker.ts#L4) with `hashDataMapAsync` should have been elliminated already in favor of single worker that does both in it's side?

That's right, now i can remove them as 1206 has been merged

src/prebuild/index.ts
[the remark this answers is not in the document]
gmxer: sure

vite.config.ts
```diff
@@ -40,6 +40,7 @@ export default defineConfig({
       "pages": path.resolve(__dirname, "src/pages"),
       "styles": path.resolve(__dirname, "src/styles"),
       "typechain-types": path.resolve(__dirname, "src/typechain-types"),
+      "prebuit": path.resolve(__dirname, "src/prebuilt"),
```
gmxer, line 43: typo, maybe not needed at all?

package.json
```diff
@@ -14,7 +14,7 @@
     "test:ci": "vitest run",
     "lint": "node node_modules/.bin/eslint src",
     "check:ci": "yarn lint && yarn test:ci && yarn tscheck",
-    "pre-commit": "vitest run && yarn run lint",
+    "pre-commit": "yarn run prebuild && vitest run && yarn run lint",
```
gmxer, line 17: I think we don't need to run this per each commit. it's enough to just run it per new market addition

src/config/static/markets.ts
```diff
@@ -0,0 +1,673 @@
...
+
+import { ARBITRUM, AVALANCHE, AVALANCHE_FUJI } from "./chains";
+
+const p = (date: string) => parse(date, "dd MMM yyyy", new Date());
+
+export const DEFAULT_LISTING = p("01 Jan 1970");
+
+type MarketUiConfig = {
+  listingDate: Date;
+  tokens: {
```
gmxer, line 18: I disagree with `.tokens` nesting because:

* we already have plain structures with tokens, e.x. `MarketInfo` and I didn't notice any struggle with tokens aligned with other data

* we any have index**Token**Address in each field name, so it should rather be `indexAddress`/`index` which condtradicts how we do it in other places

gmxer: Let's explicitly have `enabled: boolean` field, and use it in `MARKET_INDEX` map

gmxer (changes requested):

gmxer:
I bit sad of `static` folder but I guess there's really no better way. 
Maybe at least better naming but I couldn't think of, and I'm ok to have like this for now

vite.config.ts
onemicky: Thanks. It’s really not needed in most cases, but in some situations like using inside workers, these aliases are necessary. Since we already have aliases for all folders at this level, let’s have one for this as well.

src/config/static/markets.ts
onemicky: Migrated to the flat structure and added enabled key

gmxer (approved):

PR #1210, opened 2024-09-05
fix worker throw error

onemicky:
---
  - https://app.asana.com/0/0/1208210628602379

### DIFF
--- a/src/lib/multicall/executeMulticall.ts
+++ b/src/lib/multicall/executeMulticall.ts
@@ -74,7 +74,7 @@ async function executeChainMulticall(chainId: number, calls: MulticallFetcherCon
       return `Executing multicall for chainId: ${chainId}. Call count: ${callCount}. Execution in ${executionIn}.`;
     });
 
-    if (callCount > CALL_COUNT_MAIN_THREAD_THRESHOLD && !isOldIOS(16)) {
+    if (callCount > CALL_COUNT_MAIN_THREAD_THRESHOLD && !isOldIOS()) {
       responseOrFailure = await executeMulticallWorker(chainId, requestConfig);
     } else {
       responseOrFailure = await executeMulticallMainThread(chainId, requestConfig);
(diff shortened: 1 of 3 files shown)


src/lib/multicall/multicall.worker.ts
```diff
@@ -28,7 +28,13 @@ async function run(event) {
       result,
     });
   } catch (error) {
-    postMessage({ id, error: error });
+    postMessage({
+      id,
+      error: {
+        message: error.message,
+        stack: error.stack,
```
gdev8317, line 35: i think it makes sense to also pas error.name and error.cause

src/lib/multicall/executeMulticall.ts
```diff
@@ -361,3 +359,18 @@ function getRequest(callEntries: [string, { callData: MulticallFetcherConfig[num
 
   return requests;
 }
+
+function isOldIOS(threshold: number) {
```
gdev8317, line 363: maybe let's remove threshold argument? we need to detect specific version because of quite specific error. 

and i think it makes sense to add some comment that says that there is an issue with serialization/deserialization on iOS <= 16

gdev8317: and 16 could be moved to a constant

src/lib/multicall/executeMulticallWorker.ts
```diff
@@ -85,15 +91,24 @@ export async function executeMulticallWorker(
...
       );
-      try {
-        const result = await executeMulticallMainThread(chainId, request);
+    } else {
+      emitMetricEvent({
+        event: "worker.multicall.error",
+        isError: true,
+        data: {
+          isInMainThread: false,
+          errorMessage: error.message,
```
gdev8317, line 100: iirc in other places we also pass some other fields, let's pass them here as well 
image

src/lib/multicall/multicall.worker.ts
onemicky: Added error.name, but error.cause is often used to pass an original error object which can cause the sendMessage issue again. Let's keep this place simple as possible

src/lib/multicall/executeMulticallWorker.ts
onemicky: Added errorName and errorStack, but errorSource and errorStackHash are fields that the errorReporting module automatically adds to all unhandled errors caught at the window level. In this case, we have a handled error with a specific name, so I believe we can skip those fields

src/lib/multicall/multicall.worker.ts
gdev8317: good point

gdev8317 (dismissed):

gdev8317 (approved):

PR #1227, opened 2024-09-11
smart rpc switching

onemicky:


src/context/SubaccountContext/SubaccountContext.tsx
```diff
@@ -303,8 +297,7 @@ function useSubaccountCustomSigners() {
   const privateKey = useSubaccountPrivateKey();
 
   return useMemo(() => {
-    const publicRpc = getRpcUrl(chainId);
-    const fallbackRpc = getFallbackRpcUrl(chainId);
+    const { default: publicRpc, fallback: fallbackRpc } = getBestRpc(chainId);
```
gmxer, line 300: I think this problem was before but now it seems it would cache rpc forever, not sure if it's what we want

src/lib/multicall/Multicall.ts
```diff
@@ -23,6 +22,11 @@ const CHAIN_BY_CHAIN_ID = {
   [AVALANCHE]: avalanche,
 };
 
+export type MulticallProviderUrls = {
+  default: string;
+  fallback: string;
```
gmxer, line 27: regarding naming
"default" also has "fallback" meaning, let's somehow distinguish it 

I'd suggest
"current" and "fallback"
"primary" and "secondary"
"main" and "fallback"

gmxer: another thought after looking at the rest of the code.
even after changing these names we still have `getBestRpc(...).fallback` which also doesn't make sense.

let's have `getBestRpc` (or `getBestRpcs`) internally how it is now, 
but let's also introduce `getRpcUrl` and `getFallbackRpcUrl` that would call `getBestRpc(s)` internally.

That way we don't need `MulticallProviderUrls` structure, at least outside of this file, as well as calling `getBestRpc` directly

EDIT `MulticallProviderUrls` is also needed in `src/lib/multicall/multicall.worker.ts`

src/lib/multicall/Multicall.ts
```diff
@@ -372,10 +332,6 @@ export class Multicall {
       isAlchemy: isFallbackMode,
```
gmxer, line 332: we still have isAlchemy + isFallback which mean same thing? let's pick only one?

src/lib/multicall/Multicall.ts
```diff
@@ -372,10 +332,6 @@ export class Multicall {
...
     });
```
gmxer, line 334: also thinking if we should add tests for this file, it doesn't seem to complex but quite frigile and important

src/lib/rpc/bestRpcTracker.ts
```diff
@@ -0,0 +1,368 @@
...
+// DataStore field used for probing
+const PROBE_SAMPLE_FIELD = "minCollateralFactor";
+// Markets used for `PROBE_SAMPLE_FIELD` reading
+const PROBE_SAMPLE_MARKET = {
+  [ARBITRUM]: "0x70d95587d40A2caf56bd97485aB3Eec10Bee6336", // ETH/USD
+  [AVALANCHE]: "0xB7e69749E3d2EDd90ea59A4932EFEa2D41E245d7", // ETH/USD
+  [AVALANCHE_FUJI]: "0xbf338a6C595f06B7Cfff2FA8c958d49201466374", // ETH/USD
+};
+
+type RpcProbe = {
```
gmxer, line 41: ProbeData / Probe?

src/lib/rpc/bestRpcTracker.ts
```diff
@@ -0,0 +1,368 @@
...
+  [chainId: number]: {
+    lastUsage: Date;
+    currentBestProviderUrl: string;
+    providers: {
+      [providerUrl: string]: ProviderData;
+    };
+  };
+};
+
+let trackingIntervalId: NodeJS.Timeout | null = null;
```
gmxer, line 64: you can use `window.setInterval` instead of just `setInterval` that way you would have `number` instead of nodejs type

gmxer (changes requested):
submitting comments that doesn't relate to best rpc specifically, will check best rpc logic separately later today

src/lib/rpc/bestRpcTracker.ts
```diff
@@ -0,0 +1,368 @@
...
+      const nextProbeBlockNumber = arr[i + 1]?.blockNumber;
+
+      // Rare case when RPC returned a block number from the far future
+      if (
+        i === 0 &&
+        currProbeBlockNumber &&
+        nextProbeBlockNumber &&
+        currProbeBlockNumber - nextProbeBlockNumber > BLOCK_FROM_FUTURE_THRESHOLD
+      ) {
+        bestBlockNumberProbe = probe;
```
gmxer, line 135: I think you want `bestBlockNumberProbe = arr[i + 1]` here?

src/lib/rpc/bestRpcTracker.ts
```diff
@@ -0,0 +1,368 @@
...
+      return;
+    }
+
+    const probeResultsByBlockNumber = orderBy(successProbeResults, ["blockNumber"], ["desc"]);
+    let bestBlockNumberProbe = probeResultsByBlockNumber[0];
+
+    const probeStats = probeResultsByBlockNumber.map((probe, i, arr) => {
+      let isValid = probe.isSuccess;
+
+      const bestBlockNumber = bestBlockNumberProbe?.blockNumber;
```
gmxer, line 124: no need for `?` here

src/lib/rpc/bestRpcTracker.ts
```diff
@@ -0,0 +1,368 @@
...
+      const currProbeBlockNumber = probe.blockNumber;
+      const nextProbeBlockNumber = arr[i + 1]?.blockNumber;
+
+      // Rare case when RPC returned a block number from the far future
+      if (
+        i === 0 &&
```
gmxer, line 130: imagine same provider holding 3 top places in successProbeResults, that way you would challenge provider vs itself. you need to determine bestBlockNumberProbe separately outside of this loop

gmxer (changes requested):

src/lib/rpc/bestRpcTracker.ts
```diff
@@ -0,0 +1,368 @@
...
+  responseTime: number;
+  blockNumber: number | null;
+  timestamp: Date;
+};
+
+type ProviderData = {
+  url: string;
+  provider: Provider;
+  probeData: RpcProbe[];
+};
```
gmxer, line 52: probeData -> probes

src/lib/rpc/bestRpcTracker.ts
```diff
@@ -0,0 +1,368 @@
...
+    if (
+      differenceInMilliseconds(Date.now(), chainTracker.lastUsage) > DISABLE_UNUSED_TRACKING_TIMEOUT ||
+      providers.length < 2
+    ) {
+      return;
+    }
+
+    const probes = providers.map(async (providerInfo) => {
+      providerInfo.probeData = providerInfo.probeData.filter((probe) => {
+        return Date.now() - probe.timestamp.getTime() < PROBE_SLIDING_WINDOW;
```
gmxer, line 95: probes -> probesPromise

src/lib/rpc/bestRpcTracker.ts
```diff
@@ -0,0 +1,368 @@
...
+    };
+  });
+}
+
+function initTrackerState() {
+  const now = Date.now();
+
+  return SUPPORTED_CHAIN_IDS.reduce<RpcTrackerState>((acc, chainId) => {
+    const providers = RPC_PROVIDERS[chainId].reduce((acc, rpcUrl) => {
+      acc[rpcUrl] = {
```
gmxer, line 302: here types are messed up, can do
```
    const list = RPC_PROVIDERS[chainId] as string[];
    const providers = list.reduce<Record<string, ProviderData>>((acc, rpcUrl) => {
```

src/lib/rpc/bestRpcTracker.ts
```diff
@@ -0,0 +1,368 @@
...
+    }, {});
+
+    let currentBestProviderUrl: string = RPC_PROVIDERS[chainId][0];
+
+    const storageKey = JSON.stringify(getRpcProviderKey(chainId));
+    const storedProviderData = localStorage.getItem(storageKey);
+
+    if (storedProviderData) {
+      let rpcUrl, timestamp;
+
```
gmxer, line 318: let rpcUrl: string | undefined, timestamp: number | undefined;

src/context/SubaccountContext/SubaccountContext.tsx
onemicky: Good point, thx

src/lib/multicall/Multicall.ts
onemicky: Renamed everything to `isAlchemy`

src/lib/multicall/Multicall.ts
onemicky: Yes, it sounds very reasonable. But I think I'll do it a little later, because now I have a feeling that this logic can change many times in the near future.

As an example, we are doing back and forth with this pr now: https://github.com/gmx-io/gmx-interface/pull/1237

src/lib/rpc/bestRpcTracker.ts
onemicky: Indeed, you are right

src/lib/rpc/bestRpcTracker.ts
onemicky: Could you elaborate on this thought? I don't understand how one provider can take several places when different providers compete with each other in design

gmxer (approved):

src/lib/multicall/Multicall.ts
gdev8317: > I changed naming to the "primary" and "fallback" providers if it's ok with you

i think it's better to use "primary" and "secondary" then

src/lib/rpc/bestRpcTracker.ts
```diff
@@ -0,0 +1,346 @@
...
+  return localStorage.getItem(JSON.stringify(SHOW_DEBUG_VALUES_KEY)) === "true";
+}
+
+async function probeRpc(chainId: number, provider: Provider, providerUrl: string): Promise<ProbeData> {
+  const controller = new AbortController();
+
+  return await Promise.race([
+    sleep(PROBE_FAIL_TIMEOUT).then(() => {
+      controller.abort();
+      throw new Error("Probe timeout");
```
gdev8317, line 193: it's not a big issue, but this error will always be thrown, because timeout is not cleared

i think it's better to use setTimeout and clear it after the successful RPC request. at first glance there is no need to use Promise.race

src/lib/rpc/bestRpcTracker.ts
```diff
@@ -0,0 +1,346 @@
...
+        } catch (error) {
+          if (error.name !== "AbortError") {
+            throw error;
+          }
+        }
+      }
+
+      return {
+        url: providerUrl,
+        responseTime: Date.now() - startTime,
```
gdev8317, line 265: not critical, but might be better to measure time of the http requests specifically, not the whole function call that e.g. includes some encoding/decoding, that might take some extra time and make the measurement a bit inaccurate

src/lib/rpc/bestRpcTracker.ts
```diff
@@ -0,0 +1,346 @@
...
+        blockNumber,
+        timestamp: new Date(),
+        isSuccess,
+      };
+    })(),
+  ]).catch(() => {
+    return {
+      url: providerUrl,
+      responseTime: PROBE_FAIL_TIMEOUT,
```
gdev8317, line 274: just to confirm, why default responseTime is 10s? i guess request could fail for various reasons, not only timestamp?

src/lib/multicall/Multicall.ts
gdev8317: to be fair "default" and "fallback" also clear good to me

src/lib/rpc/bestRpcTracker.ts
```diff
@@ -0,0 +1,346 @@
...
+      const currProbeBlockNumber = probe.blockNumber;
+      const nextProbeBlockNumber = arr[i + 1]?.blockNumber;
+
+      // Rare case when RPC returned a block number from the far future
+      if (
+        i === 0 &&
+        currProbeBlockNumber &&
+        nextProbeBlockNumber &&
+        currProbeBlockNumber - nextProbeBlockNumber > BLOCK_FROM_FUTURE_THRESHOLD
+      ) {
```
gdev8317, line 121: this validation doesn't handle a case when there are 3 probes:
1. block = 1001
2. block = null
3. block = 1

since only (1) and (2) are compared to each other, (1) will be considered valid

src/lib/rpc/bestRpcTracker.ts
```diff
@@ -0,0 +1,346 @@
...
+      return {
+        ...probe,
+        isValid,
+      };
+    });
+
+    const bestResponseTimeValidProbe = minBy(
+      probeStats.filter((probe) => probe.isValid),
+      "responseTime"
+    );
```
gdev8317, line 140: wondering if pure js would be simpler here 
```
const bestResponseTimeValidProbe = Math.min(... probeStats.filter(p => p.isValid).map(p.responseTime))
```

src/lib/rpc/bestRpcTracker.ts
```diff
@@ -0,0 +1,346 @@
...
+      setCurrentProvider(chainId, getFallbackRpcUrl(chainId));
+
+      return;
+    }
+
+    const probeResultsByBlockNumber = orderBy(successProbeResults, ["blockNumber"], ["desc"]);
+    let bestBlockNumberProbe = probeResultsByBlockNumber[0];
+
+    const probeStats = probeResultsByBlockNumber.map((probe, i, arr) => {
+      let isValid = probe.isSuccess;
```
gdev8317, line 109: isValid is always true here, because `probeResultsByBlockNumber` is sorted `successProbeResults`

src/lib/rpc/bestRpcTracker.ts
```diff
@@ -0,0 +1,346 @@
...
+
+          if (result) {
+            const multicallResult = multicall.interface.decodeFunctionResult("blockAndAggregate", result);
+            const [sampleFieldValue] = dataStore.interface.decodeFunctionResult(
+              "getUint",
+              multicallResult.returnData[0].returnData
+            );
+
+            blockNumber = Number(multicallResult.blockNumber);
+            isSuccess = sampleFieldValue && sampleFieldValue > 0;
```
gdev8317, line 254: in what case does it happen?

src/lib/rpc/bestRpcTracker.ts
gdev8317: tbh i personally don't like to use adjacent array items inside the filter/map loop. i think it's less readable. imo something instead of current loop like this would be simpler:

```
// remove item with block from future
if (
  probeResultsByBlockNumber.length >= 2 &&
  probeResultsByBlockNumber[0].blockNumber - probeResultsByBlockNumber[1].blockNumber > BLOCK_FROM_FUTURE_THRESHOLD
) {
  probeResultsByBlockNumber.shift();
}
const bestBlockNumber = probeResultsByBlockNumber[0].blockNumber;

// filter out items with lagging blocks
probeResultsByBlockNumber = probeResultsByBlockNumber.filter(p => bestBlockNumber - p.blockNumber <= BLOCK_LAGGING_THRESHOLD);
```

unless i'm missing something, there is no need to create probeStats. you just need to filter out "bad" results

src/lib/rpc/bestRpcTracker.ts
```diff
@@ -0,0 +1,346 @@
...
+    JSON.stringify({
+      rpcUrl: newProviderUrl,
+      timestamp: Date.now(),
+    })
+  );
+}
+
+function isDebugMode() {
+  return localStorage.getItem(JSON.stringify(SHOW_DEBUG_VALUES_KEY)) === "true";
+}
```
gdev8317, line 185: maybe we can move isDebugMode to some utils file?

src/lib/multicall/Multicall.ts
onemicky: renamed to "primary" and "secondary"

src/lib/rpc/bestRpcTracker.ts
onemicky: I also don't like this place, but the reason to keep this cycle - is to get the probeStats array which is used to show the debug table with probe data in debug mode.

If you'd like it more, we can do something like that: 

```
let bestBlockNumberProbe = probeResultsByBlockNumber[0];
const nextProbeBlockNumber = probeResultsByBlockNumber[1]?.blockNumber;

if (
  bestBlockNumberProbe?.blockNumber &&
  nextProbeBlockNumber &&
  bestBlockNumberProbe.blockNumber - nextProbeBlockNumber > BLOCK_FROM_FUTURE_THRESHOLD
) {
  bestBlockNumberProbe = probeResultsByBlockNumber[1];
}

const probeStats = probeResultsByBlockNumber.map((probe, i) => {
  let isValid = true;

  const bestBlockNumber = bestBlockNumberProbe.blockNumber;
  const currProbeBlockNumber = probe.blockNumber;

  if (bestBlockNumber && currProbeBlockNumber && bestBlockNumber - currProbeBlockNumber > BLOCK_LAGGING_THRESHOLD) {
    isValid = false;
  }

  return {
    ...probe,
    isValid,
  };
});
```

src/lib/rpc/bestRpcTracker.ts
onemicky: I think the benefit of the `race` expression is that we can leave this error as it is, and don't care about timers as the other promise wins the race. As for me it makes code more readable. Correct me if i'm wrong

src/lib/rpc/bestRpcTracker.ts
onemicky: Good point, replaced `PROBE_FAIL_TIMEOUT` with `Date.now() - startTime`

src/lib/rpc/bestRpcTracker.ts
```diff
@@ -0,0 +1,346 @@
...
+
+    if (storedProviderData) {
+      let rpcUrl: string | undefined;
+      let timestamp: number | undefined;
+
+      try {
+        const storedProvider = JSON.parse(storedProviderData);
+
+        rpcUrl = storedProvider.rpcUrl;
+        timestamp = storedProvider.timestamp;
```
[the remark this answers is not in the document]
onemicky: There is a check few rows further: the rpcUrl should be in the list of known providers

src/lib/rpc/bestRpcTracker.ts
onemicky: The benefit of `minBy` is that it returns the object itself instead of the minimal responseTime

src/lib/rpc/bestRpcTracker.ts
onemicky: replaced with `isValid = true`

src/lib/rpc/bestRpcTracker.ts
onemicky: Moved to the `lib/localStorage`

src/lib/rpc/bestRpcTracker.ts
gdev8317: in this specific case it looks a bit redundant to me 🤔 because on the timeout the ongoing fetch will also be rejected because of controller.abort(), so both promises will fail in fact (although the first one would be sleep timeout)

src/lib/rpc/bestRpcTracker.ts
gdev8317: i believe there are cases when it takes more time on slower machines? what is the downside of measuring time of request specifically?

src/lib/rpc/bestRpcTracker.ts
gmxer: > both promises will fail in fact

if promise is fulfilled once it can't be rejected. same with request - if it's already returned response - we don't care if smth is happened with it later.

I'm ok with this approach, we use it in several critical places and it works 
https://github.com/search?q=repo%3Agmx-io%2Fgmx-interface%20promise.race&type=code

src/lib/rpc/bestRpcTracker.ts
gmxer: I vote to move this logic outside of a loop since we only want to check only first provider

btw don't we want to deal with cases when 2 or more providers are from future? I thought logic could be - find the most "correct" block

now we're fixing case: 1 2 3 4 5 100
what if we have case: 1 2 3 4 5 100 100?

maybe logic should prioritize majority and punish extereme difference?

src/lib/rpc/bestRpcTracker.ts
```diff
@@ -0,0 +1,343 @@
...
+        responseTime: Date.now() - startTime,
+        blockNumber,
+        timestamp: new Date(),
+        isSuccess,
+      };
+    })(),
+  ]).catch(() => {
+    return {
+      url: providerUrl,
+      responseTime: Date.now() - startTime,
```
gdev8317, line 271: btw is there a reason to return response time for failed requests? i guess time of failed response should not be considered anyway?

src/lib/rpc/bestRpcTracker.ts
onemicky: 1. Updated this place. Now the "block from the future" case handling is outside from the cycle
2. Discussed, that it's nearly impossible that the 1 2 3 4 5 100 100 case can happen

gmxer (dismissed):

midas-myth (approved):

midas-myth (approved):

midas-myth (dismissed):

midas-myth (approved):

PR #1259, opened 2024-09-23
TBTC incentives

midas-myth:
---
  - https://app.asana.com/0/0/1208359641593904

### DIFF
--- a/src/components/AprInfo/AprInfo.tsx
+++ b/src/components/AprInfo/AprInfo.tsx
@@ -88,22 +88,17 @@ export function AprInfo({
             </Trans>
           </div>
         )}
-        {isIncentiveActive &&
-          (isTbtcIncentive ? (
-            <div>
-              <Trans>
-                The Bonus APR will be airdropped as {airdropTokenTitle} and T tokens.{" "}
-                <ExternalLink href={TBTC_INFORMATION_URL}>Read more</ExternalLink>.
-              </Trans>
-            </div>
-          ) : (
-            <div>
-              <Trans>
-                The Bonus APR will be airdropped as {airdropTokenTitle} tokens.{" "}
-                <ExternalLink href={getIncentivesV2Url(chainId)}>Read more</ExternalLink>.
-              </Trans>
-            </div>
-          ))}
+        {isIncentiveActive && (
+          <div>
+            <Trans>
+              The Bonus APR will be airdropped as {airdropTokenTitle} tokens.{" "}
+              <ExternalLink href={isTbtcIncentive ? TBTC_INFORMATION_URL : getIncentivesV2Url(chainId)}>
+                Read more
+              </ExternalLink>
+              .
+            </Trans>
+          </div>
+        )}
         {isEthenaSatsIncentive && (
           <div>
             <Trans>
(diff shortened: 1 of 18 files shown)


src/components/AprInfo/AprInfo.tsx
```diff
@@ -54,6 +55,7 @@ export function AprInfo({
   const isIncentiveActive = !!incentivesData && incentivesData?.rewardsPerMarket[tokenAddress] !== undefined;
   const isLidoApr = lidoApr !== undefined && lidoApr > 0n;
   const isEthenaSatsIncentive = isEthenaSatsIncentivizedMarket(tokenAddress);
+  const isTbtcIncentive = isTbtcIncentivizedMarket(tokenAddress);
```
gmxer, line 58: just noticed tokenAddress is actually a marketAddress, let's please rename this prop to avoid confusion

src/components/AprInfo/AprInfo.tsx
```diff
@@ -86,14 +88,22 @@ export function AprInfo({
...
-        )}
+        {isIncentiveActive &&
+          (isTbtcIncentive ? (
+            <div>
+              <Trans>
+                The Bonus APR will be airdropped as {airdropTokenTitle} and T tokens.{" "}
+                <ExternalLink href={TBTC_INFORMATION_URL}>Read more</ExternalLink>.
+              </Trans>
+            </div>
+          ) : (
```
gmxer, line 99: as I understand we won't support simultaneous incentives programs. so it's either T incentives or ARB?

src/components/TokenCard/TokenCard.tsx
```diff
@@ -117,22 +117,9 @@ export default function TokenCard({ showRedirectModal }: Props) {
...
-      );
-    } else if (arbitrumIncentiveState?.lp?.isActive) {
-      return (
-        <Trans>
-          {arbitrumLink} GM Pools are <span className="whitespace-nowrap">incentivized{sparkle}.</span>
-        </Trans>
-      );
-    } else if (avalancheIncentiveState?.lp?.isActive) {
+
+    if (avalancheIncentiveState?.lp?.isActive) {
```
gmxer, line 122: what's the point of removing logic that should be turned off by keeper config?

gmxer (changes requested):

src/components/TokenCard/TokenCard.tsx
midas-myth: but arbitrum will be active. but with just one pool. and we want to not take it into account

src/components/AprInfo/AprInfo.tsx
midas-myth: both tokens would be rewarded

image
https://x.com/GMX_IO/status/1833896397709463921

src/components/AprInfo/AprInfo.tsx
midas-myth: indeed, done

midas-myth: indeed, done

gmxer (dismissed):

src/components/AprInfo/AprInfo.tsx
midas-myth: @gmxer in the end we decided to remove tbtc so only arb is left

gmxer (approved):

PR #1278, opened 2024-09-30
alchemy for large accounts

onemicky:
---
  - https://app.asana.com/0/0/1208263214312296


src/lib/rpc/bestRpcTracker.ts
```diff
@@ -79,7 +83,7 @@ function trackRpcProviders({ warmUp = false } = {}) {
       const hasMultipleProviders = Object.keys(providers).length > 1;
       const isUnusedChain =
         !lastUsage || differenceInMilliseconds(Date.now(), lastUsage) > DISABLE_UNUSED_TRACKING_TIMEOUT;
-      const isChainTrackingEnabled = (warmUp || !isUnusedChain) && hasMultipleProviders;
+      const isChainTrackingEnabled = !isLargeAccount && (warmUp || !isUnusedChain) && hasMultipleProviders;
```
gmxer, line 86: I think this part of requirements wasn't implemented:

>If paid RPC does not work, usual logic of RPC selection must be used (competitive selection, switching, etc, including free RPCs)

gmxer (changes requested):

src/domain/synthetics/accountStats/useAccountVolumeStats.ts
```diff
@@ -0,0 +1,94 @@
...
+import useSWR from "swr";
+import { subDays, format, eachDayOfInterval } from "date-fns";
+import { toUtcDayStart } from "lib/dates";
+import { CONFIG_UPDATE_INTERVAL } from "lib/timeConstants";
+import { ARBITRUM, AVALANCHE } from "config/chains";
+
+const LARGE_ACCOUNT_CHAINS = [ARBITRUM, AVALANCHE];
+
+export function useAccountVolumeStats(params: { account?: string; enabled?: boolean }) {
+  const { account, enabled = true } = params;
```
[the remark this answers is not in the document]
onemicky: Indeed, removed this props

src/lib/rpc/bestRpcTracker.ts
onemicky: My initial thought was that after the last ABs, we proved Alchemy is reliable and don't need this 'fallback for fallback.' However, after syncing with G, it became clear that it's better to do it.
Thank you for highlighting this point.

gmxer (dismissed):

midas-myth (approved):

src/domain/synthetics/accountStats/useAccountVolumeStats.ts
```diff
@@ -0,0 +1,95 @@
...
+import { useMemo } from "react";
+import useSWR from "swr";
+import { subDays, format, eachDayOfInterval } from "date-fns";
+import { toUtcDayStart } from "lib/dates";
+import { CONFIG_UPDATE_INTERVAL } from "lib/timeConstants";
+import { ARBITRUM, AVALANCHE } from "config/chains";
+
+const LARGE_ACCOUNT_CHAINS = [ARBITRUM, AVALANCHE];
+
+export function useAccountVolumeStats(params: { account?: string }) {
```
divhead, line 12: We already have useAccountStats and usePeriodAccountStats

as i see this hook request daily buckes, probably it make sense to rename this to something specific to this logic, because at first glance it's a bit confusing to create a new hook with that general name

divhead: also usually we create a hook with chainId param but there you are iterating through all the chains, I'm not sure it may be needed in other cases if this data will be reused

It might make sense to call multiple hooks and wrap them in an aggregating hook to count the total values

src/domain/synthetics/accountStats/useAccountVolumeStats.ts
```diff
@@ -0,0 +1,95 @@
+import { gql } from "@apollo/client";
+import { getSubsquidGraphClient } from "lib/subgraph";
+import { useMemo } from "react";
+import useSWR from "swr";
+import { subDays, format, eachDayOfInterval } from "date-fns";
+import { toUtcDayStart } from "lib/dates";
+import { CONFIG_UPDATE_INTERVAL } from "lib/timeConstants";
+import { ARBITRUM, AVALANCHE } from "config/chains";
+
+const LARGE_ACCOUNT_CHAINS = [ARBITRUM, AVALANCHE];
```
divhead, line 10: this is just SUPPORTED_CHAIN_IDS

src/domain/synthetics/accountStats/useAccountVolumeStats.ts
```diff
@@ -0,0 +1,95 @@
...
+          volume,
+        };
+      });
+
+      return {
+        totalVolume,
+        dailyVolume,
+      };
+    },
+    refreshInterval: CONFIG_UPDATE_INTERVAL,
```
divhead, line 91: i think it's more noticeable to have refreshInterval field before fetcher function

src/domain/synthetics/accountStats/useIsLargeAccountData.ts
```diff
@@ -0,0 +1,30 @@
+import { useMemo } from "react";
+import { USD_DECIMALS } from "config/factors";
+import { expandDecimals } from "lib/numbers";
+
+import { useAccountVolumeStats } from "./useAccountVolumeStats";
+
+// Thresholds to recognise large accounts
+const MAX_DAILY_VOLUME = expandDecimals(220_000n, USD_DECIMALS);
```
divhead, line 8: sound like constants that should be places in config/factors

src/lib/account/isLargeAccount.ts
```diff
@@ -0,0 +1,38 @@
+import { IS_LARGE_ACCOUNT_KEY } from "config/localStorage";
+import { useEffect } from "react";
+import { useLocalStorageSerializeKey } from "lib/localStorage";
+import { useIsLargeAccountData } from "domain/synthetics/accountStats/useIsLargeAccountData";
+import { getStorageItem } from "lib/metrics/storage";
+
+let isLargeAccount = getIsLargeAccountStoredValue();
```
divhead, line 7: src/lib directory is more for abstract infrastructure code that not related to GMX domain

this logic looks like utils for domain/accountStats

src/lib/multicall/Multicall.ts
```diff
@@ -153,7 +153,12 @@ export class Multicall {
     private abFlags: Record<string, boolean>
   ) {}
 
-  async call(providerUrls: MulticallProviderUrls, request: MulticallRequestConfig<any>, maxTimeout: number) {
+  async call(
+    providerUrls: MulticallProviderUrls,
+    request: MulticallRequestConfig<any>,
+    maxTimeout: number,
+    isLargeAccount: boolean
```
divhead, line 160: i think Multicall shouldn't know about large or small accounts, probably it's better to use some technical naming here, like shouldUseAlchemy

src/lib/multicall/Multicall.ts
```diff
@@ -214,6 +218,7 @@ export class Multicall {
           isInMainThread: !isWebWorker,
           requestType,
           rpcProvider,
+          isLargeAccount,
```
divhead, line 221: i think it could be added to useConfigureMetrics hooks and settled as GlobalMetricData

src/lib/account/isLargeAccount.ts
```diff
@@ -0,0 +1,38 @@
...
+
+function getIsLargeAccountStoredValue() {
+  return getStorageItem(IS_LARGE_ACCOUNT_KEY, true) === "true";
```
divhead, line 10: oh, sorry, could you not use getStorageItem? 
This is code for early experiments with metrics, when the plan was to store something in localStorage. 

I'm thinking of removing it because the flag behavior is not obvious

maybe just use localStorage.getItem?

src/lib/account/isLargeAccount.ts
```diff
@@ -0,0 +1,38 @@
...
+  const [isLargeAccountStoredValue, setIsLargeAccountStoredValue] = useLocalStorageSerializeKey<boolean>(
+    IS_LARGE_ACCOUNT_KEY,
+    false
+  );
+
+  useEffect(() => {
+    if (!account) {
+      isLargeAccount = false;
+    } else if (isLargeCurrentAccount !== undefined) {
+      setIsLargeAccountStoredValue(isLargeCurrentAccount);
```
divhead, line 28: isCurrentAccountLarge?

src/domain/synthetics/accountStats/useAccountVolumeStats.ts
onemicky: The difference is that we don't like to collect volume data stats for AVALANCHE_FUJI

src/domain/synthetics/accountStats/useIsLargeAccountData.ts
onemicky: I'm not sure. This is very specific constants that are only related to this feature. I think it's better to keep them with their logic

src/lib/multicall/Multicall.ts
onemicky: My thought is that it's pretty important to have the isLargeAccount flag for all events across the app, even in the multicall. The reason is that it's the most important cohort for us, and it's better to be able to group any metrics by this flag without name mapping

divhead: you can add the flat to GlobalMetricData

divhead: and set it in useConfigureMetrics

onemicky: Yep. This is what i've already done. But as i can see GlobalMetricData works only for events. Not for counter and timings. So i kept this flags for this calls inside multicall.

divhead (approved):

PR #1302, opened 2024-10-10
Positions load speed optimizations

divhead:
---
  - https://app.asana.com/0/0/1208508360257866


src/components/Synthetics/PositionItem/PositionItem.tsx
```diff
@@ -273,21 +279,25 @@ export function PositionItem(p: Props) {
   }
 
   function renderLiquidationPrice() {
+    if (!p.position.marketInfo) {
+      return <Skeleton width={60} count={1} baseColor="#B4BBFF1A" highlightColor="#B4BBFF1A" />;
```
hubert-de-lalye, line 283: Does this intended behavior also in mobile? Why does this differ from `...` in other places? 

Screenshot 2024-10-14 at 13 44 39

src/config/static/sortedMarkets.ts
```diff
@@ -0,0 +1,27 @@
+export const SORTED_MARKETS = [
```
hubert-de-lalye, line 1: Not quite understand what is it and what should I do when I will add new market? Maybe we can automate this and put into prebuilt script?

src/domain/synthetics/positions/utils.ts
```diff
@@ -393,6 +394,10 @@ export function getMinCollateralFactorForPosition(position: PositionInfo, openIn
   return minCollateralFactor;
 }
 
+export function getIsPositionInfoLoaded(pos?: PositionInfo): pos is PositionInfoLoaded {
```
hubert-de-lalye, line 397: Strange signature of typeguard, even not sure why does it work as PositionInfoLoaded is supertype of PositionInfo. Maybe it's better to explicitly say that pos `PositionInfo | PositionInfoLoaded`

hubert-de-lalye (approved):

PR #1303, opened 2024-10-10
Speed optimizations tv

divhead:
---
  - https://app.asana.com/0/0/1208508360257868


src/domain/synthetics/tradingview/SyntheticsTVDataProvider.ts
```diff
@@ -19,7 +19,7 @@ export class SyntheticsTVDataProvider extends TVDataProvider {
     period: string,
     onFallback?: (ex: Error) => void
   ): Promise<FromOldToNewArray<Bar>> {
-    const limit = 5000;
+    const limit = 500;
```
hubert-de-lalye, line 22: Seems that for 1m period this will fetch bars for last ~8 hours. Which doesn't cover full X-axis width, not sure if it's ok. As for me it's ok, as it doesn't make much sense to zoom out more than 12hours range with 1m period, so if it's intended just ignore this commit.

src/domain/tradingview/TVDataProvider.ts
```diff
@@ -140,19 +140,32 @@ export class TVDataProvider {
     return getTokenChartPrice(chainId, ticker, period, onFallback);
   }
 
+  promise?: Promise<any>;
```
hubert-de-lalye, line 143: Seems that `any` could be replaces with return type of `getTokenChartPrice`

src/lib/metrics/Metrics.ts
```diff
@@ -280,12 +282,12 @@ export class Metrics {
 
   handleWindowCounter = (event: Event) => {
     const { detail } = event as CustomEvent;
-    this.pushCounter(detail.event, detail.data);
+    this.pushCounter<any>(detail.event, detail.data);
```
hubert-de-lalye, line 285: This part is pretty confusing, as in `pushCounter/pushTiming` we're expecting pretty strict structure, but in the only one call of `pushCounter/pushTiming` we're manually pass `<any>`, not quite understood why even update pushCounter/pushTiming signature typings

divhead:
fixed

src/lib/metrics/Metrics.ts
divhead: I've updated typing to require specific event name and params, plan to store all types of events in one place in lib/metrics/. It's pretty hard to walk the code and look for all of possible events

in this place **any** is used because this function is window event handler that uses to handle custom window events emitted by emitMetricEvent / emitMetricCounter / emitMetricTiming -  which is typed in the same way as pushEvent / pushCounter / pushTiming

hubert-de-lalye (approved):

hubert-de-lalye (approved):

PR #1334, opened 2024-10-21
1000PEPE init

midas-myth:


src/domain/tradingview/useTVDatafeed.ts
```diff
@@ -181,10 +196,12 @@ function buildFeeder({
           has_intraday: true,
           has_daily: true,
           currency_code: "USD",
-          visible_plots_set: "ohlc",
           data_status: "streaming",
           isStable: stableTokens.includes(symbolName),
-        } as unknown as LibrarySymbolInfo;
+          visualMultiplier,
+          // @ts-ignore
+          visible_plots_set: "ohlc",
```
gmxer, line 203: is it ok?

gmxer (approved):

src/domain/tradingview/useTVDatafeed.ts
midas-myth: its wronfully typed in our type definitions. I decided that instead of changing the whole types .d.ts rn i would ts-ignore this place

Later when migrating to new trading view this will go away

gmxer (approved):

PR #1366, opened 2024-11-04
Fix Market pnl precision

divhead:


src/pages/SyntheticsStats/SyntheticsStats.tsx
```diff
@@ -792,8 +794,22 @@ export function SyntheticsStats() {
               }
 
               const netPnlMax = getMarketNetPnl(market, true);
-              const longPnlMax = getMarketPnl(market, true, true);
-              const shortPnlMax = getMarketPnl(market, false, true);
+              let longPnlMax = getMarketPnl(market, true, true);
+              let shortPnlMax = getMarketPnl(market, false, true);
+
+              longPnlMax = getCappedPoolPnl({
```
gmxer, line 800: let's use different variables, doesn't make sense to reuse variables here

gmxer (changes requested):

gmxer (approved):

PR #1387, opened 2024-11-17
Mixpanel integration

divhead:
---
  - https://app.asana.com/0/0/1207919781222038


gmxer (dismissed):

src/lib/userAnalytics/utils.ts
```diff
@@ -0,0 +1,208 @@
...
+      data: {
+        action: "BuyConfirm",
+        type: "GLV",
+        poolName: metricData.marketName || "",
+        glvAddress: metricData.glvAddress || "",
+        amount: metricData.glvTokenUsd || 0,
+        isFirstBuy: Boolean(metricData.isFirstBuy),
+      },
+    });
+  }
```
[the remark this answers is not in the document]
divhead: I realized that switch / case is not very appropriate here because there could be the same handler for several order types

Screenshot 2024-11-19 at 2 37 57 PM

src/components/Referrals/AffiliatesStats.tsx
```diff
@@ -70,7 +73,15 @@ function AffiliatesStats({
 
   const [isClaiming, setIsClaiming] = useState(false);
   const [, copyToClipboard] = useCopyToClipboard();
-  const open = () => setIsAddReferralCodeModalOpen(true);
+  const open = () => {
```
hubert-de-lalye, line 76: useCallback with `setIsAddReferralCodeModalOpen`

src/components/Referrals/AffiliatesStats.tsx
```diff
@@ -70,7 +73,15 @@ function AffiliatesStats({
...
+    userAnalytics.pushEvent<ReferralCreateCodeEvent>({
+      event: "ReferralCodeAction",
+      data: {
+        action: "CreateCode",
+      },
+    });
+    setIsAddReferralCodeModalOpen(true);
+  };
   const close = () => setIsAddReferralCodeModalOpen(false);
```
hubert-de-lalye, line 85: Ideally this too

src/components/Synthetics/TradeboxPoolWarnings/TradeboxPoolWarnings.tsx
```diff
@@ -133,6 +141,55 @@ export const useTradeboxPoolWarnings = (
...
+  const marketIndexName = marketInfo ? getMarketIndexName(marketInfo) : "";
+  const marketPoolName = marketInfo ? getMarketPoolName(marketInfo) : "";
+
+  useEffect(() => {
+    if (!marketIndexName || !marketPoolName) {
+      return;
+    }
+
+    if (
+      showHasNoSufficientLiquidityInAnyMarketWarning ||
```
hubert-de-lalye, line 154: Hmm, I think this may trigger extra events when we implement automatic pool switching (which now is on Hold)? Say we will implement a hook which switches pools based on user input, and `showHas...` for a short moment may be true. 

Unfortunately can't see good solution, so if you don't have something robust from top of your head - please skip.

src/lib/cookies.ts
```diff
@@ -0,0 +1,14 @@
+export function getCookie(name: string) {
```
hubert-de-lalye, line 1: As we adding common cookies utils let's escape characters with `encodeURIComponent`, otherwise in future we may set cookie value which can broke this code

divhead: oh, we don't use cookie eventually, i've just deleted this file

hubert-de-lalye (approved):

PR #1440, opened 2024-12-10
Update stats borrowing rates

onemicky:
---
  - https://app.asana.com/0/0/1208865803017770

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "prettier": "prettier --write src/**/*.{js,ts,jsx,tsx,css,scss}",
     "typechain": "typechain --target ethers-v6 --out-dir ./src/typechain-types './sdk/src/abis/*.json'",
     "postinstall": "yarn typechain && cd ./sdk && echo \"Building SDK\" && yarn && yarn build",
-    "prebuild": "yarn tsx scripts/prebuild"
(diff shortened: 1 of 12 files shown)


src/domain/synthetics/markets/useMarketsInfoRequest/buildMarketsConfigsRequest.ts
```diff
@@ -220,6 +220,30 @@ export async function buildMarketsConfigsRequest(
           methodName: "getBytes32",
           params: [prebuiltHashedKeys.virtualLongTokenId],
         },
+        optimalUsageFactorLong: {
```
divhead, line 223: if we need these values only on stats page, let's move it to separated request please, we already have too many of calls for markets, plan to separate it to several requests in the future

divhead:
the base branch need to be changed

divhead (changes requested):
left some comments

src/domain/synthetics/markets/useMarketsInfoRequest/buildMarketsConfigsRequest.ts
onemicky: Good suggestion! Moved this keys to a separate prebuilt file and separate request.

divhead (dismissed):

divhead (approved):

PR #1466, opened 2024-12-20
Add sl/tp count to event

divhead:
---
  - https://app.asana.com/0/0/1208988934667683


midas-myth:
Im not sure if the revert commit in the PR is intended, but the [add sl/tp count](https://github.com/gmx-io/gmx-interface/pull/1466/commits/8aed5df40fea895b244d18209066ca1f83901b89) commit is good, so ok from me.

hubert-de-lalye:
I guess revert will disappear after sync

PR #1483, opened 2024-12-31
Final duplicates removal 

hubert-de-lalye:
Code changes beside just duplicates removal: 
 - some refactoring (like in `getExecutionFee`), to separate view from business logic
 - run unit tests on `pre-push` not on `pre-commit`
 - a bunch of unit-tests
 - removed some utils based on `ethers` and rewrite it on `viem` to make it common to SDK and UI (e.g. referral codes encode/decode)

---
  - https://app.asana.com/0/0/1209055419218393


sdk/src/types/markets.ts
```diff
@@ -137,13 +129,8 @@ export type FastMarketInfo = Omit<MarketInfo, keyof MarketPoolTokens | keyof Mar
...
 };
 
-export type ClaimableFunding = {
-  claimableFundingAmountLong: bigint;
-  claimableFundingAmountShort: bigint;
-};
-
-export type ClaimableFundingData = {
-  [marketAddress: string]: ClaimableFunding;
+export type FastMarketInfoData = {
```
divhead, line 132: FastMarketInfo is more UI specific thing as it is roughly speaking just a duplicate query in subsquid to speed up the display of markets

sdk/src/utils/positions.spec.ts
```diff
@@ -0,0 +1,212 @@
+import { MarketInfo } from "types/markets";
```
divhead, line 1: probably it make sense put the tests in the __test__ folder like we do in UI (and also in keeper)
to avoid confusing files with code

sdk/src/types/markets.ts
hubert-de-lalye: Extracted from SDK back to `domain/synthetics/markets/types`

sdk/src/utils/positions.spec.ts
hubert-de-lalye: Move all utils tests into __tests__ folder

divhead (dismissed):

divhead (approved):

PR #1487, opened 2025-01-01
Add testing workflow

divhead:


.github/workflows/update-testing.yml
```diff
@@ -0,0 +1,26 @@
...
+  merge-master-to-test:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
+      - name: Merge master into test
```
hubert-de-lalye, line 18: If it's possible to merge commits from this branch back into release/master then here should be [GPG signature enabled](https://github.com/gmx-io/gmx-interface/blob/release-54/.github/workflows/publish-sdk.yml#L40). If not, just skip this comment

hubert-de-lalye (approved):

PR #1490, opened 2025-01-03
UI version feature

divhead:
---
  - https://app.asana.com/0/0/1208595345585136


src/lib/useHasOutdatedUi.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  if (isDevelopment()) {
+    const localStorageVersion = localStorage.getItem(REQUIRED_UI_VERSION_KEY);
+
+    if (localStorageVersion) {
+      hasOutdatedUi = Boolean(parseFloat(localStorageVersion) > UI_VERSION);
+    }
+  }
+
+  return { data: hasOutdatedUi, mutate };
```
gmxer, line 41: let's simply return hasOutdatedUi as a boolean, without object with mutate

gmxer (changes requested):

gmxer (dismissed):

gmxer (approved):

PR #1491, opened 2025-01-04
Validate signer and receiver

divhead:
---
  - https://app.asana.com/0/0/1209075491811362


hubert-de-lalye (approved):
Ok for me, but I wonder should we take any action for SDK? We don't support subaccounts for SDK and don't use ethers either, so seems like no, but would like to confirm

hubert-de-lalye (approved):

PR #1526, opened 2025-01-22
fix glv apy calculation

hubert-de-lalye:


hubert-de-lalye:
Closed in favor of #1527 1527

PR #1583, opened 2025-02-26
Rpc fallbacks

divhead:


src/config/chains.ts
```diff
@@ -232,8 +226,51 @@ export function getChainName(chainId: number) {
...
+  priority: "public" | "private",
+  blackListedUrls: string[] = []
+): string {
+  let url: string;
+
+  const filteredPublicUrls = RPC_PROVIDERS[chainId].filter((url) => !blackListedUrls.includes(url));
+  const filteredPrivateUrls = PRIVATE_PROVIDERS[chainId].filter((url) => !blackListedUrls.includes(url));
+
+  const [urls1, urls2] =
+    priority === "public" ? [filteredPublicUrls, filteredPrivateUrls] : [filteredPrivateUrls, filteredPublicUrls];
```
gmxer, line 260: I think whole function could be just

```
priority === 'public' ? getPublic() ?? getPrivate() : getPrivate() ?? getPublic();
```

src/lib/rpc/bestRpcTracker.ts
```diff
@@ -166,59 +160,140 @@ async function getBestRpcProvidersForChain({ providers, chainId }: RpcTrackerSta
...
-    if (privateRpcResult) {
-      nextPrimaryRpc = privateRpcResult;
+    // Do new probe if can't use cached results
+    if (!probeResults) {
+      const filteredProviders = Object.values(providers).filter(({ url }) => !blackListedUrls.includes(url));
+      const providersList = getIsLargeAccount()
+        ? filteredProviders
+        : filteredProviders.filter(({ isPublic }) => isPublic);
+
+      probeResults = await probeProviders(chainId, providersList);
```
gmxer, line 208: in case of empty list shouldn't we whitelist back some of rpcs here?

src/lib/rpc/bestRpcTracker.ts
```diff
@@ -450,3 +525,33 @@ export function useCurrentRpcUrls(chainId: number) {
...
+
+  emitMetricCounter<RpcFailureCounter>({
+    event: "rpcTracker.provider.failed",
+    data: {
+      rpcProvider: getProviderNameFromUrl(rpcUrl),
+    },
+  });
+
+  // Recalculate best providers with failed RPC blacklisted
+  const nextProviderUrls = await getBestRpcProvidersForChain(chainState, [rpcUrl]);
```
gmxer, line 542: am I understanding correctly that there could be only one blacklisted rpc at time? 
why don't you store blacklistedRpcs?

gmxer (changes requested):

PR #1718, opened 2025-05-30
Hotfix: cancel market orders

qwinndev:
to test it you can you console, example:

```
window.mockMarketOrder_updatedAtTime = 1748620794n
```

---
  - https://app.asana.com/0/0/1210414084798145


src/components/Synthetics/OrderList/OrderList.tsx
```diff
@@ -183,6 +188,8 @@ export function OrderList({
     }
   }, []);
 
+  const oracleSettings = useSelector(selectOracleSettings);
```
midas-myth: if we already use `useSelector` in list item why not move this to the list item too?

src/domain/synthetics/orders/useDisabledCancelMarketOrderMessage.ts
```diff
@@ -0,0 +1,38 @@
...
+import { isMarketOrderType } from ".";
+import { OrderInfo } from ".";
+import { OracleSettingsData } from "../common/useOracleSettingsData";
+
+export const useDisabledCancelMarketOrderMessage = (
+  order: OrderInfo,
+  oracleSettings: OracleSettingsData | undefined
+) => {
+  const [diff, setDiff] = useState(
+    Number(order.updatedAtTime) + Number(oracleSettings?.requestExpirationTime) - Date.now() / 1000
```
midas-myth: Would not it be NaN when oracleSettings is undefined ?

divhead (approved):

PR #1719, opened 2025-06-01
Gassless fixes

divhead:


src/components/SettingsModal/SettingsModal.tsx
```diff
@@ -182,38 +198,30 @@ export function SettingsModal({
     if (!subaccountState.subaccount) {
       const days = secondsToPeriod(DEFAULT_SUBACCOUNT_EXPIRY_DURATION, "1d");
 
-      return plural(Number(days), {
-        one: "1 day",
-        other: `${days} days`,
-      });
+      return `${days} days`;
```
gmxer, line 201: days always > 1?
same questions below, why remove `plural`?

src/config/localStorage.ts
```diff
@@ -64,7 +64,7 @@ export const LP_INTERVIEW_INVITATION_SHOWN_TIME_KEY = "lp-interview-invitation-s
  */
 export const TOKEN_FAVORITE_PREFERENCE_SETTINGS_KEY = "token-favorite-preference";
 export const SETTINGS_WARNING_DOT_VISIBLE_KEY = "settings-warning-dot-visible";
-export const EXPRESS_TRADING_GAS_TOKEN_SWITCHED = "express-trading-gas-token-switched";
+export const EXPRESS_TRADING_GAS_TOKEN_SWITCHED = "gas-token-switched";
```
gmxer: is this intentional change? I think `express-trading-gas-token-switched-v2` or similar would be more transparent then

src/lib/useBlockNumber.ts
```diff
@@ -0,0 +1,19 @@
...
+
+    if (!blockNumber) {
+      return undefined;
+    }
+
+    return BigInt(blockNumber);
+  });
+
+  return data;
+}
```
gmxer, line 23: does useSWR use some default interval here? maybe it's me but not very obvious that it would be refreshed

src/domain/synthetics/orders/simulation.ts
```diff
@@ -75,7 +75,7 @@ export async function simulateExecution(chainId: number, p: SimulateExecuteParam
     blockTag = Number(result.blockNumber);
   }
 
-  const priceTimestamp = blockTimestamp + 10n;
+  const priceTimestamp = blockTimestamp + 30n;
```
gmxer, line 78: could you please explain or give me link to an issue?

gmxer (changes requested):

src/components/SettingsModal/SettingsModal.tsx
divhead: it's rounded down, if we have 24+ hours left, we display 2 days

If it's less than 24 hours, we display “x” hours, and so on

So plural is not necessary

src/domain/synthetics/orders/simulation.ts
divhead: This has long been fixed in the prod, but missed in express

PR #1778, opened 2025-07-07
Hotfix: fix ui fee receiver format

divhead:


sdk/src/utils/twap/uiFeeReceiver.ts
```diff
@@ -55,5 +54,5 @@ export function isValidTwapUiFeeReceiver(address: string) {
 export function setUiFeeReceiverIsExpress(uiFeeReceiver: string, isExpress: boolean): string {
```
qwinndev, line 54: let's add test for this function

qwinndev (dismissed):

PR #1794, opened 2025-07-17
Liquidty section

mdev-fe:


src/pages/Home/sections/LiqiuditySection/components/PoolCard.tsx
```diff
@@ -0,0 +1,66 @@
+import { Trans } from "@lingui/macro";
```
divhead, line 1: looks like too deep nesting

We usually try to keep flat structure 

Probably just put it in src/pages/Home would be enough or even in src/components

divhead (changes requested):
Would be good to keep folders structure flat

src/pages/Home/sections/LiqiuditySection/components/PoolCardContainer.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { PoolCard } from "./PoolCard";
+
+type Pool = "GM" | "GLV" | "GMX";
+
+type Props = {
+  pool: Pool;
+};
+
+const DESCRIPTIONS: Record<Pool, string> = {
+  GMX: t`Stake for rewards and governance rights`,
```
divhead: oh, these translations won't work, since (t) should be called in render lifecycle

src/pages/Home/sections/LiqiuditySection/components/PoolCardContainer.tsx
```diff
@@ -0,0 +1,57 @@
...
+  GM: IcGmPool,
+};
+
+const NAMES: Record<Pool, string> = {
+  GMX: "GMX",
+  GLV: "GLV",
+  GM: "GM",
+};
+
+const COIN_IMAGES: Record<Pool, string> = {
```
divhead: probably it could be a single object like

{
name:
icon:
image:
description:
}

src/pages/Home/sections/LiqiuditySection/components/PoolCard.tsx
mdev-fe: If you allow me I would prefer to leave it here, it is simply used only in one place, created for ease of perception. Plus for separated tailwindcss configs, it will be much easier to configure purge only for the Home folder and single rare files.

src/pages/Home/sections/LiqiuditySection/LiqiuditySection.tsx
qwinndev, line 1: maybe `src/components/home/LiquiditySection/LiquiditySection.tsx`?

PR #1805, opened 2025-07-30
fix express errors

divhead:
---
  - https://app.asana.com/0/0/1210935930884816


src/lib/wallets/signing.ts
```diff
@@ -75,7 +78,19 @@ export async function signTypedData({
     message: typedData,
   };
 
-  const signature = await (provider as any).send("eth_signTypedData_v4", [from, JSON.stringify(eip712)]);
+  const signature = await withRetry<string>(
+    () => {
+      return (provider as any).send("eth_signTypedData_v4", [from, JSON.stringify(eip712)]);
+    },
+    {
+      retryCount: 2,
```
gdev8317: i think it's possible that error happens after user clicks "Sign" button in the wallet. in that case it might be a bit weird UX: user will see sign popup 3 times in a row. so maybe safer to use 1 retry

i'm actually thinking maybe it's better to just instantly fall back to typical approvals instead, what do you think? i.e. if this error happens then automatically start approval process instead (i.e. try to send transaction with approval). not sure if there could be any issues with that fallback

divhead: The error in this case is not for approval but for order creation

divhead: decreased to 1 retry

gdev8317 (approved):

PR #1814, opened 2025-08-11
Gemini wallet

midas-myth:
---
  - https://app.asana.com/0/0/1211019320378516


divhead (dismissed):

src/lib/wallets/rainbowKitConfig.ts
```diff
@@ -34,6 +35,7 @@ const popularWalletList: WalletList = [
       injectedWallet,
       // The Safe option will only appear in the Safe Wallet browser environment.
       safeWallet,
+      geminiRainbowKitConnector,
```
gmxer, line 38: doc says it should be wrapped into a function,
would be good to understand, why
https://www.npmjs.com/package/@gemini-wallet/rainbow
image

PR #1845, opened 2025-08-29
Hotfix: added Safe accounts guide for claims

hubert-de-lalye:


src/domain/synthetics/claims/useIsSmartAccountSigned.ts
```diff
@@ -0,0 +1,71 @@
+import { Contract } from "ethers";
```
divhead: 1. we're trying to migrate to viem from ethers, you can use viem functions to call contract
2. hash and 0x1626ba7e quite not readable, could we work with original values and hash them in runtime to understand what's happening?
3. Would be good to move abi to sdk/abis
4. loooks like you can use swr instead of useState + useEffect

divhead (approved):

PR #1901, opened 2025-09-17
Use keeper performance apis

qwinndev:
---
  - https://app.asana.com/0/0/1211331056920540
  - https://app.asana.com/0/0/1211395167844185


sdk/src/utils/numbers.ts
```diff
@@ -571,13 +571,24 @@ export const padDecimals = (amount: BigNumberish, minDecimals: number) => {
   return amountStr;
 };
 
-export function getPlusOrMinusSymbol(value?: bigint, opts: { showPlusForZero?: boolean } = {}): string {
+export function getPlusOrMinusSymbol(
```
hubert-de-lalye: I think it's odd to have a function named getPlusOrMinusSymbol which by options may return "" for positive values. Maybe omitting plus sign can be handled on top layer? Like 

```
const sign = getPlusOrMinusSymbol(percentage);
const displaySign = signed & sign !== "+" ? sign : "";
```

Still not sure, but logic seems complicated here

qwinndev: sounds reasonable, will fix

PR #1923, opened 2025-09-23
V2.2 Part 3

qwinndev:
---
  - https://app.asana.com/0/0/1211371181287610

### DIFF
--- a/sdk/src/utils/fees/priceImpact.ts
+++ b/sdk/src/utils/fees/priceImpact.ts
@@ -513,19 +513,13 @@ export function applyImpactFactor(diff: bigint, factor: bigint, exponent: bigint
 export function getCappedPriceImpactPercentageFromFees({
   fees,
   isSwap,
-  isIncrease,
 }: {
   fees: TradeFees | undefined;
   isSwap: boolean;
-  isIncrease: boolean;
 }): bigint | undefined {
   if (isSwap) {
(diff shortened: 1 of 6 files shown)


sdk/src/utils/__tests__/fees.spec.ts
```diff
@@ -42,10 +42,10 @@ describe("getFundingFactorPerPeriod", () => {
     } as MarketInfo;
 
     const forLongs = getFundingFactorPerPeriod(marketInfo, true, second);
-    expect(forLongs.toString()).toBe(toFactor("50%").toString());
```
qwinndev, line 45: funding calc update task https://app.asana.com/1/1538917451150/project/1204313444805313/task/1211371181287618?focus=true

sdk/src/utils/fees/priceImpact.ts
```diff
@@ -508,3 +509,17 @@ export function applyImpactFactor(diff: bigint, factor: bigint, exponent: bigint
 
   return result;
 }
+
+export function getCappedPriceImpactPercentageFromFees({
```
divhead, line 513: for increase actions we don't use totalPendingImpact, it's only for decrease orders
better value is positionNetPriceImpact

divhead: Or let's explicitly specify isIncrease flag

PR #1935, opened 2025-09-26
Migrate to typescript ABIs

midas-myth:
---
  - https://app.asana.com/0/0/1211409663145378

### DIFF
--- a/sdk/src/abis/UniswapV3Factory.ts
+++ b/sdk/src/abis/UniswapV3Factory.ts
@@ -1,0 +1,198 @@
+export default [
+  {
+    anonymous: false,
+    inputs: [
+      {
+        indexed: true,
+        internalType: "uint24",
+        name: "fee",
+        type: "uint24",
+      },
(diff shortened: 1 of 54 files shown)


src/components/UserIncentiveDistributionList/utils.tsx
```diff
@@ -40,22 +40,18 @@ export async function checkValidity({
...
-        abi: abis.SmartAccount,
-        functionName: "isValidSignature",
-        args: [CLAIM_TERMS_HASH, "0x"],
-      })
-      .catch(() => "0x"),
+    publicClient.readContract({
+      address: account,
+      abi: abis.SmartAccount,
+      functionName: "isValidSignature",
+      args: [hash, claimTermsAcceptedSignature ?? "0x"],
```
divhead: should not be called with zero address

midas-myth: done

PR #1943, opened 2025-10-02
Hotfix: Force trading view show

midas-myth:


src/components/TVChartContainer/TVChartContainer.tsx
```diff
@@ -215,6 +218,30 @@ export default function TVChartContainer({
       });
     });
 
+    setTimeout(() => {
```
qwinndev: let's descibe why we do that in comment. should be helpful if leave it for long time

PR #1948, opened 2025-10-06
Support limited squid queries

hubert-de-lalye:
---
  - https://app.asana.com/0/0/1211660635385833

### DIFF
--- a/sdk/src/modules/markets/index.ts
+++ b/sdk/src/modules/markets/index.ts
@@ -479,7 +479,7 @@ type PositionVolumeInfosResponse = {
 
 const POSITIONS_VOLUME_INFOS_QUERY = `
 {
-  positionsVolume(where: {period: "1d}) {
+  positionsVolume(where: {period: "1d"}) {
     market
     volume
   }
--- a/sdk/src/utils/subgraph.ts
+++ b/sdk/src/utils/subgraph.ts
@@ -128,10 +128,10 @@ export async function queryPaginated<T>(
 
   // eslint-disable-next-line no-constant-condition
   while (true) {
-    const res = await fetcher(limit, offset);
-    results.push(...res);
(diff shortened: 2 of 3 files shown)


sdk/src/modules/markets/index.ts
```diff
@@ -487,8 +478,8 @@ type PositionVolumeInfosResponse = {
 };
 
 const POSITIONS_VOLUME_INFOS_QUERY = `
-query PositionVolumeInfoResolver($timestamp: Float!) {
-  positionsVolume(where: {timestamp: $timestamp}) {
+{
+  positionsVolume(where: {period: "1d}) {
```
qwinndev: looks like there is missing "

sdk/src/utils/subgraph.ts
```diff
@@ -115,3 +117,26 @@ export function buildFiltersBody(filters: GraphQlFilters, options?: { enums?: Re
...
+  let offset = 0;
+
+  const results: T[] = [];
+
+  // eslint-disable-next-line no-constant-condition
+  while (true) {
+    const res = await fetcher(limit, offset);
+    results.push(...res);
+
+    if (res.length < limit) {
```
qwinndev: i suppose results length should be checked, isn't it?

qwinndev (changes requested):

sdk/src/modules/markets/index.ts
hubert-de-lalye: You're absolutely right!

sdk/src/utils/subgraph.ts
hubert-de-lalye: No, results may have any length, but if res.length < limit it means that it was last page. Will rename res to page

qwinndev (approved):

PR #2370, opened 2026-04-06
Net Rates (and Borrowing / Funding) on Chart fedev-2137

timgmxio:


src/components/NetRateChart/NetRateChart.tsx
```diff
@@ -0,0 +1,471 @@
...
+    return buildChartData(snapshots, activeRateType, activeProjection);
+  }, [snapshots, activeRateType, activeProjection]);
+
+  const yAxisDecimals = useMemo(() => {
+    if (chartData.length === 0) return 4;
+    const allValues = chartData.flatMap((d) => [d.longRate, d.shortRate]);
+    return getPercentageDisplayDecimals(allValues);
+  }, [chartData]);
+
+  const rateTypeTabs = useMemo(() => RATE_TYPES.map((type) => ({ label: RATE_TYPE_LABELS[type], value: type })), []);
```
qwinndev: since you are using `t` inside useMemo it will not get updated on language change. in such cases you can use <Trans>label</Trans> or useLocalizedMap

src/components/NetRateChart/NetRateChart.tsx
```diff
@@ -0,0 +1,471 @@
...
+
+  const timeframeTabs = useMemo(() => TIMEFRAMES.map((tf) => ({ label: TIMEFRAME_LABELS[tf], value: tf })), []);
```
qwinndev: same problem with labels

src/components/NetRateChart/NetRateChart.tsx
```diff
@@ -0,0 +1,471 @@
...
+    case "netRate":
+      return { longKey: "netRateLong", shortKey: "netRateShort" };
+    case "borrowingRate":
+      return { longKey: "borrowingRateLong", shortKey: "borrowingRateShort" };
+    case "fundingRate":
+      return { longKey: "fundingRateLong", shortKey: "fundingRateShort" };
+  }
+}
+
+function rateToProjectedPercent(rateString: string, projection: RateProjection): number {
```
qwinndev, line 92: there is a chance that you will lose precision because of usage Number. but you can avoid it this way:

```
const PRECISION = 10n ** 30n;
function rateToProjectedPercent(rateString: string, projection: RateProjection): number {
  const rate = BigInt(rateString);
  const multiplier = BigInt(PROJECTION_MULTIPLIERS[projection]);
  return Number((rate * multiplier * 100n) / PRECISION);
}
```

src/components/NetRateChart/NetRateChart.tsx
```diff
@@ -0,0 +1,471 @@
...
+  raw: RatesSnapshot;
+  projection: RateProjection;
+  suffix: string;
+  decimals: number;
+}) {
+  const longNet = rateToProjectedPercent(raw.netRateLong, projection);
+  const shortNet = rateToProjectedPercent(raw.netRateShort, projection);
+  const longFunding = rateToProjectedPercent(raw.fundingRateLong, projection);
+  const shortFunding = rateToProjectedPercent(raw.fundingRateShort, projection);
+  const longBorrowing = rateToProjectedPercent(raw.borrowingRateLong, projection);
```
qwinndev: i am not sure, but seems like there is a problem with borrowing signs. i see values always positive which should not be

src/components/NetRateChart/NetRateChart.tsx
```diff
@@ -0,0 +1,471 @@
...
+      }
+      return format(value, "dd/MM");
+    },
+    [activeTimeframe]
+  );
+
+  const renderTooltip = useCallback(
+    ({ active, payload }: TooltipProps<number, string>) => {
+      if (!active || !payload || payload.length === 0) return null;
+      const point = payload[0]!.payload as ChartDataPoint;
```
qwinndev: i would add type guard utility to avoid type casting and ! sign

src/domain/synthetics/markets/useRateSnapshots.ts
```diff
@@ -0,0 +1,56 @@
...
+      },
+      refreshInterval: REFRESH_INTERVAL,
+    }
+  );
+
+  const isWaitingForDeps = !sdk || !marketAddress;
+
+  return {
+    snapshots: data,
+    isLoading: isLoading || (isWaitingForDeps && !data),
```
qwinndev: this will result infinite loading on unsupported api chains

src/locales/de/messages.po
```diff
@@ -512,6 +512,11 @@ msgstr "Token auswählen"
 msgid "Image generation failed. Refresh and try again."
 msgstr "Bilderzeugung fehlgeschlagen. Aktualisieren Sie die Seite und versuchen Sie es erneut."
 
+#: src/components/NetRateChart/NetRateChart.tsx
+#: src/components/NetRateChart/NetRateChart.tsx
+msgid "Funding"
+msgstr ""
```
qwinndev: you can ask llm to add translations, would be good

qwinndev (changes requested):

src/components/NetRateChart/NetRateChart.tsx
timgmxio: `RATE_TYPE_LABELS` are translated using `<Trans>` component, so it works as expected

qwinndev (dismissed):

qwinndev (dismissed):

qwinndev (approved):

PR #2429, opened 2026-05-04
FEDEV-3706 fix derived leverage with limit price

qwinndev:
## Summary

Fixes [FEDEV-3706](https://linear.app/gmx-io/issue/FEDEV-3706/bug-displayed-leverage-is-incorrect-when-manual-leverage-is-disabled). With manual leverage disabled and a limit order set, the auto-derived leverage shown in the leverage field could diverge from `displayedSize / displayedMargin` because `sizeDeltaUsd` was computed from a stale `toTokenInputValue` that had been derived using a different conversion price than the current trigger price.

## Root cause

In USD display mode the user-typed Size USD is stored in `sizeInputValue`; `toTokenInputValue` is derived via `usdToTokens` using the current `sizeConversionPrice`:

- For market orders, `sizeConversionPrice = markPrice`.
- For limit orders, `sizeConversionPrice = triggerPrice`.

The selector then computes `sizeDeltaUsd = indexTokenAmount × indexPrice` (where `indexPrice = triggerPrice` for limit orders).

When a user typed Size before setting a limit price (or before changing it), `toTokenInputValue` was derived using the old conversion price. After the trigger price changed, the existing passive sync effect bailed out when `focusedInput === "to"`, so the stale token amount remained — and `sizeDeltaUsd` (and the auto-derived leverage built from it) no longer matched the displayed Size USD.

## Fix

Treat the user-typed Size USD as the anchor for limit orders. When `triggerPrice` changes in USD mode, set the existing `shouldForceSizeUsdSyncRef` flag so the next sync pass re-derives `toTokenInputValue` from `sizeInputValue` using the new conversion price — overriding the focus guard the way other forced syncs (market/token swaps) already do. The user-typed Size USD remains untouched.

The behaviour is scoped to limit/stop orders in USD display mode; market mode and token mode are not affected.

## Test plan

- [x] `tsc --noEmit` clean
- [x] `eslint` clean on changed files
- [x] Added Playwright CT regression test in `TradeboxMarginFieldsIntegration.ct.spec.tsx` covering: type Size USD with no trigger price, then set a limit price, and assert the token amount re-derives from the typed USD
- [ ] Manual: with manual leverage off, Long, Limit, GMX-USDC, collateral GMX, Margin = 150 GMX, Size = 1500, Limit = 1 — leverage matches displayed size/margin
- [ ] Manual: same flow in token display mode — leverage and displayed values stay consistent
- [ ] Manual: Short side reproduces the same correct behaviour
- [ ] Manual: Stop / market modes unchanged
- [ ] Manual: liquidation price and fees still align with the displayed leverage

### DIFF
--- a/src/components/TradeboxMarginFields/TradeboxMarginFields.tsx
+++ b/src/components/TradeboxMarginFields/TradeboxMarginFields.tsx
@@ -65,6 +65,8 @@ export function TradeboxMarginFields({
   const { toTokenAddress, marketAddress } = useSelector(selectTradeboxState);
 
   const prevSizeSyncContextRef = useRef<string>();
+  // null sentinel = "uninitialized"; bigint | undefined = previously seen value (undefined means no triggerPrice)
+  const prevTriggerPriceRef = useRef<bigint | undefined | null>(null);
   const shouldForceSizeUsdSyncRef = useRef(false);
 
   const toToken = getByKey(tokensData, toTokenAddress);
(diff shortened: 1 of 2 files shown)


qwinndev:
Superseded by inclusion in `tradebox-todo-review` (commit 37daaf8a6). Note: regression test was dropped because the test infrastructure isn't on that branch.

PR #2455, opened 2026-05-14
Fix invalid liquidation price from Edit Margin withdraw max (FEDEV-3737)

timgmxio:
In the Edit Margin > Withdraw modal, the displayed Max amount (and the 100% slider) could produce a submission whose nextLiqPrice crossed markPrice, leaving the modal blocked with "Invalid liquidation price".

Root cause: maxWithdrawAmount reserved only the leverage threshold (minCollateralFactor + global minCollateralUsd + pending fees). It did not account for the inputs getLiquidationPrice uses internally — closing fee, capped negative price impact, pendingImpactAmount, and the stricter minCollateralFactorForLiquidation threshold. For losing positions or positions with negative pending impact, the liquidation constraint can be tighter than the leverage one, so the max was unsafe.

Changes:
- Add getMinCollateralUsdForLiquidationPrice (SDK): solves `nextLiqPrice = markPrice` against the formula in getLiquidationPrice and returns the minimum nextCollateralUsd that keeps the position safe.
- Add getMaxWithdrawAmount (SDK): pure helper combining leverage, global minCollateralUsd, and liquidation-price constraints. Replaces inline duplicates in PositionEditor.tsx and usePositionEditorButtonState.tsx so the slider, displayed Max, and detectAndSetMaxSize binary search agree.
- Show "Withdrawal not available" CTA when no safe withdraw exists, instead of falling through to the generic "Enter an amount".
- Unit tests for getMinCollateralUsdForLiquidationPrice plus an integration regression test that builds a losing LONG with negative pending impact and verifies nextLiqPrice stays on the safe side of markPrice.


src/locales/de/messages.po
```diff
@@ -1866,6 +1862,10 @@ msgstr "Schlüssel"
 msgid "Distributions"
 msgstr "Verteilungen"
 
+#: src/domain/synthetics/trade/utils/validation.ts
+msgid "Withdrawal not available"
+msgstr ""
```
qwinndev: lets add translations to other languages

sdk/src/utils/trade/__tests__/getMaxWithdrawAmount.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+  function buildScenario() {
+    const tokensData = mockTokensData({
+      ETH: { prices: { minPrice: ETH_PRICE, maxPrice: ETH_PRICE } },
+      USDC: { prices: { minPrice: USDC_PRICE, maxPrice: USDC_PRICE } },
+    });
+
+    const marketKey = "ETH-ETH-USDC";
+    const marketsInfoData = mockMarketsInfoData(tokensData, [marketKey], {
+      [marketKey]: {
+        // Deep balanced OI suppresses getPriceImpactForPosition so pendingImpactAmount
```
qwinndev: this comment looks wrong. settings balanced or imbalanced OI does not change test behaviou anyhow

sdk/src/utils/positions/utils.ts
```diff
@@ -295,6 +295,59 @@ export function getLiquidationPrice(p: {
   return liquidationPrice;
 }
 
+// Solves `nextLiqPrice = markPrice` against the formula in getLiquidationPrice above; body
```
qwinndev: i would remove this comment, does not make much sense and just confuse imo

qwinndev (approved):

PR #2464, opened 2026-05-15
Hotfix: Support v2 JIT liquidity info in UI FEDEV-3885

qwinndev:
## Summary
- Add UI flag control for `isV2JitLiquidityInfoEnabled`.
- Switch JIT liquidity fetching from `/v1/jit/liquidity_info` to `/v2/jit/liquidity_info` when the flag is enabled.
- Parse v2 direction-specific `long` / `short` liquidity info and pass the matching JIT GLV shift params into order simulation.
- Cover v1 and v2 JIT liquidity parsing with unit tests.

## Root Cause
The UI only understood the v1 JIT liquidity response shape. The v2 endpoint nests liquidity and shift params by direction, so the parser and transaction simulation path needed to become version-aware.

Ticket: [FEDEV-3885](https://linear.app/gmx-io/issue/FEDEV-3885/support-v2jitliquidity-info-in-ui)

### DIFF
--- a/src/domain/synthetics/jit/useJitLiquidityRequest.ts
+++ b/src/domain/synthetics/jit/useJitLiquidityRequest.ts
@@ -4,15 +4,18 @@ import useSWR from "swr";
 import { getApiUrl } from "sdk/configs/api";
 import type { ContractsChainId } from "sdk/configs/chains";
 
-import { JitLiquidityData, JitLiquidityInfo, safeParseBigInt } from "./utils";
+import { JitLiquidityData, JitLiquidityInfo, parseJitLiquidityResponse } from "./utils";
+import { getIsV2JitLiquidityInfoEnabled, useUiFlagsRequest } from "../uiFlags/useUiFlagsRequest";
 
 const JIT_LIQUIDITY_UPDATE_INTERVAL = 30 * 1000;
 
 export function useJitLiquidityRequest(chainId: ContractsChainId, options?: { enabled?: boolean }): JitLiquidityData {
   const enabled = options?.enabled !== false;
+  const { uiFlags } = useUiFlagsRequest();
+  const isV2JitLiquidityInfoEnabled = getIsV2JitLiquidityInfoEnabled(uiFlags);
 
   const { data } = useSWR<Record<string, JitLiquidityInfo> | undefined>(
-    enabled ? ["jitLiquidity", chainId] : null,
+    enabled ? ["jitLiquidity", chainId, isV2JitLiquidityInfoEnabled ? "v2" : "v1"] : null,
     async () => {
       try {
         const apiUrl = getApiUrl(chainId);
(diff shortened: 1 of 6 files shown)


src/domain/synthetics/jit/useJitLiquidityRequest.ts
```diff
@@ -62,3 +47,11 @@ export function useJitLiquidityRequest(chainId: ContractsChainId, options?: { en
     [data]
   );
 }
+
+function getJitApiUrl(apiUrl: string, isV2JitLiquidityInfoEnabled: boolean): string {
+  if (!isV2JitLiquidityInfoEnabled) {
+    return apiUrl;
+  }
+
+  return apiUrl.replace(/\/v1\/?$/, "/v2");
```
gmxer: what's the point of such replacement, let's use config / constant instead?

src/domain/synthetics/jit/utils.ts
```diff
@@ -33,7 +35,138 @@ export function getJitMaxReservedUsd(
...
+    if (!market) {
+      continue;
+    }
+
+    if (isV2JitLiquidityInfoEnabled) {
+      result[market.toLowerCase()] = parseV2JitLiquidityInfo(rawInfo);
+      continue;
+    }
+
+    result[market.toLowerCase()] = parseV1JitLiquidityInfo(rawInfo);
```
gmxer: just in case, is it UI-specific that market address is lower cased here? will it properly work?

src/domain/synthetics/jit/utils.ts
```diff
@@ -33,7 +35,138 @@ export function getJitMaxReservedUsd(
...
+    }
+
+    const market = getString(rawInfo.market);
+
+    if (!market) {
+      continue;
+    }
+
+    if (isV2JitLiquidityInfoEnabled) {
+      result[market.toLowerCase()] = parseV2JitLiquidityInfo(rawInfo);
```
gmxer: werid that `isV2JitLiquidityInfoEnabled` is pointless because `parseV2JitLiquidityInfo` includes `parseV1JitLiquidityInfo`

qwinndev: bug, fixed

src/domain/synthetics/jit/utils.ts
qwinndev: yes, it is behaviour from existing code, jit addressed lowercased, but there is no need. removed lowercasing for jit

src/domain/synthetics/jit/useJitLiquidityRequest.ts
qwinndev: it was easy path to test v2 route, improved getting api url for better usage

gmxer (dismissed):

gmxer (approved):

PR #2483, opened 2026-05-21
(test-only) Add order execution stats page

qwinndev:
## Summary
- Adds an order execution stats page with success-rate charting by time bucket.
- Adds order type, market, date range, and bucket-size controls.
- Adds a failure-reasons table with decoded reasons and row/header checkboxes that exclude selected failed/cancelled reasons from the chart.
- Links the page from stats navigation and extends the shared date picker options needed by this view.


qwinndev:
Combined into #2556.

PR #2488, opened 2026-05-23
Set Privy wallet timeout to max safe integer

qwinndev:
## Summary

Sets Privy's wallet request timeout override to `Number.MAX_SAFE_INTEGER` for external wallet RPC/signature requests.

Privy's SDK defaults these requests to 120000 ms and only supports wallet-client keyed timeout overrides. This change provides a proxy-backed timeout map so any wallet client key resolves to the max safe integer without hardcoding Privy's wallet client list.

## Impact

Users should no longer hit Privy's 2 minute wallet request timeout while waiting on wallet confirmation flows.

### DIFF
--- a/src/lib/wallets/WalletProvider.tsx
+++ b/src/lib/wallets/WalletProvider.tsx
@@ -14,6 +14,7 @@ import {
   getSupportedChains,
   PRIVY_APP_ID,
   PRIVY_LOGIN_METHODS,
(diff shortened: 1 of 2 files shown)


qwinndev:
Superseded by hotfix PR against master: https://github.com/gmx-io/gmx-interface/pull/2489

PR #2553, opened 2026-06-10
Fix GLV simulation price path FEDEV-3971

qwinndev:
Linear: https://linear.app/gmx-io/issue/FEDEV-3971/fix-glv-simulation-when-new-constituent-prices-are-not-ui-listed

## Root cause

GLV execution can use the GLV primary price to value the vault without requiring primary prices for every constituent market. The UI was dropping direct GLV ticker prices because concrete GLV addresses are not normal static tokens, so GLV simulation fell back to constituent prices and could fail with `EmptyPrimaryPrice` for newly staged markets such as SPCX.

## Changes

- Preserve ticker prices for configured GLV token addresses.
- Prefer direct GLV ticker prices in GLV market data when available.
- Include the selected GLV token price in settlement-chain GLV deposit and withdrawal simulations.
- Add a focused regression test for GLV ticker address handling.

### DIFF
--- a/src/domain/synthetics/tokens/useTokenRecentPricesData.ts
+++ b/src/domain/synthetics/tokens/useTokenRecentPricesData.ts
@@ -2,13 +2,14 @@ import { useMemo } from "react";
 import { useLocation } from "react-router-dom";
 
 import { ContractsChainId } from "config/chains";
+import { GLV_MARKETS } from "config/markets";
 import { parseContractPrice, TokenPricesData } from "domain/synthetics/tokens";
 import { FreshnessMetricId, metrics, TickersErrorsCounter, TickersPartialDataCounter } from "lib/metrics";
 import { freshnessMetrics } from "lib/metrics/reportFreshnessMetric";
 import { _debugOracleKeeper } from "lib/oracleKeeperFetcher/_debug";
 import { useOracleKeeperFetcher } from "lib/oracleKeeperFetcher/useOracleKeeperFetcher";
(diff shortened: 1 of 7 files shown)


divhead (approved):

src/domain/synthetics/tokens/useTokenRecentPricesData.ts
```diff
@@ -130,3 +126,22 @@ export function useTokenRecentPricesRequest(
...
+    const glvConfig = Object.values(GLV_MARKETS[chainId] ?? {}).find(
+      (glv) => glv.glvTokenAddress.toLowerCase() === tokenAddress.toLowerCase()
+    );
+
+    if (!glvConfig) {
+      return undefined;
+    }
+
+    return {
+      ...getTokenBySymbol(chainId, "GLV"),
```
gmxer, line 143: there're two different GLVs on arbitrum (ETH GLV and BTC GLV), you sure it would work this way?

divhead: i guess address solves it

gmxer: ok

gmxer (approved):

src/domain/synthetics/tokens/useTokenRecentPricesData.ts
qwinndev: yes it will work because currently it's just stub token object in our config

Image

PR #2557, opened 2026-06-10
Warn when funding fee settlement may fail from negative margin fedev-3907

timgmxio:
The zero-size settle order doesn't realize PnL, so when the margin after pending borrow and funding fees is negative there is nothing to pay the funding side with, and the order gets cancelled with InsufficientFundsToPayForCosts("funding"). Such positions are no longer preselected in the settle modal; they get a warning icon with a tooltip, and a banner suggests adding margin or reducing/closing the position so realized profit covers the shortfall.


src/components/SettleAccruedFundingFeeModal/utils.ts
```diff
@@ -3,12 +3,16 @@ import { PositionInfo } from "domain/synthetics/positions";
 const PENDING_FUNDING_FEE_THRESHOLD = 10n * 10n ** 30n; // 10$
 const SETTLE_FEE_RATIO_THRESHOLD = 10n;
 
-export function shouldPreSelectPosition(
-  { pendingClaimableFundingFeesUsd, marketInfo }: PositionInfo,
-  networkFee: bigint
-) {
+export function getIsSettlementLikelyToFail({ remainingCollateralUsd }: PositionInfo) {
```
divhead, line 6: Traced this predicate end-to-end: because the settle order uses `sizeDeltaUsd: 0n`, no position/closing fee is charged, so the only collateral deduction is pending borrow + funding fees — which makes `remainingCollateralUsd < 0n` an exact match for the on-chain `InsufficientFundsToPayForCosts("funding")` revert rather than a heuristic. BigInt/30-dec/signs are correct and the all-positive case produces no warning. Looks right.

src/components/SettleAccruedFundingFeeModal/SettleAccruedFundingFeeRow.tsx
```diff
@@ -25,6 +34,9 @@ export const SettleAccruedFundingFeeRow = ({ position, isMarketDisabled, isSelec
         {position.isLong ? t`Long` : t`Short`} {indexName}
       </span>{" "}
       <span className="subtext">[{poolName}]</span>
+      {!isMarketDisabled && isSettlementLikelyToFail && (
+        <WarnIcon className="ml-4 self-center text-yellow-300" aria-label={t`Warning icon`} />
```
divhead, line 38: Nit (non-blocking): this `aria-label` may be ignored by screen readers since `WarnIcon` is an inline SVG with no `role="img"`; the adjacent tooltip already carries the meaning, so `aria-hidden` on the icon would be cleaner.

divhead (commented):
Reviewed the funding-fee settlement warning — math, signs, BigInt/30-dec, the two-step keeper semantics and lingui coverage all check out. One non-blocking a11y nit inline. LGTM.

divhead (approved):

PR #2562, opened 2026-06-11
Fix market order pending state recovery FEDEV-3976, FEDEV-3977

qwinndev:
## Summary
- Recover market-order status notifications when websocket execution events arrive before creation events or the creation event is missed.
- Backfill pending market-create notifications from trade history so executed orders can close the toast instead of loading indefinitely.
- Clear optimistic opening position mocks when a later position event covers the pending update.
- Add focused tests for order-status backfill and optimistic position merging.

## Root Cause
Market-order toasts only completed when local websocket state had a matched order with `executedTxnHash`. `OrderExecuted` events were ignored if `OrderCreated` was not already in local state, and there was no indexer backfill path for the toast. Optimistic opening positions also relied on event state without tracking whether a later position event covered the pending update.

## Linear
- [FEDEV-3976](https://linear.app/gmx-io/issue/FEDEV-3976/market-order-notifications-can-stay-loading-after-execution)
- [FEDEV-3977](https://linear.app/gmx-io/issue/FEDEV-3977/harden-optimistic-opening-position-cleanup-for-market-orders)


divhead (dismissed):

divhead:
**Review pass — market-order pending-state recovery** (T3; traced the optimistic/backfill paths — the happy-path reconciliation looks sound: absolute `SET` semantics, orderKey-scoped coverage, the block-number→timestamp fix in `useOptimisticPositions` is correct)

**⚠️ Live `OrderExecuted` recovery doesn't compose with the polling backfill → market-order toast can stick at "Fulfilling order request…"**
In the missed-`OrderCreated` race (WS drops `OrderCreated` but delivers `OrderExecuted` in the same session), the `OrderExecuted` handler writes a **data-less** status `{key, createdAt, executedTxnHash}` (`SyntheticsEventsProvider.tsx:346-350`) and — unlike the `OrderUpdated`/`OrderCancelled` handlers — does **not** drop the order from `awaitingBackfillOrders`. The polling backfill then can't repair it: `getOrderBackfillMatches` excludes any already-applied action (`orderStatusesBackfill.ts:116`), and `getIsActionAlreadyApplied` treats the executed action as applied because `executedTxnHash` is now set (`:136-140`), so `getOrderCreatedDataFromPendingOrder` (`SyntheticsEventsProvider.tsx:1287`) never runs and `status.data` is never attached. The market-order toast can only resolve its status via `status.data` (its own `orderKey` is undefined for a market create, and the data-less status is keyed by the on-chain order key, not the synthetic `getPendingOrderKey`), so it never matches → `isCompleted` stays false (`OrderStatusNotification.tsx:266`) and the toast never auto-closes. The position still shows via optimistic positions; only the notification sticks.
Fix: separate "has data" from "has txn hash" in `getIsActionAlreadyApplied` (so the backfill still attaches `data` for a create whose status only carries a txn hash), and/or have the `OrderExecuted` handler drop the matched create from `awaitingBackfillOrders`.
Reachability: edge case (missed-`OrderCreated` WS race on a market order), but squarely within this PR's recovery-hardening scope.

**💡 Market-order recovery omits `OrderFrozen`**
`getExpectedEventNames` returns `[OrderExecuted, OrderCancelled]` for market creates (`orderStatusesBackfill.ts:52-53`); `OrderFrozen` is a real `TradeActionType`, so a market order that terminates Frozen isn't recovered. This mirrors the existing live-path filter (pre-existing gap, not a regression from this PR), but it's a hole against the "recover terminal state" goal — worth confirming whether current contracts can freeze market orders.

**💡 Backfill address match uses raw `===` instead of checksum-normalizing**
`getIsRawTradeActionMatchingPendingOrder` compares `account` / `marketAddress` / `initialCollateralTokenAddress` / `swapPath` with bare `===` (`orderStatusesBackfill.ts:171-175`), whereas the sibling `createRawTradeActionTransformer` `getAddress()`-checksums the same fields. Safe today (the V2 subsquid stores checksummed addresses, same as the working `account_eq` / `marketAddress_eq` filters), so not a live bug — but fragile to any indexer/chain returning lowercase, and both specs use all-numeric addresses where checksumming is identity, so a casing regression wouldn't be caught. Consider `getAddress()`-normalizing both sides, or add one checksummed-with-letters fixture.

qwinndev:
> **⚠️ Live `OrderExecuted` recovery doesn't compose with the polling backfill → market-order toast can stick at "Fulfilling order request…"** In the missed-`OrderCreated` race (WS drops `OrderCreated` but delivers `OrderExecuted` in the same session), the `OrderExecuted` handler writes a **data-less** status `{key, createdAt, executedTxnHash}` (`SyntheticsEventsProvider.tsx:346-350`) and — unlike the `OrderUpdated`/`OrderCancelled` handlers — does **not** drop the order from `awaitingBackfillOrders`. The polling backfill then can't repair it: `getOrderBackfillMatches` excludes any already-applied action (`orderStatusesBackfill.ts:116`), and `getIsActionAlreadyApplied` treats the executed action as applied because `executedTxnHash` is now set (`:136-140`), so `getOrderCreatedDataFromPendingOrder` (`SyntheticsEventsProvider.tsx:1287`) never runs and `status.data` is never attached. The market-order toast can only resolve its status via `status.data` (its own `orderKey` is undefined for a market create, and the data-less status is keyed by the on-chain order key, not the synthetic `getPendingOrderKey`), so it never matches → `isCompleted` stays false (`OrderStatusNotification.tsx:266`) and the toast never auto-closes. The position still shows via optimistic positions; only the notification sticks. Fix: separate "has data" from "has txn hash" in `getIsActionAlreadyApplied` (so the backfill still attaches `data` for a create whose status only carries a txn hash), and/or have the `OrderExecuted` handler drop the matched create from `awaitingBackfillOrders`. Reachability: edge case (missed-`OrderCreated` WS race on a market order), but squarely within this PR's recovery-hardening scope.

Fixed

> **💡 Market-order recovery omits `OrderFrozen`** `getExpectedEventNames` returns `[OrderExecuted, OrderCancelled]` for market creates (`orderStatusesBackfill.ts:52-53`); `OrderFrozen` is a real `TradeActionType`, so a market order that terminates Frozen isn't recovered. This mirrors the existing live-path filter (pre-existing gap, not a regression from this PR), but it's a hole against the "recover terminal state" goal — worth confirming whether current contracts can freeze market orders.

We haven't handled this before, so it good to have out-of-scope for this task

> **💡 Backfill address match uses raw `===` instead of checksum-normalizing** `getIsRawTradeActionMatchingPendingOrder` compares `account` / `marketAddress` / `initialCollateralTokenAddress` / `swapPath` with bare `===` (`orderStatusesBackfill.ts:171-175`), whereas the sibling `createRawTradeActionTransformer` `getAddress()`-checksums the same fields. Safe today (the V2 subsquid stores checksummed addresses, same as the working `account_eq` / `marketAddress_eq` filters), so not a live bug — but fragile to any indexer/chain returning lowercase, and both specs use all-numeric addresses where checksumming is identity, so a casing regression wouldn't be caught. Consider `getAddress()`-normalizing both sides, or add one checksummed-with-letters fixture.

this okay, as we use checksummed in all places

divhead (approved):

PR #2563, opened 2026-06-11
Fix account position list order row FEDEV-3978

qwinndev:
## Summary
- Hide account page position-list order actions for unauthenticated viewers or when the connected account does not own the viewed account.
- Keep order edit/cancel actions available in the trade/synthetics position list.
- Explicitly suppress the final Orders section bottom border in the mobile/card position layout.

## Root Cause
The Orders card section inherited the shared `AppCardSection` bottom border. The account dashboard also used the shared position-list order row without an ownership-aware read-only gate, so unauthenticated/non-owner viewers could see edit and cancel actions.

Linear: https://linear.app/gmx-io/issue/FEDEV-3978/fix-account-position-list-order-row-visual-and-actions

### DIFF
--- a/src/components/PositionItem/PositionItemOrders.tsx
+++ b/src/components/PositionItem/PositionItemOrders.tsx
@@ -34,9 +34,11 @@ import { TwapOrderProgress } from "../OrderItem/OrderItem";
 export function PositionItemOrdersSmall({
   positionKey,
   onOrdersClick,
+  hideActions,
 }: {
   positionKey: string;
   onOrdersClick?: (key?: string) => void;
+  hideActions?: boolean;
 }) {
   const ordersWithErrors = usePositionOrdersWithErrors(positionKey);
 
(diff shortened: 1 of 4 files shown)


src/components/PositionItem/PositionItemOrders.tsx
```diff
@@ -150,7 +166,7 @@ function PositionItemOrder({
             <ChevronRightIcon className="ml-4 size-14" />
           </div>
         </Button>
-        {!isTwapOrder(order) && !isMarketOrderType(order.orderType) && (
+        {!hideEditButton && !isTwapOrder(order) && !isMarketOrderType(order.orderType) && (
```
divhead: This gates **edit** on `!hideEditButton` (derived as "viewer isn't the owner" in `HistoricalLists.tsx:105`), but the sibling **cancel** button just below (`cancelButton`, `:154-158`, rendered at `:174-178`) isn't gated by the same condition — only by `disabledCancelMarketOrderMessage`. It's wired to `useCancelOrder(order)` directly (`:138`), not the `onCancelOrder` prop, so the `onCancelOrder={noop}` the dashboard passes (`HistoricalLists.tsx:147`) doesn't neutralize it either. Net: on a foreign account a user can still click cancel and fire a transaction that can only revert on-chain — the same mismatched-viewer UX this PR fixes for edit. Suggest gating `cancelButton` on `!hideEditButton` too, or renaming the flag to `hideOrderActions`/`readOnly` and hiding both. If keeping cancel visible on foreign accounts is intentional, could you confirm?

divhead (commented):
The fix correctly hides the edit button for non-owners, but the sibling cancel button is left live — see inline. The mobile border fix (`!border-b-0`) and the `isShareAvailable`→`isPositionOwner` rename look good.

divhead (changes requested):

divhead (approved):

PR #2581, opened 2026-06-24
Warn when limit/stop increase may be liquidatable at trigger price FEDEV-2399

qwinndev:
Warns users when a **Limit Increase / Stop Increase** order that increases an existing position would be liquidatable at its **trigger (execution) price**, so they can add collateral or reduce leverage before it triggers.

Linear: https://linear.app/gmx-io/issue/FEDEV-2399

### Behavior
- **Warn (non-blocking)** for a resting Limit/Stop Increase on an existing position when the simulated resulting position would be liquidatable at the trigger price — submit/update stays enabled. Shown in the tradebox and the order editor (and via the editor that a chart-line drag opens).
- **Block (unchanged)** remains the existing Market Increase guard (liquidatable at the mark price).
- The check compares the next-position liquidation price against the **trigger price** (execution price) rather than the current mark price.

### Changes
- `getIsPositionLiquidatableAtPrice` shared helper (+ unit tests).
- Tradebox increase liquidation guard scoped to Market Increase only (`!isLimit`); new `selectTradeboxIncreaseLiquidationRiskWarning` selector.
- Shared `LiquidatableIncreaseWarningCard`, reused in the tradebox and the order editor.
- Order editor computes the warning inline (non-blocking); the chart-line drag path is unchanged, so the warning surfaces in the editor it opens.
- Warning copy translated across all locales.

### Out of scope
- Dynamic volatility / near-liquidation distance buffers (FEDEV-3906).
- Failed-order `LiquidatablePosition` messaging and the Market Increase block copy (already implemented).

### DIFF
--- a/src/components/OrderEditor/OrderEditor.tsx
+++ b/src/components/OrderEditor/OrderEditor.tsx
@@ -80,6 +80,7 @@ import {
 } from "domain/synthetics/trade";
 import { useCloseSizeInput } from "domain/synthetics/trade/useCloseSizeInput";
 import { getExpressError, getIsMaxLeverageExceeded } from "domain/synthetics/trade/utils/validation";
+import { getIsPositionLiquidatableAtPrice } from "domain/synthetics/trade/utils/warnings";
 import { TokensRatioAndSlippage } from "domain/tokens";
 import {
   FULL_POSITION_CLOSE_SIZE_DELTA_USD,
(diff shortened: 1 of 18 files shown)


qwinndev:
Superseded by #2582 (branch renamed to fedev-2399-liquidatable-increase-warning for readability).

PR #2583, opened 2026-06-24
Warn when TP/SL prices are beyond current liquidation price FEDEV-4002

qwinndev:
## Summary

Adds a **non-blocking** warning across all TP/SL create and edit surfaces when a trigger price is at or beyond the position's current liquidation price. Users keep full freedom to create or update the order — the UI just makes the liquidation-price risk explicit instead of silently presenting a normal protective exit.

For long positions the warning shows when the trigger is at/below the liquidation price; for shorts, at/above. It works for both manual price entry and gain/loss-derived entry, and always accompanies the Est. PnL / receive / next-position previews.

## Changes

- New shared `getTpSlLiqPriceWarning` helper (`src/domain/tpsl/utils.ts`) — single source of truth for the condition + copy.
- `handleEntryError` derives a non-blocking `EntryField.warning` via the helper (never `error`), surfaced in the tradebox sidecar and the Add TP/SL modal through `TPSLInputRow` (yellow border + inline alert). Submit gating still keys off `error` only, so warnings never block.
- **Order Editor:** removed the beyond-liq hard block from `getPositionOrderError`; the editor now shows the same non-blocking warning and lets you submit.
- Regression coverage for the shared helper and the `handleEntryError` warning path.
- Translations added for the new string in all locales.

[FEDEV-4002](https://linear.app/gmx-io/issue/FEDEV-4002/bug-warn-when-tpsl-prices-are-beyond-current-liquidation-price)


src/components/OrdersModal/TPSLInputRow.tsx
```diff
@@ -340,7 +363,26 @@ export function TPSLInputRow({
...
-  const estimatedPnl = estimatedPnlProp ?? derivedEstimatedPnl;
+  const estimatedPnl = useMemo(() => {
+    const base = estimatedPnlProp ?? derivedEstimatedPnl;
+    if (!base) return base;
+
+    const pnlCollateralUsd = getCollateralUsdForPnl(currentPriceValue);
+    const cappedPnlUsd = getCappedTpSlLossUsd({
+      pnlUsd: base.pnlUsd,
+      collateralUsd: pnlCollateralUsd,
+      triggerPrice: currentPriceValue,
```
timgmxio, line 375: seems it doesn't account for `visualMultiplier`, the `currentPriceValue` is already multiplied by `visualMultiplier`, but `liquidationPrice` is raw price

Image

src/components/OrdersModal/AddTPSLModal.tsx
```diff
@@ -518,6 +522,27 @@ export function AddTPSLModal({
...
+    liquidationPrice: position.liquidationPrice,
+    isLong,
+  });
+  const activeEstimatedPnlPercentage = activeDecreaseAmounts
+    ? activeIsBeyondLiquidation
+      ? -BASIS_POINTS_DIVISOR_BIGINT
+      : activeDecreaseAmounts.estimatedPnlPercentage
+    : undefined;
+  const activeEstimatedPnl = activeDecreaseAmounts
+    ? activeIsBeyondLiquidation && activeDecreaseAmounts.estimatedPnlPercentage !== 0n
```
timgmxio: These are two forms of the same value set independently, and the logic visibly desyncs them: the percentage is hard-forced to −100%, while the USD is back-derived from the rounded `estimatedPnlPercentage`. So when the trigger is beyond liq but PnL at that price rounds to 0 bps (e.g. a long whose liq sits above entry due to fees, trigger near entry), the USD keeps raw `estimatedPnl` and the row shows "+$0.12 (−100.00%)".

maybe it's worth doing something like this:
```typescript
const { activeEstimatedPnl, activeEstimatedPnlPercentage } = useMemo(() => {
  if (!activeDecreaseAmounts) return {};

  const showFullLoss = activeIsBeyondLiquidation && activeDecreaseAmounts.estimatedPnlPercentage !== 0n;
  if (!showFullLoss) {
    return {
      activeEstimatedPnl: activeDecreaseAmounts.estimatedPnl,
      activeEstimatedPnlPercentage: activeDecreaseAmounts.estimatedPnlPercentage,
    };
  }

  return {
    activeEstimatedPnl: bigMath.mulDiv(
      activeDecreaseAmounts.estimatedPnl,
      -BASIS_POINTS_DIVISOR_BIGINT,
      activeDecreaseAmounts.estimatedPnlPercentage
    ),
    activeEstimatedPnlPercentage: -BASIS_POINTS_DIVISOR_BIGINT,
  };
}, [activeDecreaseAmounts, activeIsBeyondLiquidation]);
```

timgmxio (changes requested):

divhead:
**Review pass — TP/SL beyond-liq warning** (the existing review already flags the `visualMultiplier` mismatch and the PnL/percentage desync; adding two more)

**⚠️ Beyond-liq trigger is still a hard block in Close/Trigger + tradebox-Trigger, while this PR softens it to a warning everywhere else**
The PR removes the beyond-liq block from `getPositionOrderError.tsx` (replaced by the "surfaces as a non-blocking warning" comment at `getPositionOrderError.tsx:94`) and adds warnings in AddTPSLModal / OrderEditor / tradebox sidecar. But the byte-identical check still hard-blocks in `getDecreaseError` — `validation.ts:595-600` (`Set trigger price above/below liquidation price`) — which feeds PositionSeller's Close/Trigger tab (`PositionSeller.tsx:502`) and the tradebox Trigger tab (`selectTradeboxTradeErrors.ts:142`). Net: a long SL below liq (or short SL above liq) is **blocked** when set via Close/Trigger but only **warned** when set via Add TP/SL — same action, opposite behavior, which partially defeats the PR's goal.
Fix: route `validation.ts:594-600` through the same non-blocking `getTpSlLiqPriceWarning` path, or confirm Close/Trigger is intentionally out of scope.

**Scope note on the `visualMultiplier` mismatch (issue #1):** the same `currentPriceValue` (×`visualMultiplier`) vs raw `liquidationPrice` pairing appears at three sites in `TPSLInputRow.tsx` — the `formatGainLossValue` branches at `:234` and `:245`, and the `estimatedPnl` memo at `:374` — not only line 374. A fix at 374 alone leaves the Loss/Gain field wrong on visualMultiplier markets (kPEPE etc.); please confirm all three are corrected.

qwinndev:
> **⚠️ Beyond-liq trigger is still a hard block in Close/Trigger + tradebox-Trigger, while this PR softens it to a warning everywhere else** The PR removes the beyond-liq block from `getPositionOrderError.tsx` (replaced by the "surfaces as a non-blocking warning" comment at `getPositionOrderError.tsx:94`) and adds warnings in AddTPSLModal / OrderEditor / tradebox sidecar. But the byte-identical check still hard-blocks in `getDecreaseError` — `validation.ts:595-600` (`Set trigger price above/below liquidation price`) — which feeds PositionSeller's Close/Trigger tab (`PositionSeller.tsx:502`) and the tradebox Trigger tab (`selectTradeboxTradeErrors.ts:142`). Net: a long SL below liq (or short SL above liq) is **blocked** when set via Close/Trigger but only **warned** when set via Add TP/SL — same action, opposite behavior, which partially defeats the PR's goal. Fix: route `validation.ts:594-600` through the same non-blocking `getTpSlLiqPriceWarning` path, or confirm Close/Trigger is intentionally out of scope.

not valid since Trigger is no more part of PositionSeller or TradeBox

src/components/OrdersModal/TPSLInputRow.tsx
qwinndev: Good catch — fixed in 8712629a. Rather than patch each call site, I made `currentPriceValue` the raw price at the source (`parseValue(priceValue) / visualMultiplier`), since `getPnlUsdForPrice`, `getCollateralUsdForPnl` and `isTriggerBeyondLiquidation` all expect raw. That corrects all three `getCappedTpSlLossUsd` sites plus the underlying PnL/collateral calc on k-tokens, and matches how `handleEntryError` derives the raw entry value (`parseValue(input)/multiplier`), so the cap and the warning agree on 'beyond liq'.

src/components/OrdersModal/AddTPSLModal.tsx
qwinndev: Fixed in 8712629a — went with your single-`useMemo` shape so the USD and percentage are always derived together: force −100% only when it can back-derive the USD from a non-zero `estimatedPnlPercentage`, otherwise keep both raw. So the 'PnL rounds to 0 bps' case now shows a consistent raw value instead of '+$0.12 (−100%)'. Thanks!

timgmxio (approved):

PR #2595, opened 2026-06-26
Wallet receive/send and GMX Account funding fedev-3882

timgmxio:


src/components/GmxAccountModal/MainView.tsx
```diff
@@ -1,393 +1,436 @@
...
+  const [isCopied, setIsCopied] = useState(false);
+  const copyTimeoutRef = useRef<number | undefined>(undefined);
   const handleDisconnect = useDisconnectAndClose();
+  const { walletUsd } = useAvailableToTradeAssetSettlementChain();
+  const { wallets } = useWallets();
+  const { exportWallet } = useExportWallet();
+
+  const embeddedWallet = getEmbeddedConnectedWallet(wallets);
+  const canExport = Boolean(
+    embeddedWallet && account && embeddedWallet.address.toLowerCase() === account.toLowerCase()
```
qwinndev: i do not think we need address lower case here

src/components/GmxAccountModal/getAccountModalMode.ts
```diff
@@ -0,0 +1,23 @@
+import type { ContractsChainId, SourceChainId } from "config/chains";
+import { isValidVisualSettlementChain } from "config/multichain";
+
+export type AccountModalMode = "walletAndGmxAccount" | "gmxAccount" | "walletOnly";
+
+export function getAccountModalMode(
```
qwinndev: pure chain-config logic in a component folder, lets move to config/multichain.ts

src/components/GmxAccountModal/getAccountModalMode.ts
```diff
@@ -0,0 +1,23 @@
...
+  }
+
+  if (srcChainId !== undefined) {
+    return "gmxAccount";
+  }
+
+  return "walletAndGmxAccount";
+}
+
+export function isWalletOnlyChain(chainId: number): boolean {
```
qwinndev: do we have any reason to have it instead of just use `!isValidVisualSettlementChain(chainId)` directly?

src/components/GmxAccountModal/WalletSendView.tsx
```diff
@@ -0,0 +1,646 @@
...
+  const walletBalanceUsd =
+    selectedToken && walletBalance !== undefined
+      ? convertToUsd(walletBalance, selectedToken.decimals, selectedToken.prices?.maxPrice)
+      : undefined;
+  const nextWalletBalanceUsd =
+    walletBalanceUsd !== undefined && amountUsd !== undefined ? walletBalanceUsd - amountUsd : undefined;
+
+  const isRecipientValid = isAddress(recipient);
+  const isInputEmpty = amount === undefined || amount <= 0n;
+  const isInsufficientBalance = walletBalance !== undefined && amount !== undefined && amount > walletBalance;
```
qwinndev: should we check balance including gas reserve?

src/context/GmxAccountContext/GmxAccountContext.tsx
```diff
@@ -50,6 +55,11 @@ export type GmxAccountContext = {
 
   selectedTransferGuid: string | undefined;
   setSelectedTransferGuid: React.Dispatch<React.SetStateAction<string | undefined>>;
+
+  // available assets view
```
qwinndev, line 67: not sure that we use such separtion anywhere else

src/locales/zh-tw/messages.po
```diff
@@ -465,6 +466,10 @@ msgstr "GMX Account 餘額"
 msgid "Search in perpetuals markets"
 msgstr "搜尋永續合約市場"
 
+#: src/components/GmxAccountModal/WalletReceiveView.tsx
+msgid "Address copied!"
+msgstr ""
```
qwinndev: lets ask llm to add translations for new strings

qwinndev (dismissed):

qwinndev (dismissed):

qwinndev:
For a same-chain native send (e.g. send ETH → another address on Arbitrum), gasPaymentTokenAmount is passed as undefined (the ternary only supplies the fee when !isSameChain). Because the send token is the gas token, the full-balance early-return is skipped; because effectiveGasPaymentTokenAmount = 0n, the minimal-buffer fallback (> 0n guard) is also skipped. It falls through to the $20 min residual — the reserve intended for GMX order execution, not a ~21k-gas transfer.

Balance < ~$20 of ETH: maxAvailableAmount = 0n → any typed amount trips isInsufficientBalance → button locks to “Insufficient balance.” User cannot send at all.
Larger balance: “Max” strands ~$20–$40 of ETH; the user can never send more than balance − $20.
The cross-chain native path is fine — it passes the real LayerZero fee (> 0n), so it correctly reserves only 1.4× the fee. The bug is the same-chain native branch specifically.

Fix: for a plain native same-chain send, reserve the actual send gas (≈21k × gasPrice) instead of the order-execution residual — e.g. pass that estimate as gasPaymentTokenAmount/fallbackGasPaymentTokenAmount so the minimal-buffer path applies, mirroring the cross-chain branch.

qwinndev (dismissed):

qwinndev (dismissed):

qwinndev (approved):

PR #2597, opened 2026-06-26
Market delistings: combined exit announcement FEDEV-4022

qwinndev:
Adds a combined market-delisting exit announcement: affected users see warning toasts prompting them to close positions in delisting markets and/or withdraw direct (non-GLV) GM liquidity from frozen pools.

## What it does

- Curated hardcoded per-chain delisting list (`DELISTING_MARKET_ADDRESSES`: 12 Arbitrum + 2 Avalanche markets, including the DAI swap pools) with an `isDelistingMarket` helper.
- Computes the user's exposure inside `SyntheticsStateContext`, reusing already-fetched positions + GM balances — **no new network requests**. The toast renders via the app-root Toaster and persists across navigation.
- Two independent "Market delistings" warning toasts: one for positions (markets to close), one for liquidity (pools to withdraw, with a "Manage liquidity" link to `/pools`). GLV depositors are excluded automatically (detection keys off GM token balances).
- Dismissal: 24h cooldown that re-shows immediately when a new market/pool enters the user's affected set (not permanently dismissed).
- `EventToast` now renders an internal `link.href` via a react-router `<Link>` so the `/pools` link navigates client-side instead of full-reloading to the wrong page under the app's `HashRouter` (no behavior change for existing external links).

## Notes / deviations from the issue's written acceptance criteria (per implementation decisions)

- Two separate toasts (positions + liquidity) instead of one combined announcement.
- Targeting uses a curated hardcoded list (Option B in the issue), not live OI-cap detection.

https://linear.app/gmx-io/issue/FEDEV-4022


divhead (dismissed):

divhead:
**Review pass — delisting exit announcements**

**⚠️ Duplicate banners when two `SyntheticsStateContextProvider`s are mounted at once**
`DelistingExitAnnouncements` is rendered by every provider instance (`SyntheticsStateContextProvider.tsx:512`) and each `createPortal`s into the single shared `#delisting-announcements-slot` (`DelistingExitAnnouncements.tsx:52`; slot defined in `WhatsNewToastContainer.tsx:64`) with no singleton guard. The only render guard is `pageType === "accounts"`, so any other pageType passes. When two providers are mounted simultaneously both portal into the same slot and the combined banner renders twice — reachable via the GMX Account modal (`pageType="gmxAccount"`, `GmxAccountModalShared.tsx:75`) opened over any page, or the nested `pageType="referrals"` providers (`AddAffiliateCode.tsx:580`, `CreateReferralCode.tsx:61`). Both instances load the same connected-account positions / GM balances (those hooks aren't pageType-gated), so both produce the same non-empty `toShow`.
Fix: hoist the announcement to the shell (next to `WhatsNewToastContainer`) or add a mounted-singleton guard so only one instance portals.

**⚠️ User-facing copy is hardcoded English, bypassing lingui**
Titles, body text and link text are plain strings — `delistingExitAnnouncementsLogic.ts:115` (`"Market delistings"`), `:174` (`"Manage liquidity"`), plus hand-rolled pluralization at `:32` (`positionCount === 1 ? "position" : "positions"`). The sibling toasts it renders alongside (`WhatsNewToast`, `uiFlagEvents`) use `<Trans>`, so non-English users get English banners and the pluralization can't localize. Use `t` / `<Trans>` / `<Plural>` with market-name interpolation. (If this is a deliberate launch-speed shortcut, feel free to downgrade.)

**💡 `toDismiss` is computed + unit-tested but never consumed**
`getDelistingAnnouncementActions` returns `{ toShow, toDismiss }` (`delistingExitAnnouncementsLogic.ts:179`), but the only caller destructures `{ toShow }` (`DelistingExitAnnouncements.tsx:38`). `toDismiss` (and its tests) is dead output. If it was meant to clear stale localStorage dismissals when exposure disappears (so re-entering a delisted market re-warns), that isn't wired; otherwise drop it.

divhead (dismissed):

divhead (approved):

PR #2616, opened 2026-07-02
Use block-time minCollateralFactorForLiquidation for historical liquidation UI FEDEV-4003

qwinndev:
## Summary

Historical liquidation rows/tooltips in Trade History reconstructed **Minimum required margin** and the max-leverage context from `tradeAction.marketInfo.minCollateralFactorForLiquidation` — the *current* market config. For markets that switch risk parameters (e.g. RWA/commodity markets toggling off-hours vs on-hours), this renders a misleading historical value.

This change reads the **block-time** `minCollateralFactorForLiquidation` now indexed on `TradeAction` (Subsquid slot `6230c4`) and uses it for the liquidation tooltip, so a past liquidation is explained with the config that applied at its block. It falls back to the current market config only when the historical value is unavailable (old data).

Example (WTIOIL/USD): renders the historical off-hours `1%` factor → `$9,526.02` minimum required margin instead of the later on-hours `0.5%` → `$4,763.01`.

## Changes

- **Subsquid**: point all chains except Fuji at slot `6230c4` (which indexes `TradeAction.minCollateralFactorForLiquidation`); bump the codegen source URL.
- **SDK**: add `minCollateralFactorForLiquidation` to the generated `TradeAction` type, both trade-history GraphQL queries, the raw→domain transformer, and `PositionTradeAction`.
- **UI**: `formatPositionMessage` liquidation branch uses the event-time factor (with fallback) for max leverage and minimum required margin.
- **Tests**: regression coverage for a liquidation whose factor changes after the execution block, plus the missing-value fallback.

Linear: https://linear.app/gmx-io/issue/FEDEV-4003/bug-use-block-time-market-config-for-historical-interface-calculations

### DIFF
--- a/src/components/TradeHistory/TradeHistoryRow/utils/position.ts
+++ b/src/components/TradeHistory/TradeHistoryRow/utils/position.ts
@@ -538,10 +538,13 @@ export const formatPositionMessage = (
     //#endregion StopLossDecrease
     //#region Liquidation
   } else if (ot === OrderType.Liquidation && ev === TradeActionType.OrderExecuted) {
-    const maxLeverage =
-      tradeAction.marketInfo.minCollateralFactorForLiquidation === 0n
-        ? 0n
-        : PRECISION / tradeAction.marketInfo.minCollateralFactorForLiquidation;
+    // Use the block-time minCollateralFactorForLiquidation captured on the trade action so historical
+    // liquidation context is not reconstructed with the current (possibly changed) market config.
+    // Falls back to the current market config only when the historical value is unavailable (old data).
+    const minCollateralFactorForLiquidation =
+      tradeAction.minCollateralFactorForLiquidation ?? tradeAction.marketInfo.minCollateralFactorForLiquidation;
+
+    const maxLeverage = minCollateralFactorForLiquidation === 0n ? 0n : PRECISION / minCollateralFactorForLiquidation;
     const formattedMaxLeverage = Number(maxLeverage).toFixed(1) + "x";
 
     const initialCollateralUsd = convertToUsd(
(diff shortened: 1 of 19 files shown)


src/components/TradeHistory/TradeHistoryRow/utils/position.ts
```diff
@@ -538,10 +538,10 @@ export const formatPositionMessage = (
     //#endregion StopLossDecrease
     //#region Liquidation
   } else if (ot === OrderType.Liquidation && ev === TradeActionType.OrderExecuted) {
-    const maxLeverage =
-      tradeAction.marketInfo.minCollateralFactorForLiquidation === 0n
-        ? 0n
-        : PRECISION / tradeAction.marketInfo.minCollateralFactorForLiquidation;
+    const minCollateralFactorForLiquidation =
+      tradeAction.minCollateralFactorForLiquidation ?? tradeAction.marketInfo.minCollateralFactorForLiquidation;
```
timgmxio: on @6230c4 the indexer returns the string "0" (not null) for every liquidation that pre-dates the on-chain config — checked live: 2023–2025 liquidations all come back as "0", real values only start around H2 2025. "0" is truthy, so the transformer yields `0n`, and `0n ?? fallback` never falls back. Net effect for ~two years of history: `maxLeverage = 0n` → "max leverage of 0.0x", and `applyFactor(size, 0n)` clamps the tooltip to "Minimum required margin: $1.00" — a worse regression than the bug we're fixing.

`||` instead of `??` fixes it (a zero factor can only mean the config didn't exist at that block). The fallback test currently models `undefined` only — that's why it passes; worth extending it to cover `0n` too.

timgmxio: Separate from the "0" issue: the DoD says that when the block-time factor can't be resolved, we shouldn't silently present current config as historical truth — and this `??` fallback is exactly that, with nothing in the tooltip hinting the value isn't historical. Once the "0" handling is fixed, every pre-H2-2025 liquidation takes this path, so the misleading scenario from the issue quietly lives on for two years of history (where the threshold was governed by `minCollateralFactor` anyway, so today's `minCollateralFactorForLiquidation` isn't even the right knob).

let's hide the factor-derived rows when the block-time value is missing — the conditional-`undefined` pattern in `lines()` (like the OrderFrozen branch) handles that cleanly, with a generic "max allowed leverage" phrasing for the headline. If we'd rather keep the fallback, fine, but that's a product call to agree with xhiroz and record as a DoD deviation.

timgmxio (approved):

PR #2618, opened 2026-07-06
Add TradeBox component tests with prod-like mock infrastructure FEDEV-4024

qwinndev:
## What

Linear: [FEDEV-4024](https://linear.app/gmx-io/issue/FEDEV-4024/mocks-for-integration-tests)

54 Playwright component tests for the trade widget (`TradeBoxHeaderTabs` + `TradeBox`), running against the production provider stack with mocked chain data.

## Mock infrastructure

- **`MockSyntheticsStateProvider`** — assembles `SyntheticsState` from the real stateful hooks (`useTradeboxState`, settings, order/position editor states, permits, subaccount) over static fixtures, so every derived selector (`increaseAmounts`, fees, leverage, swap path) computes through the production pipeline.
- **Fixtures**: ETH/USD market (+ second ETH-ETH pool, capped-OI variant), USDC/ETH/native tokens, 2x long position, gas limits/price, positions constants.
- **wagmi mock connector** (`connected` story prop) unlocks submit-button validation states.
- **SyntheticsEvents** replaced with an idle no-events stub via vite alias in `playwright-ct.config.ts` (websocket layer; documented limitation: txn/approval event flows are not observable in CT).
- **Hermetic network**: all non-localhost http(s) aborted per test; localStorage seeding for stored options/leverage/denomination scenarios.

## Covered

- Field interactions in slider mode (margin ↔ size, percentage slider, token/USD display modes) and manual leverage mode (independent inputs, fixed-percentage slider, max-leverage validation)
- Limit/Stop flows: trigger price participation in sizing, mark-price fallback, price resets, direction validations
- Swap mode (real swap path, switch ball, min receive, ratio input) and wrap 1:1; TWAP rows and parts bounds
- Submit validations (enter amount/price, balances, min margin/position size, OI cap), TP/SL entries, pool switch + pool/collateral warnings, high swap impact card
- Seeded scenarios: native ETH gas buffer on max, leverage clamp to market cap (95x), persisted size denomination, Trigger-mode fallback invariant


timgmxio (commented):
Local review for FEDEV-4024. No criticals; CI is green on `8b1d14a` including Playwright CT.

The infrastructure itself is solid: fixtures use real token/market addresses from the sdk configs with consistent 30-decimals pricing, and state is assembled from the production hooks, so the 54 tests exercise the real selector pipeline rather than the mocks. Isolation and determinism check out (fresh page/localStorage per test, hermetic network, static prices).

2 warnings + 2 suggestions inline. The main ones: the WETH-WETH fixture inherits USDC-decimals short-pool amounts, and the `SyntheticsEvents` stub lives outside typecheck, so nothing guards it against drifting from `SyntheticsEventsContextType`.

playwright/mocks/syntheticsEventsProvider.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ * Replaces the websocket-driven SyntheticsEventsProvider in CT builds with its idle
+ * state (no pending events). All setters are noops, so transaction/approval event
+ * flows are not observable in CT; inject approvalStatuses here to test approve states.
+ */
+
+const noop = () => undefined;
+
+const emptySyntheticsEvents = {
```
timgmxio: **[Warning]** `emptySyntheticsEvents` is an untyped literal and `useSyntheticsEvents()` has no return type, while `playwright/` is not covered by tsconfig (`"include": ["src", "scripts", "viem-override.d.ts"]`) — so neither `tsc` in CI nor the IDE checks this file against the real API. Today the object matches `SyntheticsEventsContextType` field-for-field, but this context is an actively evolving surface (`pendingExpressTxns`, `gelatoTaskStatuses`, the whole `MultichainEventsState` were added recently). When a new field lands, components under CT will silently get `undefined` — tests either start failing mysteriously or, worse, keep passing while diverging from real behavior.

Suggestion: move the typed object into `src/domain/testUtils` (covered by `tsc` in CI) and keep this file as a thin re-export — the `domain` alias is already set up in `ctViteConfig.resolve.alias`:

```ts
// src/domain/testUtils/emptySyntheticsEventsState.ts
import type { SyntheticsEventsContextType } from "context/SyntheticsEvents/types";

const noop = () => undefined;

export const emptySyntheticsEventsState: SyntheticsEventsContextType = {
  // ...same object as here, moved unchanged
};
```

Then adding a field to `SyntheticsEventsContextType` breaks `yarn typecheck` instead of breaking tests at runtime.

src/components/TradeBox/__tests__/TradeBox.ct.stories.tsx
```diff
@@ -0,0 +1,298 @@
...
+  const marketsInfoData = useMemo<MarketsInfoData | undefined>(() => {
+    if (marketScenario === "cappedLongOI") {
+      const data: MarketsInfoData = {
+        [MOCK_MARKET_ADDRESS]: createMockMarketInfo(undefined, { maxOpenInterestLong: expandDecimals(1500, 30) }),
+      };
+      return data;
+    }
+
+    if (withSecondEthPool) {
+      const secondPool = createMockMarketInfo(undefined, {
```
timgmxio, line 179: **[Warning]** `withSecondEthPool` overrides `shortToken` to `ETH_TOKEN` (18 decimals), but `shortPoolAmount`/`maxShortPoolAmount` are inherited from the base fixture where the short token is USDC (6 decimals): `expandDecimals(2000000, 6)` = `2e12` wei of WETH ≈ 0.000002 WETH ≈ $0.004 — the short side of this prod-like pool has effectively zero liquidity. Current tests long through it, so they pass; the first short/reserves test through this pool will hit "insufficient liquidity" with a confusing cause.

The fee invariant is also inverted: the second pool ends up with `positionFeeFactorForBalanceWasNotImproved` (2e25) below the inherited `positionFeeFactorForBalanceWasImproved` (5e25), while on real markets "not improved" is always ≥ "improved".

```diff
       const secondPool = createMockMarketInfo(undefined, {
         marketTokenAddress: SECOND_ETH_MARKET_ADDRESS,
         shortTokenAddress: ETH_ADDRESS,
         shortToken: ETH_TOKEN,
         isSameCollaterals: true,
         name: "ETH/USD [WETH-WETH]",
+        shortPoolAmount: expandDecimals(1000, 18),
+        maxShortPoolAmount: expandDecimals(10000, 18),
         // cheaper than the default pool
+        positionFeeFactorForBalanceWasImproved: expandDecimals(1, 25),
         positionFeeFactorForBalanceWasNotImproved: expandDecimals(2, 25),
         positionImpactFactorNegative: 0n,
         swapImpactFactorNegative: 0n,
       });
```

src/components/TradeBox/__tests__/TradeBox.ct.stories.tsx
```diff
@@ -0,0 +1,298 @@
...
+      collateralToken: withWethCollateralPosition ? ETH_TOKEN : undefined,
+    });
+    return { [position.key]: position };
+  }, [withPosition, withWethCollateralPosition]);
+
+  const seedEntries = useMemo(() => {
+    const entries: Array<[string, string]> = [];
+
+    if (seedPayNativeEth || seedSwapWrap || seedTradeMode !== undefined) {
+      const storedOptions = {
```
timgmxio: **[Suggestion]** The blob seeded under `getSyntheticsTradeOptionsKey` is an untyped literal, and `seedTradeMode` is declared as `string` (the spec passes `"Trigger"`). Today the shape matches `StoredTradeOptions` from `useTradeboxState` (including the required `isFromTokenGmxAccount`), but the persisted-options schema has evolved before and will again — without the type, the seed silently goes stale, `useTradeboxState` falls back to defaults, and seeded scenarios start testing something else. `"Trigger"` also only works because the enum value happens to equal its name.

```diff
+import type { StoredTradeOptions } from "domain/synthetics/trade/useTradeboxState";
 ...
-  seedTradeMode?: string;
+  seedTradeMode?: TradeMode;
 ...
-      const storedOptions = {
+      const storedOptions: StoredTradeOptions = {
```

(and in the spec: `seedTradeMode={TradeMode.Trigger}` instead of `seedTradeMode="Trigger"`)

src/domain/testUtils/MockSyntheticsStateProvider.tsx
```diff
@@ -0,0 +1,189 @@
...
+          error: undefined,
+        },
+        ordersInfo: { ordersInfoData, count: Object.keys(ordersInfoData).length, isLoading: false },
+        positionsConstants: MOCK_POSITIONS_CONSTANTS,
+        uiFeeFactor,
+        userReferralInfo: undefined,
+        depositMarketTokensData: undefined,
+        progressiveDepositMarketTokensData: undefined,
+        multichainMarketTokensBalancesResult: { tokenBalances: {}, isLoading: false },
+        glvInfo: { glvData: undefined, isLoading: false } as SyntheticsState["globals"]["glvInfo"],
```
timgmxio: **[Suggestion]** These `as` casts are redundant and hide missing required fields: `glvInfo` lacks `glvs` (the real `useGlvMarketsInfo` returns `{ glvs, glvData, isLoading }`), and `jitLiquidityData: {} as ...` lacks `jitLiquidityMap`. Full literals typecheck without the casts, and future type changes get caught by `tsc` instead of being swallowed:

```diff
-        glvInfo: { glvData: undefined, isLoading: false } as SyntheticsState["globals"]["glvInfo"],
+        glvInfo: { glvs: undefined, glvData: undefined, isLoading: false },
 ...
-        jitLiquidityData: {} as SyntheticsState["globals"]["jitLiquidityData"],
+        jitLiquidityData: { jitLiquidityMap: undefined },
```

(the `leaderboard: {} as ...` cast is fine — page-scoped state, already commented)

timgmxio (approved):

PR #2647, opened 2026-07-16
Add JIT-aware trading capacity helper FEDEV-3984

qwinndev:
## Summary

- Add the public `GmxApiSdk.getTradingCapacity` helper backed by `/v1/markets/trading-capacity`.
- Add capacity types plus typed prepare estimates, validation warnings, and prepare-error parsing.
- Document global JIT-aware capacity for market, limit, stop-market, and TWAP increases, plus native `baseAvailableLiquidity` without JIT.
- Keep the existing raw JIT helper unchanged; no keeper changes are required.
- Companion API PR: https://github.com/gmx-io/gmx-api/pull/105
- Linear: https://linear.app/gmx-io/issue/FEDEV-3984/apisdk-expose-jit-aware-trading-capacity-and-prepare-validation

## Testing

- SDK: 39 files, 504 passed
- Interface unit: 103 files passed, 2 skipped; 951 tests passed, 14 skipped
- Interface component: 155 passed
- SDK/interface lint, typecheck, builds, Prettier, and diff checks
- Independent final review found no remaining issues


sdk/src/utils/markets/api.ts
```diff
@@ -32,12 +40,29 @@ export async function fetchApiMarkets(ctx: { api: IHttp }): Promise<MarketWithTi
...
   });
-  return tickers.map((t) => deserializeBigIntsInObject(t, { handleInts: true })) as MarketTicker[];
+  return tickers.map((t) => deserializeBigIntsInObject(t, { handleInts: true })) as MarketTickerWithCapacity[];
+}
+
+export function parseTradingCapacity(raw: any): TradingCapacity {
+  return deserializeBigIntsInObject(raw, { handleInts: true }) as TradingCapacity;
+}
+
+export async function fetchApiTradingCapacity(
```
divhead, line 184: Blocker: this endpoint does not exist in production yet — `GET /v1/markets/trading-capacity` returns 404 on `arbitrum.gmxapi.io`, and the companion `gmx-io/gmx-api#105` is still open.

Publishing the SDK before that deploys ships a documented, README-headline method that fails for every integrator. Please gate the release on the API rollout (deploy gmx-api#105 first, then publish), or hold this PR until it lands.

sdk/src/utils/markets/api.ts
```diff
@@ -32,12 +40,29 @@ export async function fetchApiMarkets(ctx: { api: IHttp }): Promise<MarketWithTi
...
       addresses: params?.addresses,
       symbols: params?.symbols,
     },
   });
-  return tickers.map((t) => deserializeBigIntsInObject(t, { handleInts: true })) as MarketTicker[];
+  return tickers.map((t) => deserializeBigIntsInObject(t, { handleInts: true })) as MarketTickerWithCapacity[];
+}
+
+export function parseTradingCapacity(raw: any): TradingCapacity {
+  return deserializeBigIntsInObject(raw, { handleInts: true }) as TradingCapacity;
```
divhead: This is an unchecked cast: whatever the API returns is asserted to be a `TradingCapacity`. If a field is omitted or renamed, the integrator gets `undefined` typed as `bigint` — including in the sizing expression the README recommends (`capacity.availableLiquidity`), which then either throws on arithmetic or silently produces a wrong order size.

The strict validator already exists a few files over (`isParsedTradingCapacity` in `utils/orderTransactions/api.ts`). Since `getTradingCapacity()` is the documented entry point for sizing orders, it should be at least as strict as the prepare-warning path — reuse it here and fail loudly on a malformed payload.

sdk/src/utils/orderTransactions/api.ts
```diff
@@ -163,11 +338,71 @@ function parseEstimates(raw: any): OrderEstimates | undefined {
...
+    tradingCapacity,
+  };
+}
+
+function parseValidationWarning(raw: any): OrderValidationWarning {
+  if (raw?.code === "INSUFFICIENT_LIQUIDITY") {
+    const capacity = parseTradingCapacity(raw.details);
+    const requestedSizeUsd = parseUint256DecimalString(raw.details?.requestedSizeUsd);
+    if (!isParsedTradingCapacity(capacity) || requestedSizeUsd === undefined || typeof raw.message !== "string") {
+      throw new Error("Invalid insufficient-liquidity warning in prepare response");
```
divhead: Throwing here fails the entire `prepareOrder` call, because `parsePrepareResponse` runs as the response `transform` — the test `rejects malformed capacity warning details` confirms it.

A `validationWarning` is advisory; the prepared payload next to it is perfectly valid. Since gmx-api deploys independently of the SDK, a single shape drift in an advisory field takes integrators' order flow to zero.

Note the asymmetry: an unknown warning *code* degrades gracefully to `UNKNOWN_VALIDATION_WARNING`, but a known code with an unexpected shape is fatal. Same for the `Invalid validation warning` throw below. Suggest degrading both to `UNKNOWN_VALIDATION_WARNING` (or dropping the warning) rather than rejecting the response.

sdk/src/utils/orderTransactions/api.ts
```diff
@@ -152,8 +206,129 @@ export type PrepareCollateralRequest = {
...
+];
+
+function isParsedTradingCapacity(value: unknown): value is TradingCapacity {
+  if (typeof value !== "object" || value === null || Array.isArray(value)) {
+    return false;
+  }
+
+  const capacity = value as Record<string, unknown>;
+  return (
+    typeof capacity.availableLiquidity === "bigint" &&
```
divhead: These three only check `typeof === "bigint"`, with no range check. `deserializeBigIntsInObject` parses with `/^-?\d+$/` (`sdk/src/utils/numbers/utils.ts`), so `availableLiquidity: "-1"` deserializes to `-1n` and passes validation — while `requestedSizeUsd` in the same object must be a canonical uint256 via `parseUint256DecimalString`.

The new `isUint256` helper is exactly for this; apply it to the three liquidity fields so a negative capacity cannot reach an integrator's sizing logic.

sdk/src/utils/orderTransactions/api.ts
```diff
@@ -152,8 +206,129 @@ export type PrepareCollateralRequest = {
...
+    typeof capacity.baseAvailableLiquidity === "bigint" &&
+    typeof capacity.jitAvailableLiquidity === "bigint" &&
+    typeof capacity.limitingFactor === "string" &&
```
divhead: These three are validated as "any string" but the return type asserts closed unions (`TradingCapacityLimitingFactor`, `JitDataStatus`, `MarketDataStatus`).

An integrator writing an exhaustive `switch (capacity.limitingFactor)` gets a runtime value TypeScript believes is impossible; and the README's own example branches on `jitDataStatus === "available"`, so an unexpected value silently falls into the "no JIT" path.

The warning `reason` field is checked against its literal list further down — do the same here.

sdk/src/utils/orderTransactions/api.ts
```diff
@@ -152,8 +206,129 @@ export type PrepareCollateralRequest = {
...
+  }
+
+  if (PREPARE_ORDER_ERROR_CODES.includes(body.code as PrepareOrderErrorCode) && typeof body.message === "string") {
+    return {
+      code: body.code as PrepareOrderErrorCode,
+      message: body.message,
+    };
+  }
+
+  if (body.code !== undefined) {
```
divhead, line 282: Any `code` outside the hardcoded five makes this return `undefined`, so the error stops being a prepare error at all (covered by the `an unknown code` test case).

When gmx-api adds a documented code — rate limit, market disabled, and so on — every integrator's `parsePrepareOrderError(e)?.code` chain falls through to generic handling and loses `message` and `traceId` along the way.

The warnings path already solved this with `UNKNOWN_VALIDATION_WARNING`; mirror it for errors (e.g. return `{ originalCode, message }`) so at least the message survives an API that moves ahead of the SDK.

sdk/README.md
```diff
@@ -230,3 +230,51 @@ sdk.orders.swap({
...
+
+const safeAccountAgnosticSize = capacity.baseAvailableLiquidity;
+const indicativeGlobalCapacity =
+  capacity.jitDataStatus === "available" ? capacity.availableLiquidity : capacity.baseAvailableLiquidity;
+```
+
+`availableLiquidity` is the current global JIT-aware capacity and remains capped by max open interest.
+It is indicative because the keeper can restrict JIT by account. `baseAvailableLiquidity` is the native capacity without
+JIT, but it is still a market-level snapshot: a collateral swap or intervening market state can reduce the capacity
+available to a specific order. The keeper can use JIT for market, limit, stop-market, and TWAP increases, but availability
```
divhead, line 259: This says the keeper can use JIT for market, limit, stop-market, and TWAP increases. The PR description says the opposite: global JIT uplift for *market* increases, and native `baseAvailableLiquidity` for limit, stop-market, and TWAP. Both cannot be right, and this is the exact question an integrator sizing a limit order needs answered.

For what it's worth, the UI currently applies JIT-inclusive liquidity to TWAP increases (`src/domain/synthetics/orders/utils.tsx`, `getJitMaxReservedUsd` in the order liquidity check), which matches the README rather than the description. FEDEV-4012 (resting limit orders flagged insufficient liquidity despite the header) looks like the same ambiguity already surfacing as a bug.

Please settle which it is and state it explicitly here. If JIT eligibility depends on order type, then `GetTradingCapacityParams` (symbol + direction only) cannot answer the question, and that gap is worth calling out too.

Separately, the PR description says the helper is backed by `/v1/markets/tickers` — code, README and tests all say `/v1/markets/trading-capacity`.

sdk/src/clients/v2/index.ts
```diff
@@ -268,10 +294,14 @@ export class GmxApiSdk {
     return fetchApiMarkets(this.ctx);
   }
 
-  fetchMarketsTickers(params?: { addresses?: string[]; symbols?: string[] }): Promise<MarketTicker[]> {
+  fetchMarketsTickers(params?: { addresses?: string[]; symbols?: string[] }): Promise<MarketTickerWithCapacity[]> {
     return fetchApiMarketsTickers(this.ctx, params);
   }
 
+  async getTradingCapacity(params: GetTradingCapacityParams): Promise<TradingCapacity> {
```
divhead: `async` adds nothing here — the body is a single `return` of a promise. Every sibling method on this class (`fetchMarketsTickers`, `fetchTokensData`, `fetchMarkets`) is a plain function; drop it for consistency.

divhead (changes requested):
Blocking on rollout ordering plus a few correctness issues in how the new capacity payload is trusted.

**Rollout blocker.** `getTradingCapacity()` calls `/v1/markets/trading-capacity`, and the companion `gmx-io/gmx-api#105` is still open. Against production today:

```
curl -sS -o /dev/null -w "%{http_code}\n" \
  "https://arbitrum.gmxapi.io/v1/markets/trading-capacity?symbol=ETH%2FUSD%20%5BWETH-USDC%5D&direction=long"
# 404
```

Live `/v1/markets/tickers` also has no `capacityLong`/`capacityShort` yet, so `MarketTickerWithCapacity` is always undefined for now (harmless, since both fields are optional). The API side has to deploy before the SDK is published, otherwise the README's headline method 404s for every integrator.

**Trust model.** The same capacity payload is validated three different ways depending on where it arrives: strictly (and fatally) on the prepare-warning path, not at all on `getTradingCapacity()`, and only shallowly for the enum-typed fields. Worth making consistent before this becomes a public surface — see the inline comments.

**Docs.** The README and the PR description disagree on which order types get the JIT uplift; that is the exact question an integrator sizing a limit order needs answered.

CI is green and the test coverage on uint256 parsing and forward-compatible warning codes is solid — the gaps are in what happens when the API payload drifts, not in the tests themselves.

divhead (commented):
Following up on the earlier review with the point underneath most of those findings: the response parsing here doesn't need to be tightened field by field, it needs to follow the parser convention this SDK already has.

**The tell.** `parseTradingCapacity` casts (`deserializeBigIntsInObject(raw, { handleInts: true }) as TradingCapacity`, `utils/markets/api.ts:54`), and then `isParsedTradingCapacity` runtime-checks the type that cast just asserted. TypeScript doesn't need that guard — `capacity` is already `TradingCapacity` at the call site. The guard exists only because the cast isn't true. A parser that takes `unknown` and returns `TradingCapacity | undefined` in one step has nothing to un-assert afterwards.

**Why the generic walker is the wrong tool here.** `deserializeBigIntsInObject` doesn't know the schema. It converts any string matching `/^-?\d+$/` anywhere in the object, which produces three separate problems:

- `availableLiquidity: "-1"` deserializes to `-1n` and passes `typeof === "bigint"`, so a negative capacity reaches an integrator's sizing logic. Meanwhile `requestedSizeUsd` in the same object is held to a canonical uint256.
- If the API sends a JSON number instead of a decimal string, `handleInts` leaves it alone, the guard fails, and on the warning path that's a hard throw. Parsing success depends on a wire-format detail nothing pins down.
- `requestedSizeUsd` inside `details` gets converted by the walker and then overwritten by the separately parsed value.

And the enum-typed fields (`limitingFactor`, `jitDataStatus`, `marketDataStatus`) are checked as "any string" while the return type asserts closed unions — so an exhaustive `switch` downstream can hit a value TypeScript believes is impossible. The `reason` field on the sibling `TRADING_CAPACITY_UNAVAILABLE` warning is checked against its literal list a few lines below, which is the right shape.

**The convention already exists, in this feature area.** `utils/jitLiquidity/utils.ts` parses the JIT liquidity response the way this should work: `parseJitLiquidityResponse(response: unknown)` walks fields explicitly through `isRecord` / `getRecord` / `getString` from `utils/objects`, with no cast and no generic walker. Its amount parser closes exactly the two holes above:

```ts
function parseJitAmount(value: unknown): bigint {
  if (typeof value !== "string" && typeof value !== "number" && typeof value !== "bigint") return 0n;
  try {
    const amount = BigInt(value);
    return amount < 0n ? 0n : amount;
  } catch { return 0n; }
}
```

Accepts either wire representation, refuses negatives. This PR is the JIT capacity follow-up and it routes around the JIT parser sitting one directory over. There's no schema library in the SDK, so `utils/objects` is the house pattern.

**Suggested shape.** One `parseTradingCapacity(raw: unknown): TradingCapacity | undefined`: `isRecord`, then the six fields read by name — a uint256 parser for the three amounts, literal-set membership for the three enums — returning `undefined` on any mismatch. Around 25 lines. Every entry point calls it, **including `getTradingCapacity()`** (`clients/v2/index.ts:301`), which today bypasses all of this and hands back an unchecked cast — the documented method integrators are told to size orders with is the one with no validation at all.

That also collapses the duplication: `parsePrepareOrderErrorBody`, `isPrepareOrderError`, and `parseValidationWarning` currently each re-derive "is this an INSUFFICIENT_LIQUIDITY payload" with slightly different rules (`parseUint256DecimalString` on raw vs `isUint256` on parsed). Three copies of one rule will drift. With a single parser, `isParsedTradingCapacity` disappears, and so does the question of why identical malformed input returns `undefined` on one path and throws on another.

sdk/src/utils/orderTransactions/api.ts
```diff
@@ -152,8 +206,129 @@ export type PrepareCollateralRequest = {
...
 
+const PREPARE_ORDER_ERROR_CODES: readonly PrepareOrderErrorCode[] = [
+  "INVALID_PARAMS",
+  "MARKET_NOT_FOUND",
+  "TOKEN_NOT_FOUND",
+  "NO_SWAP_PATH",
+  "POSITION_NOT_FOUND",
+];
+
+function isParsedTradingCapacity(value: unknown): value is TradingCapacity {
```
divhead: This guard is the symptom. By the time it runs, `capacity` is already typed `TradingCapacity` — `parseTradingCapacity` cast it. So this is a runtime check of an assertion made two lines earlier, which is the shape you get when the cast isn't trustworthy.

Folding the validation into the parse (`unknown` in, `TradingCapacity | undefined` out) removes the need for this function entirely, and lets each field get the check it actually needs instead of `typeof === "bigint"` / `typeof === "string"`. See the top-level comment for the suggested shape and the existing `utils/jitLiquidity` precedent.

sdk/src/utils/markets/api.ts
divhead: Agreed on the main point — the method is new, nothing consumes it yet, so there's nothing to break. Dropping this as a blocker.

One residual worth handling cheaply: "then the API" isn't a single deploy, it's one per chain, and they demonstrably drift. Right now:

```
/v1/risk-oracle/markets
arbitrum=200  avalanche=200  megaeth=503   (three consecutive probes, megaeth returns {"message":"Internal Server Error"})
```

So after the API ships, `getTradingCapacity()` will work on some chains and fail on others for a while — and it'll fail as a raw `HttpError`, indistinguishable from a network problem, which pushes every integrator into guessing.

Not asking you to hold the PR. Just worth mapping a 404 from this route to a typed "capacity not available on this deployment" so callers can fall back to `baseAvailableLiquidity` from tickers deliberately instead of catching a generic error. Few lines, and it closes the question permanently rather than for one rollout.

(Separately, that megaeth 503 looks like a live issue unrelated to this PR — might be worth a look.)

sdk/src/utils/markets/api.ts
divhead: Fair on the file itself — I checked, and `markets/api.ts` is uniformly cast-based: `fetchApiMarketsInfo`, `fetchApiMarketsConfig`, `fetchApiMarketsValues`, `fetchApiTokensData`, `fetchApiMarketsTickers` all go through `deserializeBigIntsInObject(...) as X`. `fetchApiTradingCapacity` sits in that row consistently, and consistency with surrounding code is a real argument.

One correction on the premise though: there is a response-validation example, in this same feature area. `utils/jitLiquidity/utils.ts` — `parseJitLiquidityResponse(response: unknown)` reads fields explicitly via `isRecord` / `getRecord` / `getString` from `utils/objects`, and its `parseJitAmount` accepts either a string or a number and refuses negatives.

But the real point isn't "add validation here" — it's that the same `TradingCapacity` object gets two opposite policies inside one PR. Through `getTradingCapacity()` it's an unchecked cast; through `prepareOrder` it's validated strictly enough that a mismatch throws and fails the whole request. Same payload, same shape, opposite severity.

So it's a pick-one, and **the cheaper option is yours, not mine**: keep the cast here, matching the file, and drop the `throw` in `parseValidationWarning` so a malformed advisory field degrades instead of killing the response. That's less code than what's there now, it makes the policy consistent, and it also resolves the separate comment about `prepareOrder` rejecting on a warning. Happy with that outcome if you'd rather not add a parser.

sdk/README.md
divhead: Thanks — that settles it, and you're right. Withdrawing this.

For the record, the FEDEV-4012 reference I cited actually supports your answer rather than contradicting it: that issue was closed on 25 June by #2586, "include JIT liquidity in account dashboard resting order check" — the bug was the UI *not* applying JIT to resting limit orders when the keeper does. And current master already reflects it: `src/domain/synthetics/orders/utils.tsx` calls `getJitMaxReservedUsd` for the `LimitIncrease && !isTwapOrder` branch. So the README text here is correct as written.

Only the PR description needs updating then — it says JIT uplift applies to market increases with native `baseAvailableLiquidity` for limit, stop-market, and TWAP, which is the opposite of both the README and the keeper behaviour.

Also withdrawing the follow-on point in this comment about `GetTradingCapacityParams`: if JIT eligibility doesn't depend on order type, then symbol + direction is the right shape and there's no gap.

divhead (approved):
Re-reviewed `6c76234` — everything from the review round is addressed, approving.

- `parseTradingCapacity` is now the single strict parser (`unknown` → `TradingCapacity | undefined`): uint256-checked amounts (negatives rejected, string/bigint/safe-number wire forms accepted), literal-set membership for all three enum fields, built on `isRecord`/`getString` per the `utils/objects` house pattern. All four entry points go through it — `getTradingCapacity()` (fails loudly now), tickers (drops malformed capacity, keeps the ticker), estimates (drops), warnings (degrades).
- The advisory-path throws are gone: a malformed known-code warning degrades to `UNKNOWN_VALIDATION_WARNING`, code-less garbage is dropped, a non-array `validationWarnings` no longer rejects the response. A payload drift in an advisory field can no longer take `prepareOrder` down — this was my main remaining concern.
- Unknown prepare error codes now survive as `UNKNOWN_PREPARE_ORDER_ERROR` with `originalCode`, `message`, and `traceId` preserved — mirrors the warnings design, exactly the forward-compat shape needed.
- Description now matches the code (`/v1/markets/trading-capacity`, JIT for market/limit/stop-market/TWAP increases per keeper behavior), `async` dropped. With the API side shipping `MARKET_NOT_FOUND` on its 404, the rollout ambiguity is branchable via `HttpError.code` without further SDK changes.

One note on the red check: the failing step is the SDK e2e suite added in #2699, and the failing spec (`subaccount.spec.ts`, 1CT signOrder) dies on a server-side `400 Collateral after fees is below the $1.00 minimum` from the test API deployment — gmx-api's new min-collateral validation vs the e2e fixture, nothing this diff touches. It still gates the merge, but the fix belongs to the e2e fixtures (raise collateral enough to clear the minimum *after fees*).

Release ordering stands as agreed in the threads: deploy gmx-api#105 first, then publish the SDK.

PR #2673, opened 2026-07-29
Add v2.2c SDK token inventory parity FEDEV-4127

qwinndev:
## Summary

- Align `@gmx-io/sdk` position price impact with v2.2c token-denominated open interest and virtual inventory, including exact decrease rounding, the zero-impact boundary, and fail-closed missing data.
- Validate the required v2.2c market and position API fields and propagate exact token sizes through decrease previews and off-hours liquidation calculations.
- Keep legacy USD-mode and swap impact behavior unchanged, and bump the SDK to `1.6.4`.
- Linear: [FEDEV-4127](https://linear.app/gmx-io/issue/FEDEV-4127/api-and-sdk-parity-for-token-based-virtual-inventory-in-position-price)

## Testing

- `yarn test:ci` — 983 passed, 14 skipped.
- `yarn test:ct` — 159 Chromium component tests.
- `cd sdk && yarn test:ci` — 574 tests.
- `cd sdk && yarn test:multicall` — 14 passed, 1 skipped.
- Root and SDK lint, typecheck, Prettier, SDK build, package dry run, and CJS/loader-style ESM/TypeScript consumer smoke.
- Exercised negative/zero/positive impact, token mode on/off, missing fields, zero price and inventory, long/short rounding, full closes, unloaded positions, and swap isolation.
- No screenshots: this PR has no visual UI changes.


sdk/src/utils/trade/decrease.ts
```diff
@@ -155,16 +176,18 @@ export function getDecreasePositionAmounts(p: {
...
-      position,
-      marketInfo,
-      isLong,
-      isTrigger,
-      fixedAcceptablePriceImpactBps,
-      acceptablePriceImpactBuffer,
-      values,
-      isSetAcceptablePriceImpactEnabled,
-    });
+    if (!marketInfo.useOpenInterestInTokensForBalance) {
```
divhead, line 179: Dormant footgun: for a token-OI market with no loaded position this skips `applyAcceptablePrice`, so the returned amounts silently carry `acceptablePrice = 0n` (the new test enshrines exactly that).

Today no live submit path reaches it — I traced all of them: tradebox Trigger mode is auto-reset by the `updateTradeMode` effect (Trigger is not in `AVAILABLE_TRADE_MODES`), the OrderEditor derives its acceptable price from bps itself (`useAcceptablePrice`), and the new-position TP/SL sidecar builds a mock position with exact token sizes. But `selectTradeboxDecreaseOrderPayload` still passes `decreaseAmounts.acceptablePrice` through verbatim, so any future revival of standalone trigger creation ships orders with `acceptablePrice = 0`: unexecutable for short decreases (the no-protection sentinel for shorts is `MaxUint256`, not 0) and unprotected for long decreases.

Since `useOpenInterestInTokensForBalance` is already `true` in prod, this branch activates on every market the moment this ships. Prefer failing closed explicitly here (`undefined`/throw) over a silent 0 — consistent with the rest of this PR.

sdk/src/utils/markets/api.ts
```diff
@@ -1,22 +1,59 @@
 import { IHttp } from "utils/http/types";
+import { assertApiFields, assertApiRecords } from "utils/http/validation";
+import type { ApiFieldSpec } from "utils/http/validation";
 import { deserializeBigIntsInObject } from "utils/numbers";
 import { TokenData } from "utils/tokens/types";
 
 import { MarketTicker, MarketWithTiers, RawMarketConfig, RawMarketInfo, RawMarketValues } from "./types";
 
+const MARKET_CONFIG_V22C_FIELDS = [
```
divhead, line 9: Rollout coupling: these fields are validated fail-closed, but prod does not serve them yet. Probed `arbitrum.gmxapi.io` today: `/v1/markets/config` already returns `useOpenInterestInTokensForBalance: true` but is missing `maxCollateralSum*`, `minFundingIncreaseRatePerSecond` and `virtualIndexTokenId`; `/v1/markets/values` lacks `virtualInventoryForPositionsInTokens`.

So until gmx-io/gmx-api#114 is deployed, every `fetchMarketsConfig/Values` throws and all sessions run on the RPC fallback (graceful via `useApiDataFallbackState`, but fleet-wide RPC load). Not a code bug — just deploy order: gmx-api first, then this + the SDK 1.6.4 publish.

divhead (commented):
Deep pass on the price-impact math and the API coupling. Verdict: LGTM with two notes inline plus the CI/rollout notes below.

Contract parity checked line-by-line against gmx-synthetics `release-v2.2c`, all five points match:
- `>= 0` early return before the virtual-inventory branch (`PositionPricingUtils.getPriceImpactUsd`);
- `<=` in `isSameSideRebalance` (both position and swap paths use `<=` in the contracts);
- next-OI clamp to zero on over-decrease (`getNextOpenInterestParams`);
- virtual-inventory participation gated by `virtualTokenId != bytes32(0)` with token→USD conversion at mid price (the pre-existing offset trick for negative usdDelta matches `getNextOpenInterestForVirtualInventory`);
- decrease/increase token-delta rounding per `PositionUtils` (long roundUp / short floor on decrease; long floor@max / short roundUp@min on increase; full-close returns `sizeInTokens`).

**CI note:** no lint/tsc/test workflows ran on this PR — the base branch `qwinndev/update-v22c-contracts` is not in the workflow triggers, only Cloudflare builds ran. Ran the checks locally instead: `yarn tscheck` clean; `sdk yarn test:ci` 573/574 (the one failure is the network-dependent keeper-markets consistency test timing out in my environment, unrelated to this diff); targeted vitest for the changed `src/` files passes. Worth re-checking CI once the base PR merges and this retargets.

**Rollout:** prod API already serves `useOpenInterestInTokensForBalance: true` (token-OI mode is live), see the inline note on `api.ts` — gmx-io/gmx-api#114 must deploy before this ships, otherwise the whole fleet silently drops to the RPC fallback.

divhead (approved):

divhead:
Follow-up: the only post-review delta here is the base-branch sync (release merged into qwinndev/update-v22c-contracts) — the PR's own patch is unchanged (same 25 files, +995/-176), so the review stands for the merged state. The umbrella #2310 is merged to release and prod gmx-api already serves the required v2.2c fields (probed live), so the fail-closed validation is satisfied. Remaining from the rollout list: publish @gmx-io/sdk 1.6.4 and the gmx-api compat-types follow-up.

PR #2675, opened 2026-07-29
Fix api health tracker false positives: clock skew and tab-refocus flaps

divhead:
## Problem

At 100% api-positions rollout (Jul 20-29, Arbitrum) the `positionsListLoad` metric shows: 25% of successful loads are cold-RPC fallbacks (p50 ~11s, p90 ~28s) and 41.6% of all loads hit the 10s metric timeout — while the backend `/v1/positions` stays fast (p90 ~250ms). The dominant driver is the api health tracker (introduced Jun 16) force-falling everyone back to a cold RPC path on false-positive "markets stale" flips. Two false-positive sources:

1. **Clock skew.** `hasStaleMarketValues` compared server `updatedAt` to the client clock. A client >10s ahead makes the api permanently unhealthy: every load for that user goes through the cold RPC fallback forever.
2. **Tab refocus.** SWR pauses polling in hidden tabs, so on return the first render evaluates aged values and reported stale before the refocus refetch (~1s) could land. One stale render flipped the whole chain unhealthy for >=15s (restore cooldown) — a guaranteed api teardown plus cold RPC detour on every return to the tab (#2669 removed the data flash this caused; this PR removes the flip itself).

## Change

- `hasStaleMarketValues` never mixes clock domains: a market is stale when it lags the newest market of the same response beyond `MARKETS_STALE_THRESHOLD_MS` (server-vs-server), missing/null values stay stale as before.
- A fully frozen backend snapshot (responses keep arriving but the newest `updatedAt` never advances) is still detected, via `trackValuesSnapshot`/`getIsValuesSnapshotFrozen` comparing client receipt times only (skew-free), keyed per chain.
- `ApiHealthTracker` requires staleness to persist for `UNHEALTHY_CONFIRMATION_MS = 3s` before flipping unhealthy. Refocus blips (fresh data lands in ~1s) no longer flip; genuine staleness flips 3s later. Restore-cooldown semantics unchanged; a fresh report resets the pending confirmation window.

## Verification

- `yarn tscheck` clean
- `yarn test:ci`: 981 passed / 14 skipped / 0 failed
- Updated `apiHealthTracker.spec.ts` for the confirmation-window semantics (blip stays healthy, persistence flips, cooldown/reset/notify cases) and `useApiMarketsInfoRequest.spec.ts` for lag-based staleness, clock-skew immunity, and snapshot-freeze tracking.

## Expected effect

Fewer force-fallback windows → the `dataSource:rpc` share inside `apiSdkPositions:true` and the timeout rate should drop after deploy. Worth watching `positionsListLoad` p90 / timeout split by `dataSource` for a few days after release.

### DIFF
--- a/src/domain/api/apiHealthTracker.spec.ts
+++ b/src/domain/api/apiHealthTracker.spec.ts
@@ -4,9 +4,18 @@ import type { ContractsChainId } from "sdk/configs/chains";
 
 import { ApiHealthTracker } from "./apiHealthTracker";
 
+const CONFIRMATION_MS = 3_000;
 const COOLDOWN_MS = 15_000;
 
 const tracker = ApiHealthTracker.getInstance();
+
+// Reports stale at startAt and again once the confirmation window has elapsed.
+function makeUnhealthy(chainId: ContractsChainId, startAt = 0) {
+  vi.setSystemTime(startAt);
+  tracker.reportMarketsFreshness(chainId, true);
+  vi.setSystemTime(startAt + CONFIRMATION_MS);
+  tracker.reportMarketsFreshness(chainId, true);
+}
 
 // Each test uses a distinct chainId so the shared singleton's per-chain state stays isolated.
 describe("ApiHealthTracker", () => {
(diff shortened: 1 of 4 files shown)


divhead:
Superseded by #2678 (same commit, hotfix-* branch name to satisfy check-naming).

PR #2703, opened 2026-08-04
Rewrite README and unify docs

divhead:
[FEDEV-4142](https://linear.app/gmx-io/issue/FEDEV-4142/update-readme-and-unify-docs-in-gmx-interface)

Docs-only change. No runtime code touched.

## README

Replaced the Create React App boilerplate with a description of what the repo actually is: the trading app (`src/`), the landing page (`landing/`), and the `sdk/` workspace published as `@gmx-io/sdk`. Setup, the script table (checked against `package.json`), and links to the docs site.

## Changelog deduplication

The SDK changelog existed in two diverged copies: `sdk/CHANGELOG.md` here (terse, ending at 1.3.1 plus an Unreleased entry) and `docs/sdk/changelog.md` in gmx-docs (full dated history through 1.6.4). gmx-docs is the source of truth, so `sdk/CHANGELOG.md` is removed and its Unreleased entry moved to the docs page in gmx-io/gmx-docs#171.

The file was never part of the published package (`files: ["build"]`) and nothing in the repo or CI referenced it.

## docs/

`docs/multichain-add-chain.md` and `docs/style-guidelines.md` are removed; both are kept in the private gmx-prompts repo, where the multichain guide has been updated with the `VIEM_CHAIN_BY_CHAIN_ID` step and the Rabby allowlist note from this copy.

## Verification

Prettier clean. Grepped for references to the removed files and for CRA leftovers — none. The paired gmx-docs PR builds with `onBrokenLinks: "throw"`.

PR #2704, opened 2026-08-04
Rewrite README and unify docs

divhead:
[FEDEV-4142](https://linear.app/gmx-io/issue/FEDEV-4142/update-readme-and-unify-docs-in-gmx-interface)

Docs-only change. No runtime code touched. Replaces #2703, which was opened against `master` by mistake.

## README

Replaced the Create React App boilerplate with a description of what the repo actually is: the trading app (`src/`), the landing page (`landing/`), and the `sdk/` workspace published as `@gmx-io/sdk`. Setup, the script table (checked against `package.json`), and links to the docs site.

## Changelog deduplication

The SDK changelog existed in two diverged copies: `sdk/CHANGELOG.md` here (terse, ending at 1.3.1 plus an Unreleased entry) and `docs/sdk/changelog.md` in gmx-docs (full dated history through 1.6.4). gmx-docs is the source of truth, so `sdk/CHANGELOG.md` is removed and its Unreleased entry moved to the docs page in gmx-io/gmx-docs#171.

The file was never part of the published package (`files: ["build"]`) and nothing in the repo or CI referenced it.

## docs/

`docs/multichain-add-chain.md` and `docs/style-guidelines.md` are removed; both are kept in the private gmx-prompts repo, where the multichain guide has been updated with the `VIEM_CHAIN_BY_CHAIN_ID` step and the Rabby allowlist note from this copy.

## Verification

Prettier clean. Grepped for references to the removed files and for CRA leftovers — none. The paired gmx-docs PR builds with `onBrokenLinks: "throw"`.


divhead:
`Linting, Type Checking, and Tests` is red, and it is not from this PR — the failure is pre-existing on `release`.

The failing step is `Run SDK e2e public tests`:

```
FAIL src/clients/v2/__tests__/subaccount.spec.ts > GmxApiSdk — subaccount (1CT) > signOrder uses subaccount signer when active
HttpError: HTTP 400: Collateral after fees is below the $1.00 minimum required to open the position
```

Evidence:

- This branch differs from `release` in four markdown files only — `README.md`, `docs/multichain-add-chain.md`, `docs/style-guidelines.md`, `sdk/CHANGELOG.md`. No TypeScript, config, or workflow changes; `git diff --name-only release HEAD` returns exactly those four. Nothing under `sdk/src` references `CHANGELOG`.
- The same commit content on a `master` base passed this job twice (#2703) — the e2e suite only exists on `release`.
- The suite arrived via #2699, which merged at 22:21 on Aug 3 with its own three most recent runs red. This PR is the first `release`-based PR to run it since.
- Re-running the job reproduces the identical failure, so it is a persistent condition rather than a random flake.

The test appears to open a position with a fixed collateral amount that current fees now exceed. Fixing it is out of scope here; flagging so it is not attributed to this change.

Lint, typecheck, and the remaining test files pass in that same job; `SDK standalone build (publish parity)`, `Playwright Component Tests`, `Audit`, and both Cloudflare Pages builds are green.

PR #2705, opened 2026-08-04
Smart account multichain FEDEV-4161

midas-myth:


src/lib/wallets/signing.ts
```diff
@@ -320,15 +335,128 @@ export async function signTypedData({
...
+  const firstSignatureIsValid = await verifySignatureOnVerificationChain({
+    signedHash,
+    signature: firstResult.signature,
+    expectedAccount: from,
+    verificationChainId,
+  });
+
+  if (firstSignatureIsValid !== false) {
+    return firstResult.signature;
+  }
```
qwinndev, line 400: Declining the retry hard-fails an order that previously would have gone through — src/lib/wallets/signing.ts:398
If verifyHash says the first signature is invalid and the user then rejects the network switch (or the chain isn’t available), withSmartWalletChainSwap throws SMART_WALLET_CHAIN_UNAVAILABLE_ERROR/WRONG_CHAIN_ERROR and signTypedData rejects — discarding a first signature the PR’s own comment says may only look invalid (“verifyHash also reports false when the deployless ERC-6492 eth_call is unsupported, so the relay stays the authority”). Confirmed with a repro test. Suggested fix: try { retryResult = await ... } catch { return firstResult.signature } — let the relay reject it, as the comment already argues.

src/lib/wallets/verificationChainSigning.ts
```diff
@@ -0,0 +1,137 @@
...
+  const state = readLocalStorageItem<ChainBindingState>(storageKey, { deserializer: JSON.parse });
+
+  if (state !== undefined) {
+    forgetStateIfDelegationRevoked(address, verificationChainId, storageKey);
+  }
+
+  return state;
+}
+
+export function rememberChainBindingState(address: string, verificationChainId: number, state: ChainBindingState) {
```
qwinndev, line 48: are we sure that one fail is enough to disable feature? maybe better to add ttl or something like that to try later, because error could be temporary

src/lib/wallets/signatureDiagnostics.ts
```diff
@@ -0,0 +1,110 @@
...
+
+export type SignatureShape = {
+  signatureKind: SignatureKind;
+  signatureBytes: number | undefined;
+  ecdsaV: number | undefined;
+  isHighS: boolean | undefined;
+  isOnChainRecoverableShape: boolean | undefined;
+};
+
+/** Diagnostic only: an odd shape is not proof of rejection, since ERC-1271 is tried after ECDSA on-chain. */
```
qwinndev: i again see that llm left a lot of comments. not sure that we need all of them, lets revisit that

src/lib/wallets/signing.ts
midas-myth: added rethrow on chain switch failed

src/lib/wallets/verificationChainSigning.ts
midas-myth: added ttl - 30d
+ existing background checks after signature

src/lib/wallets/signatureDiagnostics.ts
midas-myth: sorry, cleaned up

qwinndev (approved):

PR #2718, opened 2026-08-06
Add detailed and tax-compatible CSV exports FEDEV-4133

qwinndev:
## Summary

- add versioned detailed CSV exports for complete Trade and Claims history with pagination, deterministic UTC timestamps, and audit metadata
- add Koinly, CoinTracker, and CoinLedger adapters with safe USD/address fallbacks, loss-prevention checks, and CoinLedger ZIP packaging
- add cancellable progress UI using GMX styling, stable modal layout, translations, and CSV/ZIP/export coverage

Linear: https://linear.app/gmx-io/issue/FEDEV-4133/add-detailed-and-tax-compatible-csv-exports-for-trade-and-claims


src/lib/csv.ts
```diff
@@ -1,29 +1,75 @@
...
+  return values ? `${header}${CSV_LINE_BREAK}${values}${CSV_LINE_BREAK}` : `${header}${CSV_LINE_BREAK}`;
+}
+
+function convertToCSV<T>(data: Partial<T>[], customHeaders?: Partial<Record<keyof T, string>>): string {
+  const keys = customHeaders ? Object.keys(customHeaders) : data[0] ? Object.keys(data[0]) : [];
+
+  const headers = keys.map((key) => customHeaders?.[key as keyof T] ?? key);
+  const rows = data.map((object) => Object.fromEntries(keys.map((key) => [key, object[key as keyof T]])) as CsvRow);
+
+  return serializeCsv(headers, rows);
```
timgmxio, line 63: `rows` are keyed by field name while `headers` are the display names, and `serializeCsv` looks up `row[header]` — so a call with `customHeaders` would produce a header line and empty rows. no caller passes them after this pr, so nothing is broken right now, just a sharp edge left in a shared helper.

src/domain/synthetics/historyExport/tradeExport.ts
```diff
@@ -0,0 +1,568 @@
...
+    if (isSwapOrderType(action.orderType)) {
+      return true;
+    }
+
+    const matchingFilters = collateralFilters.filter(
+      (filter) =>
+        filter.direction === (action.isLong ? "long" : "short") &&
+        areAddressesEqual(filter.marketAddress, action.marketAddress)
+    );
+    if (!matchingFilters.length) {
```
timgmxio: when no filter matches the action's market+direction we keep the action, while `processTradeActions` does the opposite — `Boolean(desiredCollateralAddresses?.[...])` is false there, so the table hides limit / trigger-decrease actions for a market whose collateral filter isn't set. with **[ETH long, USDC] + [BTC long, any collateral]** selected, a **BTC LimitIncrease** is hidden in the table but present in the csv.

src/domain/synthetics/historyExport/providerExport.ts
```diff
@@ -0,0 +1,540 @@
...
+    Boolean(sentAmount) !== Boolean(sentCurrency) || Boolean(receivedAmount) !== Boolean(receivedCurrency);
+
+  return !hasIncompleteSide && (hasCompleteSentSide || hasCompleteReceivedSide);
+}
+
+function getActionRows(rows: CsvRow[]): CsvRow[] {
+  return rows.filter((row) => getCsvString(row, "row_type") === "action");
+}
+
+function getCashflow(rows: CsvRow[], actionId: string): CsvRow | undefined {
```
timgmxio: `getCashflow` scans all canonical rows and runs once per action row (411/428/447), so the cost is quadratic with two string coercions per probe. it happens after the fetch phase, and progress only tracks fetching — on a large export the tab sits frozen once the bar has already reached 100% and the cancel button is gone.

src/domain/synthetics/historyExport/tradeExport.ts
```diff
@@ -0,0 +1,568 @@
...
+
+function formatTokenUsd(amount: string | null | undefined, price: string | null | undefined): string {
+  if (amount === undefined || amount === null || price === undefined || price === null) {
+    return "";
+  }
+
+  return formatUsdDecimal(BigInt(amount) * BigInt(price));
+}
+
+function getTwapPartByActionId(rawActions: SubsquidTradeAction[]): Map<string, number> {
```
timgmxio: parts are numbered over the actions present in this export, while twap_parts_total comes from the indexer. a date range that cuts a twap order yields twap_part=1 / total=10 for what is part 6, and the same action re-exported under different filters gets a different number — twap group/part is listed among the stable dedup references in the schema.

src/domain/synthetics/historyExport/csvSchemas.ts
```diff
@@ -0,0 +1,178 @@
...
+  "position_fee_usd",
+  "borrowing_fee_amount",
+  "borrowing_fee_usd",
+  "funding_fee_amount",
+  "funding_fee_usd",
+  "liquidation_fee_amount",
+  "liquidation_fee_usd",
+  "swap_fee_amount",
+  "swap_fee_usd",
+  "ui_fee_amount",
```
timgmxio, line 51: `ui_fee_amount` / `ui_fee_usd` are never assigned, so the `["ui_fee_amount", "ui_fee_usd"]` pair in `getFeeComponents` never emits a leg and ui fees are absent from the koinly/cointracker/coinledger files, not just blank in the canonical one. `TradeAction` exposes `uiFeeFactor` and the sdk derives the fee as `applyFactor(sizeDeltaUsd, uiFeeFactor)` (`sdk/src/utils/fees/index.ts:52`); the history query doesn't fetch it. `swap_fee_amount`, `referral_discount_usd` and `pro_discount_usd` are blank too, but the usd/aggregate columns next to them are populated and reach the providers.

timgmxio (dismissed):

timgmxio (approved):

PR #2739, opened 2026-08-11
Сonditional margin deposits at a trigger price FEDEV-2154

qwinndev:
https://linear.app/gmx-io/issue/FEDEV-2154/feature-allow-conditional-margin-deposits-at-a-trigger-price

A conditional margin deposit is a margin-only LimitIncrease (`sizeDeltaUsd = 0`, positive collateral, `autoCancel: true`) in the position's current collateral token, with the trigger in the protective direction.

- `Edit margin` → Deposit offers `Now` / `At price`. At price adds a trigger price input, values all projections at the trigger (index-token collateral is valued at the trigger price), keeps `Liquidation price` between the CTA and `Execution details`, and reorders the details rows to fees / network fee / margin / leverage / size. Insufficient deposits and the per-position auto-cancel limit hard-block with the ticket copy; a trigger at or beyond the current estimated liquidation price shows the non-blocking warning.
- Amount changes go through the replace flow: the `At price` form bound to the existing order, `Replace margin deposit` CTA, cancel + create in one batch submission. Reached from the order editor; closing without submitting changes nothing.
- TradeBox shows a dismissible suggestion card for every Limit Increase on an existing position (not Stop Market), independent of Size/Margin/Trigger values; dismissal persists via localStorage. The action opens the editor prefilled only with valid values — a pay token other than the position collateral token leaves the amount empty (wrapped/native equivalence counts as matching; no conversion).
- Orders tab labels the order `Deposit margin`, shows the deposited amount and trigger (never `$0`), and restricts executability states to the deposit-specific yellow timing warning and red insufficient-deposit error; standard increase checks (max leverage, liquidity, acceptable price) are skipped. The order editor edits the trigger in place, shows the amount read-only, and skips acceptable-price rows.
- Trade history maps created/executed/cancelled/updated/frozen to `Create/Execute/Cancel/Update/Failed margin deposit`, showing the collateral delta as size plus the order trigger price (CSV included).
- Order status toasts name the action (`Create/Cancel/Update margin deposit: …`) instead of falling into the zero-size "Depositing…" branch.


qwinndev:
### Code review

Found 1 issue:

1. Margin-deposit orders render as `Limit +$0.00` in the position's orders modal: `OrdersModal` buckets every non-TWAP increase order into its "All" tab without `isMarginDepositOrder` handling, and `TPSLOrdersList` labels any `LimitIncrease` order as `Limit` with the size cell `+{formatUsd(order.sizeDeltaUsd)}` — for the new zero-size margin deposits that is exactly the `$0` Increase presentation FEDEV-2154 forbids ("never presents it as a `$0` Increase"). The modal opens one click from the position row whose inline orders chip this PR already fixed to show "Deposit margin" with the collateral amount, so the same order shows two different identities (bug due to `src/components/OrdersModal/TPSLOrdersList.tsx` rendering `+{formatUsd(order.sizeDeltaUsd)}` for a `sizeDeltaUsd === 0n` order)

https://github.com/gmx-io/gmx-interface/blob/e2b0fa045ef4be308bddf1dcf52c73b6206b5241/src/components/OrdersModal/OrdersModal.tsx#L129-L144

https://github.com/gmx-io/gmx-interface/blob/e2b0fa045ef4be308bddf1dcf52c73b6206b5241/src/components/OrdersModal/TPSLOrdersList.tsx#L158-L177

The fix pattern already exists in this PR in `PositionItemOrders.tsx`:

https://github.com/gmx-io/gmx-interface/blob/e2b0fa045ef4be308bddf1dcf52c73b6206b5241/src/components/PositionItem/PositionItemOrders.tsx#L205-L236

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

timgmxio (dismissed):

timgmxio (approved):

PR #2767, opened 2026-08-19
Add contract-equivalent resulting-position margin check for increase orders FEDEV-4131

timgmxio:
Closes [FEDEV-4131](https://linear.app/gmx-io/issue/FEDEV-4131/bug-align-increase-order-leverage-validation-with-the-contract).

### What

Adds a contract-equivalent **resulting-position margin check** for Market / Limit / Stop increases and wires it into every surface that creates, edits, previews or auto-sizes an increase.

- `sdk/utils/trade/increaseMarginCheck.ts` — port of the gmx-synthetics build deployed on Arbitrum (commit `23c9d160` on `release_2.2.1`, the 2026-07-14 deployment, block 483923342: `release-v2.2c` plus `MarketUtils.getPositivePnl` in the per-position pnl cap; the later `release_2.2.1` changes — positive-pnl haircut, `getImpactUsdForLiquidationCheck`, cumulative capped pnl — are not deployed and not ported): `PositionUtils.isPositionLiquidatable` with `forLiquidation = false` (regular `minCollateralFactor`) applied to the position *after* the increase, plus the post-order gates of `IncreasePositionUtils.increasePosition`: `willPositionCollateralBeSufficient` on the updated open interest, the increase's pending impact, conservative price sides, the full-close fee with the actual balance-improvement tier and `max(referral, pro)` discount, and the contract's reason precedence (`min collateral` → non-positive margin → `min collateral for leverage`). `prbMath.ts` is a bit-exact `exp2 / log2 / mulDivFixedPoint` port so `applyExponentFactor` matches the chain.
- One increase projection per resting order (`makeSelectOrderIncreaseProjection`) feeds the Orders tab, the header badge, the Order Editor and chart-line drags, so the risk state and the next-position values come from the same amounts.
- Surfaces: TradeBox (hard block `Max leverage exceeded` for Market and executable-now Limit/Stop, non-blocking warning for resting orders), Order Editor, Orders tab (red risk state, one deduplicated message, header badge), max-size / slider cap / `Set max leverage`, decoder fallback copy keyed by the contract reason (`minCollateralUsdForLeverage`).
- Resting orders are evaluated at their trigger price: sizing, the margin gate and the slider cap use one evaluation price. "Executable now" is derived from the order type's real trigger rule, not from a mark/liquidation-price comparison.
- The result is independent of the "Include PnL in leverage" display setting. The order-level max-leverage check (FEDEV-3759) is preserved and stays distinguishable in diagnostics.

### Behaviour changes worth QA attention

1. **Limit/Stop increases whose collateral is the index token are valued at the trigger price.** The deposit becomes collateral only at execution, so margin, leverage and liquidation price shown in the trade box, order rows and the Order Editor move to the trigger-price valuation. Example: ETH $2 469, Margin 1 ETH, Limit $1 500, 2× — before ≈ $4 935 of size (≈ 3.3× at execution), now ≈ $2 997 (2× at execution).
2. **Size-driven Margin with an internal swap** (pay ETH, collateral USDC, limit below the market): the reverse swap is solved on the route repriced at the trigger, so the submitted pay amount changes; `Size → Margin → Size` round-trips.
3. `Max leverage exceeded` raised by the new check shows the ticket copy in the CTA tooltip together with `Set max leverage`; the search behind that action respects the new check and sizes limit/stop orders at their trigger price.
4. The pro-tier discount is read from DataStore; while it loads or if the read fails the check runs with no discount instead of waiting or degrading to warning-only.
5. **The traders' pnl cap values the pool pnl on the maximising price side**, as the deployed `getPositivePnl(..., true)` does. On markets with an index spread this slightly lowers the displayed capped pnl when the cap binds, and the check no longer over-allows there. The cap ratio is applied exactly (no 1e18 down-scaling).
6. **Every increase surface keeps the contract reason apart.** The leverage reasons (`min collateral for leverage`, `InsufficientCollateralUsd`) show the max-leverage copy and the `Set max leverage` action; `min collateral` and `< 0` keep the liquidation copy (`Invalid liquidation price`, the liquidation-risk card, the Orders-tab liquidatable message).
7. **One price basis per resting order.** With an index-token collateral the existing collateral is valued at the trigger everywhere: next-position values, the Margin row, the slider's structural bound and the Orders-tab order-level check. The Orders-tab order-level check now reads the shared projection's next leverage instead of re-quoting the swap.
8. **Resting orders are projected along the swap route saved on the order**, not the route the router would pick today. An order whose saved route no longer resolves is not blamed for leverage.
9. **Size-only resting increases (deposit 0)** are projected with their fees drawn from the position's collateral; the pending funding fee is the token amount converted at the evaluation price.
10. **Positions loading window.** While the positions of the current account and chain load (first load, account or network switch) the resulting-position check, the order projections and the Orders-tab increase checks pause instead of evaluating against an empty or another account's position set; the multicall response is tagged with the account and chain it was requested for.
11. **Decoder copy.** `InsufficientCollateralUsd` reads "Max leverage exceeded. Remaining margin $X is below the minimum for the position size…" (the argument is the remaining margin, not a shortfall); the increase copy of the leverage reason of `LiquidatablePosition` is scoped to size increases, deposits and decreases keep the neutral wording.
12. `Set max leverage` runs the whole search before offering the action (no lowest-step probe), and the manual size slider is disabled while no cap can be computed.

### Verification

- Unit: sdk + src specs (parity test = the numbers of the hardhat test "validates collateral amount"; prb-math constants checked against build-info).
- Anvil fork of Arbitrum with frozen prices, contract side = `simulateExecuteOrder` via `eth_call`: 14 scenarios on ETH/USD (long/short, USDC/WETH collateral, market/limit/stop, swaps, Order Editor, Orders tab, slider 100 %) — the UI's first signal matches the contract flip bit-for-bit; where the structural cap binds first the UI is conservative. After the review-round changes: market increase on a losing long, contract rejects from $170 454.78, the trade box blocks between $170 454 and $170 455.
- Fixture replay (no network): six contract verdicts captured on the fork (block 503710836, deployed `PositionUtils 0x0778…` / `IncreasePositionUtils 0x00F3…`, bytecode hashes pinned in the spec) are replayed in `src/domain/synthetics/__tests__/trade/increaseMarginCheck.fixtures.spec.ts` through the trade box's own helpers: the largest size the contract executed passes and the first size it rejected fails with the same reason on all six (losing and profitable long and short, pnl cap on ZEC, OI gate, size-only, WETH collateral on a stop); the saved-route case stays covered by the mocked-pool unit spec, its 57-market fork snapshot was too large to keep in the repo; the sdk boundary lies within $0.005 of the contract bracket.
- Prod (`origin/release`) vs this branch on the same fork, markets where `minCollateralFactor > minCollateralFactorForLiquidation` (the support-case markets): GOLD long with loss — contract rejects from $85 414, branch blocks at $85 414, prod allows up to $123 000; ZEC — contract $59 326, branch $59 326, prod up to $97 000+.

### Support cases this covers

28.07 NATGAS long (market increases reverted in simulation, then liquidated) and 29.07 ZEC long (increases with a swapped pay token plus small top-ups) are this class: prod's second barrier (`Invalid liquidation price`) includes PnL but uses the *liquidation* factor, while the contract validates increases with `minCollateralFactor`. The 08.07 LIT partial close is the **decrease** path and is not covered here.

### Known behaviour

- Trigger-price evaluation is a boundary: a long limit executes at oracle max ≤ trigger, so index-token collateral may be worth slightly less than at "exactly the trigger". The pre-check filters clearly failing orders; it does not guarantee execution.
- If order fees exceed deposit + existing collateral the contract reverts with `InsufficientCollateralAmount` while the pre-check shows `Max leverage exceeded`; both block, only the wording differs (the trade box intercepts earlier with "Fees exceed amount").
- The pre-existing liquidation-price heuristics (`Invalid liquidation price` CTA, "resulting position would be liquidatable at the trigger price" card) fire slightly earlier than the contract check: they use the always-"not improved" closing fee tier (band ≈ 0.02 % of the resulting size in collateral). Conservative, left as is; the new max-leverage state takes precedence when both fire.
- TWAP increases stay out of scope (FEDEV-4134); the manual-leverage slider cap and the `Set max leverage` search still apply the check to the whole TWAP size, which is conservative.
- An order whose saved swap route no longer resolves (a disabled or removed market on the path) shows no error on the Orders tab; that was the case before this PR and is not a leverage problem.
- While the positions of the current account are unavailable (a persistent positions RPC failure) the check is off together with the rest of the position-aware UI; the pre-submit simulation still catches market increases.

### Follow-ups (separate issues)

- Pro discount in the liquidation-price family (`getLiquidationPrice`, decrease estimates, `getMinCollateralUsdForLiquidationPrice`): applied when opening, not when closing. Conservative direction.
- Withdrawal false-allow on markets with `minCollateralFactor > minCollateralFactorForLiquidation` (`getMaxWithdrawAmount` reuses the liquidation-semantics minimum; GOLD: UI allows 372 USDC, contract 294.52) and partial-close remaining-leverage validation (the LIT case) — decrease path.
- Order Editor shows the current, not the projected, liquidation price for increase orders.
- Margin → Size vs Size → Margin asymmetry with accrued borrowing/funding on the existing position (`leverageByCollateral` ignores pending fees) — prod behaviour, product decision needed.


src/domain/synthetics/trade/utils/marginFields.ts
```diff
@@ -92,12 +226,14 @@ export function calcMaxSizeDeltaInUsdByLeverage(params: CalcMaxSizeDeltaParams):
...
     : marketInfo.positionFeeFactorForBalanceWasNotImproved;
 
-  const leverageBoundUsd = bigMath.mulDiv(
+  const structuralBoundUsd = bigMath.mulDiv(
     baseNumerator,
     PRECISION,
     BASIS_POINTS_DIVISOR_BIGINT * PRECISION + leverageBigInt * positionFeeFactor
   );
 
+  const leverageBoundUsd = capSizeDeltaByResultingPositionMargin({ sizeDeltaUsdBound: structuralBoundUsd, params });
```
divhead: 🔴 When the bisection finds no valid size at all (existing position's unrealized loss ≥ collateral + added margin, so remaining stays < minCollateralUsd at any size), this returns `0n` → `toIndexTokenAmount(0n)` → `undefined` → the `if (leverageBound === undefined) return liquidityBound;` below falls through to the **full pool liquidity** as the slider max. Before this PR the fall-through was benign — the structural bound was always > 0, so `undefined` could only mean "no leverage input".

Repro: position 10k size / 4k collateral / 8k unrealized loss, $100 added margin, $100M liquidity → structural bound ≈ 286 ETH, actual result ≈ 83,333 ETH (the liquidity bound), and the manual-margin size slider maps 0–100% onto pool liquidity.

Fix: `if (leverageBoundUsd <= 0n) return undefined;` before the token conversion — callers already treat `undefined` as slider-disabled.

src/domain/synthetics/orders/utils.tsx
```diff
@@ -355,16 +363,19 @@ export function getOrderErrors(p: {
         chainId: p.chainId,
       });
 
-      if (isMaxLeverageError) {
+      const isResultingPositionMarginError = p.resultingPositionMarginState?.isLiquidatable === true;
```
divhead, line 457: ⚠️ This flag is computed from the **raw** position, while the sibling checks in this same block deliberately use `positionForPreview` (set to `undefined` when `isPositionLiquidatedBeforeTrigger`). For a limit whose trigger sits beyond the position's liq price, the raw position's pnl at the trigger is deeply negative by construction, so the margin state ~always fails there → the order shows the "may execute after the current position is liquidated…" warning **and** this red max-leverage error about a position that per the warning won't exist; it also suppresses the `getIsIncreaseResultingPositionLiquidatable` heuristic below via `!isResultingPositionMarginError`.

Same choice in the OrderEditor detect flow (`existingPosition` in the bisection while neighbours use `existingPositionForPreview`) → `numericBinarySearch` can find no valid leverage and toast "No valid leverage available" spuriously.

Suggest applying the `getIsPositionLiquidatedBeforeTrigger` filter to the position inside `getOrderIncreaseResultingPositionMarginState` — it already has the trigger price and the position.

sdk/src/utils/trade/increaseMarginCheck.ts
```diff
@@ -0,0 +1,247 @@
...
+    const totalRebateAmount = applyFactor(closingFeeAmount, userReferralInfo.totalRebateFactor);
+    const discountAmount = applyFactor(totalRebateAmount, userReferralInfo.discountFactor);
+    closingFeeAmount -= discountAmount;
+  }
+
+  const closingCostUsd = convertToUsd(closingFeeAmount, collateralToken.decimals, collateralToken.prices.minPrice)!;
+
+  const remainingCollateralUsd = collateralUsd + positionPnlUsd + totalImpactUsd - closingCostUsd;
+
+  const minCollateralFactor = forLiquidation
```
divhead: ⚠️ The contract cancels an increase via **two** gates, and this ports only one of them. Besides `validatePosition`/`isPositionLiquidatable` (ported here), `IncreasePositionUtils` also runs `willPositionCollateralBeSufficient` → `InsufficientCollateralUsd`: `collateralUsd(min) ≥ max(minCollateralFactor, applyFactor(post-increase OI, minCollateralFactorForOpenInterestMultiplier)) × sizeInUsd`, with no pnl/impact/fees.

The static part is roughly covered by the existing maxAllowedLeverage check, but the OI-scaled part isn't covered anywhere client-side — `getMaxAllowedLeverage` doesn't know the multiplier, and it's non-zero on most prod markets (6e-11…4.44e-8 in gmx-synthetics `config/markets.ts`). On mid/small caps with moderate OI the contract threshold rises above anything the UI checks, and the keeper still cancels with no warning — the class of cancellations this PR targets.

The SDK already has `getMinCollateralFactorForPosition` (used by the decrease flow); worth adding the second predicate here with the projected OI — or explicitly recording it as out of scope.

src/domain/synthetics/orders/utils.tsx
```diff
@@ -452,6 +463,77 @@ export function sortSwapOrders(
...
+    initialCollateralAmount: order.initialCollateralDeltaAmount,
+    indexTokenAmount: convertToTokenAmount(sizeDeltaUsd, marketInfo.indexToken.decimals, triggerPrice),
+    externalSwapQuote: undefined,
+    triggerPrice,
+    limitOrderType: order.orderType as OrderType.LimitIncrease | OrderType.StopIncrease,
+    position,
+    findSwapPath,
+    userReferralInfo,
+    uiFeeFactor: order.uiFeeFactor ?? uiFeeFactor,
+    strategy: "independent",
```
divhead, line 629: ⚠️ The `collateralDeltaAmount` consumed by the margin check comes from `getIncreasePositionAmounts`' `independent` branch, which values collateral as `swapStrategy.amountIn × initialCollateralPrice` — swap fees and swap price impact are never subtracted, and `values.swapUiFeeUsd` is never assigned on that branch (always `0n`; compare `leverageByCollateral`, which uses `amountOut × collateralPrice`).

Pre-existing code, but this PR promotes it into a correctness predicate: for increases whose pay token ≠ collateral token, the check overestimates margin and can pass orders the contract cancels — exactly what this PR is meant to catch. Same applies to the tradebox slider-off path and the OrderEditor.

Suggest switching the `independent` branch to `amountOut` like `leverageByCollateral` does.

sdk/src/utils/trade/increaseMarginCheck.ts
```diff
@@ -0,0 +1,247 @@
...
+
+  const closingFeeFactor = balanceWasImproved
+    ? marketInfo.positionFeeFactorForBalanceWasImproved
+    : marketInfo.positionFeeFactorForBalanceWasNotImproved;
+
+  let closingFeeAmount =
+    convertToTokenAmount(applyFactor(sizeInUsd, closingFeeFactor), collateralToken.decimals, collateralToken.prices.minPrice) ??
+    0n;
+
+  if (userReferralInfo) {
```
divhead, line 117: 💡 The contract subtracts `max(pro-tier discount, referral discount)` here (`PositionPricingUtils.getPositionFees`, the 2.2c pro-traders logic); this subtracts only the referral discount. Conservative direction — fee overstated → remaining understated → slight over-blocking for pro-tier traders near the edge — so low priority, flagging for awareness.

src/context/SyntheticsStateContext/selectors/orderSelectors.ts
```diff
@@ -2,27 +2,67 @@ import {
...
   selectUiFeeFactor,
+  selectUserReferralInfo,
 } from "./globalSelectors";
 import { makeSelectOrderEditorNextPositionValuesForIncrease } from "./orderEditorSelectors";
 import { selectIsSetAcceptablePriceImpactEnabled } from "./settingsSelectors";
 import { makeSelectFindSwapPath } from "./tradeSelectors";
 
 const selectOrdersInfoData = (s: SyntheticsState) => s.globals.ordersInfo.ordersInfoData;
 
+const makeSelectOrderIncreaseResultingPositionMarginState = createSelectorFactory((orderKey: string) =>
```
divhead: 💡 This adds a second full `getIncreasePositionAmounts` (incl. `findSwapPath` for swap-collateral orders) per non-TWAP increase order, invalidated on every `marketsInfoData`/`positionsInfoData` tick — and it runs even with the Orders tab closed, since `selectOrderErrorsCount` feeds the header badge; also once per drag tick on chart order lines. Sub-ms each, so not blocking, but this chain and the existing `nextPositionValues` chain now compute nearly the same amounts twice per order per tick — worth sharing one per-order amounts selector.

src/components/TradeBox/hooks/useTradeButtonState.tsx
```diff
@@ -781,8 +791,31 @@ function useDetectAndSetAvailableMaxLeverage({
...
+            isLong,
+            existingPosition: selectedPosition,
+            sizeDeltaUsd: increaseAmounts.sizeDeltaUsd,
+            sizeDeltaInTokens: increaseAmounts.sizeDeltaInTokens,
+            collateralDeltaAmount: increaseAmounts.collateralDeltaAmount,
+            minCollateralUsd,
+            userReferralInfo,
+            indexPriceForEvaluation: getIncreaseEvaluationIndexPrice({
+              orderType:
+                tradeMode === TradeMode.StopMarket
```
divhead: 💡 This re-derives tradeMode → OrderType inline; the other two tradebox call sites use `increaseAmounts.limitOrderType ?? OrderType.MarketIncrease`, and `increaseAmounts` is already in scope here.

divhead (commented):
Ran a deep pass on this: contract parity vs gmx-synthetics 2.2c, the prb-math port diffed against canonical v2.4.3, plus tscheck and the touched test suites locally.

The core port is solid — verified matching the contracts: exp2 magic table / mulDivFixedPoint rounding / log2 loop bit-for-bit, `applyExponentFactor` semantics, price sides (pnl via `pickPriceForPnl(isLong, false)`, collateral at min, pending impact at min/max by sign), the unconditional `maxPositionImpactFactorForLiquidations` clamp, positive-total-impact zeroing, fee keyed on the close leg's `balanceWasImproved`, reason strings and check order (`"min collateral"` → `"< 0"` → `"min collateral for leverage"`), `LiquidatablePosition` ABI arg indexes, OI + virtual-inventory projection mirroring `applyDeltaToOpenInterest`(`InTokens`), and no double count of the increase impact (contract adds `baseSizeDeltaInTokens` to the position; impact goes only to `pendingImpactAmount`). Local runs: tscheck clean, all 258 touched unit tests green (note the SDK specs only run from the `sdk/` workspace, the root vitest config doesn't pick them up).

**CI is red and both failures are PR-related:**
- lint (`import/order` in `src/domain/synthetics/orders/__tests__/utils.spec.ts`) aborts the "Linting, Type Checking, and Tests" job before tscheck/vitest even run. `--fix` territory.
- 3 Playwright CT tests fail deterministically (TradeBox manual leverage: "slider drives size as a percentage of max size at max leverage" + the two that reuse its first step). Reproduced locally: 50% of max is now 24,850.89 vs the expected 45k–50k. Root cause: on the CT mock market (zero OI, quadratic impact factor 2e-5, `maxPositionImpactFactorForLiquidations` 1%) the new check correctly hits the negative-impact clamp, so max size converges to ≈ collateral/(minCF 1% + liq impact cap 1% + fees) ≈ $49.7k instead of the structural ~$95k. The contract would indeed cancel the bigger order on that market config — the check is right, the CT expectations (or the story market's impact config) need updating.

Specific findings inline. Overall direction is right and the math is trustworthy; the slider fall-through, the liquidated-before-trigger interaction and CI need fixing before merge, and I'd like an explicit decision recorded on the two coverage gaps (OI-scaled factor, swap-collateral optimism).

src/components/TradeBox/hooks/useTradeButtonState.tsx
```diff
@@ -309,6 +313,10 @@ export function useTradeboxButtonState({
 
           break;
         }
+        case ValidationButtonTooltipName.resultingPositionMaxLeverage: {
+          tooltipContent = null;
```
divhead: The old `maxLeverage` error gives the user a way out: its tooltip carries a "Set max leverage" button that calls `detectAndSetAvailableMaxLeverage`. This new case sets `tooltipContent = null`, so when the new check is what blocks the trade the user gets a disabled button with no tooltip and no one-click fix. Only the alert card tells them what to do, and it asks them to fix it by hand.

The binary search behind that button was already updated in this PR to respect the resulting-position margin state, so it would find a valid size for exactly this case. Reusing the same tooltip here (or at least the button) closes the gap.

src/components/OrderEditor/OrderEditor.tsx
```diff
@@ -742,6 +803,20 @@ export function OrderEditor(p: Props) {
       };
     }
 
+    if (isResultingPositionBlocking) {
```
divhead, line 818: Same gap as in the trade box. The `isMaxLeverageError` branch just above offers "Set max leverage" and auto-adjusts the order; this branch only shows text. `detectAndSetAvailableMaxLeverage` in this file already accounts for the new margin state, so the same button would work here.

src/domain/synthetics/trade/utils/marginFields.ts
```diff
@@ -25,16 +32,171 @@ export type CalcMaxSizeDeltaParams = {
...
+      indexPriceForEvaluation: params.indexPriceForEvaluation,
+    });
+
+    if (marginState?.isLiquidatable) {
+      return false;
+    }
+
+    if (params.indexPriceForEvaluation === undefined && resultingExistingPosition !== undefined) {
+      const nextCollateralAmount = resultingExistingPosition.collateralAmount + collateralDeltaAmount;
+      const nextLiqPrice = getLiquidationPrice({
```
divhead, line 146: `getLiquidationPrice` builds its closing fee with `getPositionFee(..., userReferralInfo)` and no pro discount (`sdk/src/utils/positions/utils.ts:211`), while the margin state a few lines above uses the discounted fee. So one predicate mixes two fee models. For a pro-tier trader this guard is stricter than the contract, and the slider max is cut below a size the contract would accept.

The same gap exists in `getDecreasePositionAmounts` and `getMinCollateralUsdForLiquidationPrice`: a pro trader now sees the discount when opening but not when closing. Either pass `proDiscountFactor` into those helpers too, or record it as out of scope.

src/context/SyntheticsStateContext/selectors/globalSelectors.ts
```diff
@@ -20,6 +20,9 @@ export const selectIsGmxAccountBalancesLoaded = (s: SyntheticsState) =>
 export const selectGmMarkets = (s: SyntheticsState) => s.globals.markets.marketsData;
 export const selectUiFeeFactor = (s: SyntheticsState) => s.globals.uiFeeFactor;
 export const selectUserReferralInfo = (s: SyntheticsState) => s.globals.userReferralInfo;
+export const selectProDiscountFactor = (s: SyntheticsState) => s.globals.proDiscountFactor;
+export const selectIsProDiscountFactorReady = (s: SyntheticsState) =>
+  s.globals.account === undefined || s.globals.proDiscountFactor !== undefined;
```
divhead, line 25: If the pro tier / discount multicall never resolves (an RPC failure, for example), `proDiscountFactor` stays `undefined`, `isResultingPositionCheckBlocking` stays false, and the new check quietly degrades to warning-only for the whole session. Failing open is the safer direction for UX, but the feature is then off and nothing in the UI says so.

Falling back to `0n` after a few failed reads would keep the check alive. For non-pro accounts, which is almost everyone, `0n` is the exact value anyway.

sdk/src/utils/markets/utils.ts
```diff
@@ -413,6 +413,53 @@ export function getOpenInterestInTokens(marketInfo: MarketInfo, isLong: boolean)
   return isLong ? marketInfo.longInterestInTokens : marketInfo.shortInterestInTokens;
 }
 
+export function getMarketInfoWithOpenInterestDelta({
```
divhead, line 416: This updates `longInterestUsd` / `longInterestInTokens` but leaves `longInterestUsdUsingLongToken` and `longInterestUsdUsingShortToken` alone, so the returned `MarketInfo` no longer satisfies "total equals the sum of the two legs".

Nothing inside the margin check reads the split fields, so it is fine today. But this is exported as a general helper, and a later caller that feeds the projected market into reserve or max-OI logic would silently read stale numbers. Either project the split fields too, or say in a comment that only the totals are projected.

sdk/src/utils/trade/increase.ts
```diff
@@ -605,11 +981,15 @@ export function getIncreasePositionPrices({
...
   let triggerThresholdType: TriggerThresholdType | undefined;
   let collateralPrice: bigint;
+  let evaluationPrice: bigint | undefined;
 
   if (triggerPrice !== undefined && triggerPrice > 0 && limitOrderType !== undefined) {
     indexPrice = triggerPrice;
+    evaluationPrice = triggerPrice;
     initialCollateralPrice = initialCollateralToken.prices.minPrice;
-    collateralPrice = collateralToken.prices.minPrice;
+    collateralPrice = getIsEquivalentTokens(collateralToken, indexToken)
```
divhead, line 990: This does more than feed the new check. For every limit and stop increase where the collateral token is the index token, the collateral is now valued at the trigger price. That moves the margin, leverage and liquidation price shown on existing screens (trade box, order rows, order editor), and in the size-driven swap path it also moves the pay amount that is actually submitted.

The logic looks right to me: an external swap happens now, so its output is valued at the trigger price, while an internal swap is run by the keeper at the trigger, so the path is re-priced there. The round-trip test covers the sizing. But this is a silent change to numbers users have been reading for a long time, and the PR description does not mention it. Worth writing it down there so QA looks at those screens.

divhead (commented):
Re-review of `3939122f` (11 commits since my pass on Aug 19).

Five of the seven earlier findings are fixed. I re-checked the contract port against gmx-synthetics `release-v2.2c`:

- the new `willPositionCollateralBeSufficient` gate matches. The contract passes `openInterestDelta = 0` because `updateOpenInterest` already ran, so projecting `OI + sizeDeltaUsd` here is equivalent.
- positive price impact is capped by the max factor only, on both legs. The contract says in a comment that it is not capped by the impact pool on an increase, so this is right.
- the closing fee now takes the larger of the pro and referral discounts, like `PositionPricingUtils`.

Local run on this commit: `lint:ci` clean, `tscheck:ci` clean, 1649 root tests and 691 SDK tests green.

**CI is red, but not because of this PR.** Both failing jobs fail on the same test: `sdk/src/configs/__tests__/markets.spec.ts > markets should be consistent with keeper for 42161`. I reproduced it locally — one market in the config (IP/USD) is no longer returned by the keeper. This PR does not touch the markets config, so it is config drift. CI stays red until that fix lands separately.

**Still open from the last round** (details in those threads):
- the duplicated per-order work in `orderSelectors.ts` — this got heavier, not lighter;
- the inline `tradeMode -> OrderType` mapping in `useTradeButtonState.tsx`.

New comments inline. Nothing here is a correctness blocker. The one worth fixing before merge is the lost "Set max leverage" action on the new error.

src/domain/synthetics/trade/utils/marginFields.ts
divhead: Fixed. `if (leverageBoundUsd <= 0n) return undefined` now runs before the token conversion, so an empty bisection result can no longer fall through to the pool liquidity bound.

src/domain/synthetics/orders/utils.tsx
divhead: Fixed. `getOrderIncreaseResultingPositionMarginState` applies the `getIsPositionLiquidatedBeforeTrigger` filter itself and projects from `positionForProjection`. The order editor detect flow uses `existingPositionForPreview`, and the size slider hook moved to `selectTradeboxExistingPositionForPreview`. The warning and the margin check now talk about the same position.

sdk/src/utils/trade/increaseMarginCheck.ts
divhead: Fixed, and I checked the port against `release-v2.2c`.

The contract calls `willPositionCollateralBeSufficient` with `openInterestDelta = 0` because `updateOpenInterest` already ran, so projecting the open interest by `sizeDeltaUsd` here is equivalent. The max with the market-level `minCollateralFactor`, the min collateral price, and the open interest totals (both collateral legs, divided by the market divisor) all match. Tests cover the post-order open interest and the precedence over the leverage reason.

src/domain/synthetics/orders/utils.tsx
divhead: Fixed. The `independent` branch now values the collateral from the swap output, and `swapUiFeeUsd` is assigned there instead of staying `0n`.

The swap is now also re-priced at the trigger for resting orders, which goes further than what I asked for. That part changes numbers outside this check — see my new comment on `getIncreasePositionPrices`.

sdk/src/utils/trade/increaseMarginCheck.ts
divhead: Fixed. The closing fee now takes the larger of the pro and referral discounts, matching the contract.

One leftover: the pro discount is not passed into the decrease and liquidation-price helpers, so it applies when opening but not when closing. Details in my new comment on `marginFields.ts`.

src/context/SyntheticsStateContext/selectors/orderSelectors.ts
divhead: Still open, and it costs more now than when I wrote this.

`getOrderIncreaseResultingPositionMarginState` runs a second `getIncreasePositionAmounts` per increase order, next to the existing `nextPositionValues` chain, and it now also re-prices the swap at the trigger price: one extra `getSwapPathStats` (plus a shallow copy of the whole `marketsInfoData`) in the `independent` strategy, four of them in `leverageBySize`.

It is also computed twice through two separate paths: this selector (order rows and the header badge, on every `marketsInfoData` tick) and the inline block in `makeSelectOrderEditorPositionOrderError`, which `DynamicLines` calls once per frame while an order line is dragged on the chart — through an LRU of 20 that misses on every new trigger price.

One shared per-order amounts selector would remove both the duplication and the drag cost.

src/components/TradeBox/hooks/useTradeButtonState.tsx
divhead: Still open. The `tradeMode -> OrderType` mapping is still inline here, while the new sibling code (`selectTradeboxIsIncreaseExecutableNow`) uses `increaseAmounts.limitOrderType ?? OrderType.MarketIncrease`. `increaseAmounts` is in scope in this callback, so the same one-liner works.

divhead (commented):
Re-review on `3939122`. Read the code at HEAD rather than going by the commit messages; did not re-run the suites locally this round.

**Closed since the 19 Aug pass — verified in the code (5 of 7):**

- 🔴 slider falling through to full pool liquidity — fixed exactly as suggested: `if (leverageBoundUsd <= 0n) return undefined;` now sits before the token conversion in `calcMaxSizeDeltaInUsdByLeverage`.
- ⚠️ raw position instead of `positionForPreview` — fixed: `getOrderIncreaseResultingPositionMarginState` derives `positionForProjection = isPositionLiquidatedBeforeTrigger ? undefined : position`, and resting orders are now evaluated at the trigger price via `indexPriceForEvaluation`.
- ⚠️ missing `willPositionCollateralBeSufficient` gate — ported: `max(minCollateralFactor, applyFactor(nextOI, multiplier)) × nextSizeInUsd`, with the new `InsufficientCollateralUsd` reason and the `getIsMaxLeverageMarginReason` helper.
- ⚠️ swap collateral valued at `amountIn` — fixed: the `independent` branch now goes through `getBaseCollateralUsdFromSwapStrategy` (i.e. `amountOut`), and `swapUiFeeUsd` is actually computed there instead of staying `0n`.
- 💡 pro-tier discount — fixed: `max(referral, pro)` in `getResultingPositionMarginState`.

**Still open, both minor — fine to split into a separate issue:**

- `selectOrderErrorsCount` still runs a second full `getIncreasePositionAmounts` per increase order per tick, with the Orders tab closed as well.
- `useTradeButtonState.tsx:811` still re-derives `OrderType` from `tradeMode` inline while `increaseAmounts.limitOrderType` is in scope.

**CI is red, but not because of this PR.** Both failing jobs fail on the same test — `sdk/src/configs/__tests__/markets.spec.ts` → "markets should be consistent with keeper for 42161" (15s timeout in one job, `expected undefined to be defined` in the other). That is the delisted IP/USD market. The fix is already on `release` (`167ec0ff`, PRs #2794 / #2797, merged 27 Aug 08:48); the last run here is from 26 Aug 18:34, and the branch is 55 commits behind `release`. Merging `release` in should turn CI green — no code change needed, and reverting 90b76e0f was the right call.

**Two things before I can approve:**

1. Merge `release` and get a green run.
2. The seven inline threads from the 19 Aug pass have no replies. The fixes are in the code, but the decisions behind them are not recorded anywhere — please answer in the threads, even in one line, so the reasoning is on the record instead of reconstructed from commit messages.

Also worth filling in the PR description: it is empty right now, so there are no requirements or acceptance criteria to review the change against.

src/domain/synthetics/trade/utils/marginFields.ts
timgmxio: Fixed: `calcMaxSizeDeltaInUsdByLeverage` returns `undefined` when the resulting-position cap leaves no positive size, so the slider is disabled instead of falling through to the pool liquidity. Covered in `marginFields.spec`.

src/domain/synthetics/orders/utils.tsx
timgmxio: Fixed: `getOrderIncreaseResultingPositionMarginState` applies `getIsPositionLiquidatedBeforeTrigger` itself and projects from the same position the neighbouring checks use; the editor's search and the slider hook read the preview position too.

sdk/src/utils/trade/increaseMarginCheck.ts
timgmxio: Ported: `willPositionCollateralBeSufficient` runs on the projected open interest (`getMarketInfoWithOpenInterestDelta`) before `validatePosition`; its `InsufficientCollateralUsd` reason maps to the max-leverage copy. Decision: in scope, it is the second cancel gate of the same increase path.

src/domain/synthetics/orders/utils.tsx
timgmxio: Fixed: the `independent` branch values the deposit from the swap output and assigns `swapUiFeeUsd`. Resting orders also reprice the internal swap at the trigger, which is what the size-driven swap fix in the description describes.

sdk/src/utils/trade/increaseMarginCheck.ts
timgmxio: Fixed: the closing fee takes `max(referral, pro)` like `PositionPricingUtils`; `proDiscountFactor` comes from `useProDiscountFactorRequest` and is threaded through the check, the slider cap and both searches.

src/context/SyntheticsStateContext/selectors/orderSelectors.ts
timgmxio: Keeping it as is in this PR. One shared amounts selector means unifying inputs the two chains currently disagree on (the order's own `uiFeeFactor`, a stop treated as a limit, the swap route of a fresh position), which changes the editor's displayed next-values and needs its own verification. Cost today is sub-ms per order per markets tick and about 1 ms per drag frame. Split into a follow-up, listed in the description.

src/components/TradeBox/hooks/useTradeButtonState.tsx
timgmxio: Done: the mapping lives in `getLimitOrderTypeByTradeMode` (shared with `makeSelectIncreasePositionAmounts`), both searches pass `limitOrderType` to `getIncreasePositionAmounts`, and the evaluation price uses `increaseAmounts.limitOrderType ?? OrderType.MarketIncrease`. This closed a real gap: the search sized limit/stop orders at the mark price while the margin state was evaluated at the trigger. Spec: the found usd notional is the same at trigger 2000 and 1800.

src/components/TradeBox/hooks/useTradeButtonState.tsx
timgmxio: Added: the `resultingPositionMaxLeverage` tooltip now carries the ticket copy and the same `Set max leverage` action; the search behind it already respects the margin state. Checked on an anvil fork with a losing long: blocked at $178 978, after the action the button is `Increase Long`.

src/components/OrderEditor/OrderEditor.tsx
timgmxio: Added the same tooltip and action to the editor branch. Checked on the fork with an executable-now limit: size 400 000 blocked, the action sets 46 521.

src/domain/synthetics/trade/utils/marginFields.ts
timgmxio: Out of scope here. The slider guard mirrors the `Invalid liquidation price` validation, which uses the same undiscounted closing fee, so the cap stays within validation. Making the liquidation-price family pro-aware touches eight call sites including the positions list and decrease, and would desync the trade-box preview from the position row. Conservative direction; tracked as a follow-up in the description.

src/context/SyntheticsStateContext/selectors/globalSelectors.ts
timgmxio: Done: `useProDiscountFactorRequest` returns `0n` when either DataStore read errors (exact for non-pro accounts, conservative for pro) and stays `undefined` only while loading; SWR keeps retrying. Unit-tested.

sdk/src/utils/markets/utils.ts
timgmxio: Done: the helper takes the collateral token and moves the delta into that token's leg in usd and in tokens, as the contract's `openInterestKey` does, so the totals stay the sum of the legs. Unit tests include `getPositiveMarketPnl` consistency.

sdk/src/utils/trade/increase.ts
timgmxio: Agreed. This was already written up in the task (Linear comment of 26 Aug, "Two bugs fixed on the way") and is now in the PR description under behaviour changes for QA.

src/context/SyntheticsStateContext/selectors/orderSelectors.ts
timgmxio: Fixed in 5f9830345: `makeSelectOrderIncreaseProjection` computes a resting order's increase amounts once per order, trigger and size, and the margin state, the next-position values, the Orders tab, the header badge, the Order Editor and chart-line drags all read from it; the order's own `uiFeeFactor`, order type and position filter now apply to every consumer. Checked on the fork: the editor's warning line matches the contract flip for a limit (182 307.39 vs 182 307.49), a stop (390 640.73 vs 390 640.72) and a swap-collateral limit on the order's route (182 387.91 vs 182 387.91); the Orders tab flags only the order beyond its line.

src/components/TradeBox/hooks/useTradeButtonState.tsx
```diff
@@ -298,14 +308,25 @@ export function useTradeboxButtonState({
...
-              <EmbeddedActionButton onClick={detectAndSetAvailableMaxLeverage}>
-                <Trans>Set max leverage</Trans>
-              </EmbeddedActionButton>
+              {setMaxLeverageButton}
             </>
           );
 
           break;
         }
+        case ValidationButtonTooltipName.resultingPositionMaxLeverage: {
```
divhead, line 317: Follow-up to my comment about the missing action: the button is now always rendered, so when no positive increase can pass, clicking it only shows "No available leverage found". For example: an existing long whose loss has already taken it below `minCollateralFactor`, with a small pay amount. The DoD says: "No actionable size or Set max leverage action is offered when no positive increase can pass."

I'd run the search first and render the button only when it finds a size. The Order Editor has its own copy of the search (`detectAndSetAvailableMaxLeverage` in `OrderEditor.tsx`), so extracting one shared search function would give both places the same gating.

src/components/OrderEditor/OrderEditor.tsx
```diff
@@ -734,10 +808,27 @@ export function OrderEditor(p: Props) {
...
+      return {
+        text: isResultingPositionMaxLeverageError ? t`Max leverage exceeded` : t`Invalid liquidation price`,
+        tooltip: isResultingPositionMaxLeverageError ? (
+          <>
+            <Trans>
+              The resulting position would exceed the maximum allowed leverage. Increase margin or reduce size.
+            </Trans>
+            <br />
+            <br />
+            {setMaxLeverageButton}
```
divhead, line 828: Same here: when the editor's search finds nothing, the action ends in "No valid leverage available".

src/context/SyntheticsStateContext/selectors/globalSelectors.ts
```diff
@@ -20,6 +20,9 @@ export const selectIsGmxAccountBalancesLoaded = (s: SyntheticsState) =>
 export const selectGmMarkets = (s: SyntheticsState) => s.globals.markets.marketsData;
 export const selectUiFeeFactor = (s: SyntheticsState) => s.globals.uiFeeFactor;
 export const selectUserReferralInfo = (s: SyntheticsState) => s.globals.userReferralInfo;
+export const selectProDiscountFactor = (s: SyntheticsState) => s.globals.proDiscountFactor;
+export const selectIsProDiscountFactorReady = (s: SyntheticsState) =>
```
divhead, line 24: While the pro tier is loading (wallet connect, account switch), this is `false`, so the resulting-position check doesn't block. For Market mode the alert card doesn't show either. Until the first multicall returns, a failing market increase is neither blocked nor warned. The pre-submit simulation still catches it, so this is minor. Evaluating with `0n` while loading would close the gap: it's exact for non-pro accounts and stricter for pro.

divhead (commented):
Re-review of `0ecc311`.

Everything from the last round is addressed. Both `release` merges look right; I checked the manual conflict resolution in `OrderEditor`, `useTradeButtonState` and `orders/utils`.

One thing is left: the Set max leverage action is offered even when no size can pass, and the DoD rules that out (inline). The loading-window comment is minor.

CI is still red on the SDK keeper consistency test. Its fix landed on `release` in #2831, after your last merge, so one more `release` merge should turn it green.

Heads-up: this PR and #2830 conflict in all 10 `.po` catalogs, whichever merges first. The one that merges second needs a catalog re-extract.

src/components/TradeBox/hooks/useTradeButtonState.tsx
timgmxio: Done in 963fcc4f45: the search runs first (`findMaxLeverageIncrease` in `domain/synthetics/trade/utils/maxLeverageSearch.ts`, shared by the trade box and the editor) and the action is rendered only when it returns a size. Follow-up in c6de305e55: the probe on the lowest step is gone, the whole binary search runs, and a step whose increase has a zero size never counts as passing, so the button cannot land on an empty result.

src/components/OrderEditor/OrderEditor.tsx
timgmxio: Same in the editor: `hasAvailableMaxLeverage` is the shared search result (963fcc4f45, c6de305e55), so the action is rendered only when the search found a size.

src/context/SyntheticsStateContext/selectors/globalSelectors.ts
timgmxio: Done in 0f77ae06d3: while the tier loads `proDiscountFactor` is undefined and the check evaluates with no discount (exact for non-pro accounts, stricter for pro); `isResultingPositionCheckBlocking` no longer waits for the read. The positions window got the same treatment in 8e52710ad9: the multicall response is tagged with the account and chain it was requested for, and the trade-box margin state, the order projections and the Orders-tab increase checks pause until the current account's positions are loaded, so a top-up is never judged as a fresh position during a wallet connect or an account switch.

timgmxio:
CI note: all four jobs on caa775e56 fail at startup with the annotation "The job was not started because your account is locked due to a billing issue"; every run in the repo since ~09:17 UTC today fails the same way on every branch, so this is not the branch. The same gate ran locally on this commit: `yarn tscheck` (root, sdk), `yarn lint:ci` (root, sdk), `yarn vitest run` (root 1975 passed, sdk 725 passed).

divhead (approved):

PR #2801, opened 2026-08-27
Make interface aware of installed app mode FEDEV-4177

qwinndev:
## Summary

- Detect installed PWA sessions across standalone display mode and the iOS home-screen fallback.
- Expose installed mode on the document before React renders.
- Apply installed-only safe-area insets to the app shell and viewport-fixed controls.

Linear: [FEDEV-4177](https://linear.app/gmx-io/issue/FEDEV-4177/make-the-interface-aware-it-is-running-as-an-installed-app)


timgmxio (changes requested):
the safe-area handling is not installed-only, it runs in browser tabs too.

`viewport-fit=cover` in `index.html` and the `env(safe-area-inset-*)` padding in `Shared.scss` are unconditional, and nothing reads the `data-installed-app` attribute (no selector for it in the built CSS). Reproduced in a Safari tab on an iPhone 16 Pro Max in landscape: the branch preview shows an extra strip under the Long/Short/Swap tabs (the curtain's safe-area bottom inset) that prod doesn't have. On base and Pro iPhones the widened viewport also crosses the `md` (768px) breakpoint, so the tab gets the desktop layout. Both break the "layout in a browser tab is unchanged" AC.

qwinndev:
> the safe-area handling is not installed-only, it runs in browser tabs too.
> 
> `viewport-fit=cover` in `index.html` and the `env(safe-area-inset-*)` padding in `Shared.scss` are unconditional, and nothing reads the `data-installed-app` attribute (no selector for it in the built CSS). Reproduced in a Safari tab on an iPhone 16 Pro Max in landscape: the branch preview shows an extra strip under the Long/Short/Swap tabs (the curtain's safe-area bottom inset) that prod doesn't have. On base and Pro iPhones the widened viewport also crosses the `md` (768px) breakpoint, so the tab gets the desktop layout. Both break the "layout in a browser tab is unchanged" AC.

fixed

timgmxio (dismissed):

timgmxio (approved):

PR #2802, opened 2026-08-27
Make installed app versions and caching deterministic FEDEV-4178

qwinndev:
## Summary

- Derive the installed app generation from the release commit timestamp so rebuilds of the same commit use the same version while later releases remain ordered.
- Serve the service worker with `Cache-Control: no-store` so updates are not held behind the CDN or browser cache.
- Remove the landing manifest link and the stale manifest named "App" so only the app site advertises the canonical GMX manifest.

Linear: [FEDEV-4178](https://linear.app/gmx-io/issue/FEDEV-4178/make-the-installed-apps-version-deterministic-and-its-caching-explicit)


landing/index.html
```diff
@@ -29,7 +29,7 @@
     <link rel="icon" type="image/png" sizes="32x32" href="/favicon/favicon-32x32.png" />
     <link rel="icon" type="image/png" sizes="96x96" href="/favicon/favicon-96x96.png" />
     <link rel="icon" type="image/png" sizes="16x16" href="/favicon/favicon-16x16.png" />
-    <link rel="manifest" href="/favicon/manifest.json" />
+    <link rel="manifest" href="/manifest.json" />
```
timgmxio: Linking `/manifest.json` from `landing/index.html` makes the landing itself installable: `display: standalone` plus `start_url`/`scope` resolving against `gmx.io` means Chrome will offer to install "GMX" on the landing, and that install opens the landing, not the app ("Launch App" then leaves the scope into an in-app browser). The old manifest was not installable, so this is new behaviour, and users can end up with two "GMX" icons.

The AC only asks for one definition with the GMX name, which is also satisfied by dropping `<link rel="manifest">` from the landing altogether.

qwinndev: agree, fixed

timgmxio (approved):

PR #2810, opened 2026-08-31
Hotfix: avoid Safari home rendering stall FEDEV-4269

qwinndev:
## Summary

- replace the home hero's large Tailwind `blur(180px)` and `blur(46px)` layers with equivalent radial gradients
- avoid a known Safari/WebKit first-frame rendering regression while preserving the purple hero glow
- leave `total_volume`, `position_stats`, and `apy` unchanged because they load after mount and do not gate the initial page render

The landing hero matched WebKit reports where large CSS blur filters—especially Tailwind's custom-property filter chain—leave the first frame blank for roughly 15–20 seconds on real Safari. The main app does not render this hero, which explains why the problem is isolated to the home app.

WebKit references:

- https://bugs.webkit.org/show_bug.cgi?id=319187
- https://bugs.webkit.org/show_bug.cgi?id=316921
- https://bugs.webkit.org/show_bug.cgi?id=322045

Linear: https://linear.app/gmx-io/issue/FEDEV-4269/safari-mobile-browser-gmxio-page-takes-25-seconds-to-load-and-shown

### DIFF
--- a/landing/src/pages/Home/HeroSection/HeroBackground.tsx
+++ b/landing/src/pages/Home/HeroSection/HeroBackground.tsx
@@ -1,19 +1,15 @@
-import heroBg from "img/bg_hero_3d.png";
+const heroGlowStyle = {
+  background: `
+    radial-gradient(ellipse 42% 72% at 72% 45%, rgba(63, 38, 196, 0.75), rgba(31, 20, 104, 0.48) 48%, rgba(9, 10, 20, 0) 78%),
+    radial-gradient(ellipse 24% 42% at 61% 30%, rgba(83, 49, 244, 0.3), rgba(9, 10, 20, 0) 72%)
+  `,
+};
 
-const styleFirstImage = {
-  background: `linear-gradient(233deg, rgba(9, 10, 20, 0.00) 1.13%, #090A14 82.68%), url(${heroBg}) #090A1400 -1876.468px -923.827px / 423.762% 461.839%`,
-};
-const styleSecondImage = {
-  background: `linear-gradient(233deg, rgba(9, 10, 20, 0.00) 1.13%, #090A14 82.68%), url(${heroBg}) #090A1400 -2378.856px -1078.792px / 423.762% 461.839%`,
-};
 export function HeroBackground() {
   return (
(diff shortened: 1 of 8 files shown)


timgmxio (approved):

divhead (commented):
Reviewed the diff and checked the diagnosis against the sources.

**The root cause holds up.** All three WebKit reports are real and on point, and the preconditions are present here:

- 316921 (duplicate of 316769) is specifically Tailwind v3's `var()`-chained `filter` plus a large blur radius — this repo is on `tailwindcss` 3.4.4.
- 322045 needs painted inline SVG in the first frame; the hero renders `IcLinkArrow` and `IcMidChevron` through `?react`. The reporter there shipped the same workaround, blur replaced by a gradient.

**Visual: not quite "equivalent".** Compared the branch preview against production at 375x812 and 1440x900. Mobile is essentially indistinguishable. On desktop the glow is visibly larger and softer, shifted down and to the left, and it loses the texture the PNG gave it. Not a code problem, but it needs a nod from design before this ships.

**The blocking question — how does this sit with #2807?** #2807 shipped for the same issue two days ago and disables multicall workers on all iOS and iPadOS, permanently. This PR's own description says the app does not render the hero and the problem is isolated to the home app, which implies #2807 did not address the reported symptom. If the hero is the real cause, #2807 is a standing performance regression for every iOS user. We should either revert it or justify it on its own evidence — please make that call explicitly rather than leaving both in.

**There is no measurement.** No before/after first-paint number, no device, no iOS version, no Testing section (#2807 had one). For a performance hotfix that is the entire claim. Both preview URLs are live; a cold-load FCP on iOS 26 Safari, before and after, would settle it.

**Partial by design?** `ProtectionBackground` (`blur-[46px]`) and `SocialBackground` (`blur-[180px]` plus `blur-[46px]`) still use the same pattern on the same image. Both sit below the first viewport, and per the 322045 repro below-the-fold blurred elements do not stall the first frame, so first paint is probably fine. They can still stall on scroll and compositing. Worth a follow-up issue rather than widening this PR.

**Branch:** this targets `master`, and `master` and `release` have diverged (+14 / -10). Needs the usual sync so the fix survives the next promotion.

Happy to approve once there is a measured before/after and a decision on #2807.

PR #2813, opened 2026-09-01
Restore path-based URLs and fix metamask mobile issue FEDEV-4237

qwinndev:
Restores path-based URLs on `master` together with the fixes that make wallet connect reliable in in-app mobile browsers — the combination that previous re-land attempts were missing. Rebuilt as four clean commits on current master; all debugging aids used during the investigation are excluded from the ship.

Linear: https://linear.app/gmx-io/issue/FEDEV-4237/fix-wallet-connect-in-metamask-in-app-browser-broken-by-path-based

## What this ships

1. **Re-land of path-based URLs** — revert of revert #2782 (`git revert 6c2bdcd84e`), restoring the #2760 tree: in-place `history.replaceState` legacy-hash rewrites (no document navigation mid-connect) and the `privy_oauth_*` leave-alone guard. Master already carries the #2783 trade-params loop fix; it stays intact.
2. **Privy hCaptcha CSP allowlist** (`public/_headers`) — Privy's bot protection now injects `js.hcaptcha.com/1/api.js` into the host page during login; our CSP only allowed the old Turnstile host, so the captcha script was blocked and the connect promise hung with no error whenever a captcha was required. Allows `hcaptcha.com` + `*.hcaptcha.com` in `script-src`/`style-src`/`frame-src`; Turnstile hosts stay.
3. **Edge redirect for the app root** (`public/_redirects`, single exact rule `/ /trade 302`) — the root path redirects at the CDN before any JavaScript runs, replacing the client-side redirect that executed mid-boot. Preview-verified: query strings preserved (`/?ref=X` → `/trade?ref=X`), hashed assets unaffected.
4. **Late injected-provider rediscovery** (`watchInjectedProviderAnnouncements`, wired in `WalletProvider`) — in-app browsers (MetaMask iOS) can inject their provider after the app's wallet scan; re-dispatching `eip6963:requestProvider` at boot, on short delays, and on `ethereum#initialized` makes the wallet discoverable regardless of timing. Announcements are idempotent. Unit-specced.

## Testing

- On-device: wallet connect verified from MetaMask iOS in-app browser on the preview build carrying these exact fixes (bare URL → instant `/trade`, connect completes).
- Scripted probe (mock injected MetaMask completing real SIWE against the live Privy app) passes on the preview under Chromium desktop, Chromium + MetaMask-iOS UA, and WebKit + iPhone profile; a stale-service-worker upgrade across deployments was also simulated with a persistent profile — old shell serves stale-but-functional, new SW takes over on the next load, no reload loop.
- `tscheck`, eslint/prettier on changed files, and the new unit spec green locally; full unit/CT/SDK matrix runs in this PR's CI.
- Note for future device tests: always use the per-commit preview URL from the deploy comment — branch-alias origins accumulate per-device SW/cache state across pushes, and Cloudflare keeps orphaned deployments alive at their hash URLs.


qwinndev:
**Probe verification** (scripted mock injected MetaMask completing real SIWE against the live Privy app):

Before — prod `app.gmx.io/#/trade` and the pre-fix preview behave identically:
```
>> GET js.hcaptcha.com/1/api.js
XX FAILED js.hcaptcha.com/1/api.js  csp        (×3)
VERDICT: STUCK — connect promise never settles
```

After — preview rebuilt with the CSP change (`f1389a76.gmx-interface.pages.dev`):
```
<< 200 js.hcaptcha.com/1/api.js
<< 200 newassets.hcaptcha.com/.../hcaptcha.html   (challenge frames)
<< 200 api.hcaptcha.com/checksiteconfig
<< 200 api.hcaptcha.com/getcaptcha/b9fc5a50-...
```
The captcha pipeline runs end to end; the headless probe is then (correctly) served a challenge it cannot solve — hCaptcha doing its job against automation. A real user gets the invisible/interactive challenge inside the Privy modal instead of an infinite spinner. The served `script-src` on the rebuilt preview now carries `https://hcaptcha.com https://*.hcaptcha.com`.

Remaining on-device check (MetaMask iOS in-app browser, add `?walletDebug=1`): the overlay should show `CHALLENGE script/iframe loaded` and `auth.privy.io` siwe calls settling. Note the prod Privy app's captcha toggle was verified OFF on 2026-08-21 and is ON again today (`hcaptcha`) — flipping it off in the Privy dashboard remains the zero-deploy mitigation, and it is worth finding out who re-enabled it.

qwinndev:
**Probe verification on this branch's preview** (`4429ac2e.gmx-interface.pages.dev`, commit `cf2d2b8` = re-land + CSP fix):

- Deep path `/trade` loads directly under path routing (Cloudflare Pages SPA fallback + BrowserRouter).
- Captcha pipeline runs end to end — `js.hcaptcha.com/1/api.js` 200 (previously blocked by CSP), challenge frames load, `api.hcaptcha.com/checksiteconfig` 200, `getcaptcha` 200. The headless probe is then served a challenge it cannot solve, which is hCaptcha correctly rejecting automation; a real user gets the invisible/interactive challenge inside the Privy modal instead of an infinite spinner.
- Legacy `/#/trade?chainId=42161` normalizes in place to `/trade?chainId=42161`, followed by a single trade-params cleanup to `/trade` — one rewrite, no `replaceState` storm (the #2783 guard holding).

Remaining before merge: on-device connect from the MetaMask iOS in-app browser on this preview with `?walletDebug=1` (overlay shows challenge load + siwe outcomes). Zero-deploy prod mitigation meanwhile: flip the captcha toggle off in the Privy dashboard — and check its audit log for who re-enabled it after 2026-08-21.

qwinndev:
**Deploy-safety verification: stale service worker upgrade across a deployment.**

Simulated a returning PWA user with a persistent MetaMask-iOS-UA profile against this branch's preview alias: installed the SW on build `1788256339934`, pushed a new deployment (`1788266850409`), revisited with the stale profile.

- First load after the deploy: old shell served from `gmx-pwa-shell-v2-1788256339934` (app renders and works), new SW installs and takes control in the background, new shell cache populated.
- Next load: new build served, both cache generations present, app healthy.

So the prod deploy of this PR follows the designed stale-while-revalidate path for returning users — one functional stale load, then the new build. No reload loop, no blank screen.

On-device status: MetaMask iOS connect **passed** on the pristine per-deployment preview (`4429ac2e...pages.dev`) with captcha disabled. A device that visited the branch *alias* during the earlier captcha-hang window can carry a wedged per-origin SW/cache state — clearing MetaMask's browser data resets it; per-deployment hash URLs are the reliable origins for device testing.

qwinndev:
Two additions after on-device reports that the bare origin (`/`) hangs on real phones while `/trade?walletDebug=1` works — the two URL differences are eliminated outright rather than requiring users to know magic links:

- **`public/_redirects` — `/ /trade 302`**: the root path now redirects at the CDN edge, before any JavaScript runs, replacing the client-side `<RedirectWithQuery to="/trade">` that executed mid-boot. Verified on the preview: `302 location: /trade`, query strings preserved (`/?ref=TESTCODE` → `/trade?ref=TESTCODE`), hashed assets still served as JS (the single exact-path rule does not trigger the SPA-fallback asset-swallowing issue a catch-all rule would).
- **`watchInjectedProviderAnnouncements`** (wired in `WalletProvider`): re-dispatches `eip6963:requestProvider` at boot, on short delays, and on `ethereum#initialized`. In-app browsers (MetaMask iOS) can inject their provider after the app's wallet scan; re-requesting announcements makes the injected wallet discoverable regardless of timing. This is the one functional side effect the `?walletDebug=1` overlay had — done deliberately. Announcements are idempotent (stores dedupe by uuid). Covered by `announceInjectedProviders.spec.ts`.

public/_redirects
```diff
@@ -0,0 +1 @@
+/ /trade 302
```
timgmxio: This file also ships with the landing build: `vite.landing.config.ts` reuses `publicDir: public`, so `build/_redirects` ends up on gmx.io as well.

On the landing preview (0ab3b040.gmx-interface-home.pages.dev) `GET /` now answers `302 /trade` (Googlebot UA included), the landing loads at `/trade`, and the `*` route in `LandingRoutes` bounces back to `/` via `history.replace`. Production gmx.io/ is a 200 today. Net effect: the marketing root becomes a redirect plus a client-side bounce, and with no canonical tag on the landing Google would treat `/trade` as the homepage's location. Referral links (`/?ref=X`) and legacy `/#/referral-terms` still work, so this is about the root URL only.

Could we keep the rule out of the landing build? Options:
- strip it in `vite.landing.config.ts`, e.g. a `closeBundle` hook that removes `build/_redirects`;
- or emit `_redirects` from the app build only instead of the shared `public/`;
- or drop the file: `MainRoutes` still redirects `/` → `/trade` client-side.

timgmxio (dismissed):

public/_redirects
qwinndev: fixed

timgmxio (approved):

PR #2814, opened 2026-09-01
Point subsquid urls at the :prod tag

divhead:
The UI is pinned to squid slot `@8abb92` and the SDK codegen / v1 test util to `@c9407b`. Those pins are what keep the extra squids alive, so the SQD project cannot be cleaned up while they stand. This points everything at the `:prod` tag.

**Runtime impact: none.** `:prod` currently resolves to the exact slot the UI was pinned to — identical introspection hash and identical `squidStatus` heights across repeated samples, while `@7cff6d` and `@c9407b` differ on both. The pin was already redundant; it only meant the UI would stay behind the next `:prod` promotion.

**Checked before opening**

- `arbitrum`, `avalanche`, `megaeth`, `arb-sepolia` all answer `squidStatus` over `:prod` with HTTP 200.
- Introspection diff `:prod` vs `@c9407b` (the SDK codegen slot): additive only, 440 new fields, nothing removed. Generated subsquid types stay valid, so no regeneration is needed in this PR.
- The deployed bundle on app.gmx.io was hitting `@8abb92`, which is how the pin was found.

**Notes**

- `sdk/README.md` pointed at `@bac941`, which 404s — that slot is already deleted. Updated to `:prod` as well.
- Fuji already used `:prod` and is untouched. Separately: `gmx-synthetics-fuji:prod` currently 404s, the fuji squid is gone.
- Targeted at `master` on purpose — the point is to free the extra slots now. `release` carries the same pin and needs the same change so the next release does not reintroduce it.

### DIFF
--- a/sdk/README.md
+++ b/sdk/README.md
@@ -18,7 +18,7 @@ const sdk = new GmxSdk({
   rpcUrl: "https://arb1.arbitrum.io/rpc",
   oracleUrl: "https://arbitrum-api.gmxinfra.io",
   walletClient: useWallet().walletClient,
-  subsquidUrl: "https://gmx.squids.live/gmx-synthetics-arbitrum@bac941/api/graphql",
+  subsquidUrl: "https://gmx.squids.live/gmx-synthetics-arbitrum:prod/api/graphql",
 });
 
 const { marketsInfoData, tokensData } = await sdk.markets.getMarketsInfo();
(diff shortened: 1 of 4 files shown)


divhead:
Superseded by #2816 — master only accepts PRs from hotfix-*/release-* branches with a matching title.

PR #2827, opened 2026-09-07
Stop reporting delisted-market price gaps as oracle keeper failures

timgmxio:
Tickers partial-data detection now forgets a token missing for longer than PRICES_CACHE_TTL and does not count tokens used only by delisting markets, so a delisting no longer bans and rotates every oracle keeper endpoint.

missedMarketPrices is pushed once per missing-prices episode per market and source instead of on every prices tick, and skips delisting markets.


divhead (commented):
Both items from the ticket are implemented. Two things I'd sort out before approving, details inline:

1. After a gap in polling longer than the TTL, the cache forgets tokens that the current endpoint omits, and a partial endpoint can then stay primary for the rest of the session. The old code banned it within a few seconds, so this is a regression in failover.
2. `DELISTING_MARKET_ADDRESSES` doesn't match what the keeper actually stopped pricing, in both directions.

CI is red only because of the SDK keeper consistency test (`markets should be consistent with keeper for 42161`), which was fixed on `release` in #2831. Merging `release` should turn it green.

src/domain/synthetics/tokens/tokenPricesCache.ts
```diff
@@ -0,0 +1,78 @@
...
+      cache.prices[address] = pricesData[address];
+      cache.updatedAt[address] = now;
+    }
+
+    for (const address of Object.keys(cache.updatedAt)) {
+      if (pricesData[address]) {
+        continue;
+      }
+
+      const isFresh = now - cache.updatedAt[address] < ttl;
```
divhead: Forgetting a token by the time since it was last received turns partial-data detection off after any polling gap longer than the TTL. Such gaps are normal: `/leaderboard` and `/competitions` poll tickers every 60s against a 30s TTL, hidden tabs get throttled to about once a minute, and laptops sleep.

Scenario: during a gap the primary becomes an endpoint that omits an active token X. Its first response finds X older than the TTL and drops it without reporting. From then on X is no longer expected, so `unexpectedMissingAddresses` stays empty, `handleFailure` is never called, and nothing moves away from that endpoint (the tracker's probe only checks that tickers are non-empty). X's markets stay without prices, and on the RPC path `usePositions` skips the user's positions in them. With the old code the same endpoint was banned about 4s after 1s polling resumed.

The TTL now controls two different things: how long a cached price may be served, and which tokens we expect to see. I'd keep the 30s for the first, and for the second forget a token only after it has also been missing for N consecutive responses (or once a second endpoint is missing it too).

src/config/markets.ts
```diff
@@ -190,3 +191,51 @@ export function isGmxAccountHoldableToken(chainId: SettlementChainId, tokenAddre
...
+  shortTokenAddress: string;
+};
+
+const DELISTING_ONLY_TOKEN_ADDRESSES: Record<number, Set<string>> = {};
+
+/**
+ * Tokens used by delisting markets only. The oracle keeper stops serving their prices once the market
+ * is delisted, so their absence in tickers is expected and must not count as a partial response.
+ */
+export function getDelistingOnlyTokenAddresses(chainId: number): Set<string> {
```
divhead: `DELISTING_MARKET_ADDRESSES` is the list behind the delisting announcements, and it doesn't match what the oracle keeper has actually stopped pricing. I checked the live Arbitrum tickers on both endpoints:

- OM/USD and WELL/USD are no longer served (`SKIPPED_KEEPER_MARKETS` in the SDK markets test already skips them), but they are `enabled: true` and not on the list. They keep reporting `missedMarketPrices` (now once per session and source), so they are the part of that counter this PR doesn't remove.
- SATS, BOME, MEME, MEW and BRETT are on the list but still served everywhere. Their absence is now treated as expected, so if one endpoint drops them before the others there is no rotation, and after the 30s TTL their prices are gone. These are close-only markets, where users need prices to exit.

A dedicated set of keeper-delisted markets would fit better, ideally shared with `SKIPPED_KEEPER_MARKETS` so the two stop drifting. Adding OM and WELL to `DELISTING_MARKET_ADDRESSES` would fix the first half, but that list also drives the exit toasts, so it's a product call.

src/config/__tests__/markets.delistingOnlyTokens.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+
+  it("returns nothing when there are no delisting markets", () => {
+    const result = collectDelistingOnlyTokenAddresses({ markets, isDelistingMarket: () => false });
+
+    expect(result.size).toBe(0);
+  });
+});
+
+describe("getDelistingOnlyTokenAddresses", () => {
+  it("never contains collaterals of live markets and is memoized per chain", () => {
```
divhead: This passes with an empty set too, since it only checks what must not be in it. It wouldn't catch `getDelistingOnlyTokenAddresses` returning nothing, e.g. if the delisting markets ever get `enabled: false` and the `isMarketEnabled` filter drops them. Worth asserting one expected member, like the TON synthetic `0xB2f7cefaeEb08Aa347705ac829a7b8bE2FB560f3`.

src/domain/synthetics/tokens/tokenPricesCache.ts
```diff
@@ -0,0 +1,78 @@
...
+import type { TokenPricesData } from "sdk/utils/tokens/types";
+
+type ChainPricesCache = {
+  prices: TokenPricesData;
+  updatedAt: { [address: string]: number };
+};
+
+export type ReconcileTokenPricesParams = {
+  chainId: number;
+  /** Prices parsed from the tickers response */
```
divhead: Nit: the field-level JSDoc restates the names, and the "delisted tokens leave tickers" rationale is written in four places (here, `missedMarketPrices.ts`, `config/markets.ts` and `useTokenRecentPricesData.ts`). One short note at the call site would be enough.

divhead (approved):

PR #2830, opened 2026-09-07
Block settling funding on positions below min collateral FEDEV-4266

timgmxio:
## Summary

Fixes [FEDEV-4266](https://linear.app/gmx-io/issue/FEDEV-4266/bug-settle-funding-fees-fails-silently-on-positions-below-min).

Settling accrued funding creates a zero-size `MarketDecrease` that withdraws 1 wei of collateral, so the position stays open and the contract re-validates it. Two independent checks cancel such an order: `willPositionCollateralBeSufficient` (collateral alone, without pnl or fees, against the size times the larger of the market's `minCollateralFactor` and the open-interest-based one → `UnableToWithdrawCollateral`) and `validatePosition` (collateral after pending fees, plus pnl and capped price impact, minus the closing fee, against the size times `minCollateralFactor` → `LiquidatablePosition("min collateral for leverage")`). The settle window offered such positions anyway, even pre-selected them, and the failure toast said only "failed to settle".

### Changes

- **`getIsPositionBelowMinCollateralForLeverage` (sdk)** mirrors both checks, including the 1 wei the settlement withdraws and the open-interest-based factor via the existing `getMinCollateralFactorForPosition`. The liquidation price-impact capping is extracted from `getLiquidationPrice`, which repeated it twice.
- **Settle window**: rows the contract would cancel are blocked the way disabled markets are — checkbox disabled, tooltip naming the way out. The FEDEV-3907 negative-margin state is blocked the same way instead of only warning (its tooltip drops "reduce or"). Pre-selection and the submitted batch go through the same gate, so a row that crosses the line while the window is open drops out of the batch instead of failing on chain.
- **Failure toast**: `OrderCancelled.reasonBytes` is stored with the order status and decoded through the shared trade-history copy: "… failed to settle: Margin is below the minimum required for the position size". That copy now separates "min collateral for leverage" from an actual liquidation state, since a position can be below the factor while far from its liquidation price.

### Testing

- Unit: settle gate spec (collateral-only check, exact 1-wei boundary, open-interest factor, pending fees, pnl, closing fee, negative margin, pre-selection), trade-history copy spec, and a spec decoding real `reasonBytes` captured from the contract. `tscheck`, eslint, prettier pass.
- Arbitrum fork (anvil, block 501618720): the gate's boundary was bisected against `SimulationRouter.simulateExecuteLatestOrder` on five planted positions (LIT long and short with USDC, LIT long with WETH collateral, ETH long with index-token collateral, ETH short). In every case the app blocks at exactly the price where the contract starts cancelling. A NATGAS position with collateral below the requirement but positive pnl is cancelled by the contract (`UnableToWithdrawCollateral`) and is now blocked; the current window version offers it and the order gets cancelled.
- Browser against the fork: blocked rows disabled with the right tooltips, healthy rows settle in one batch, pre-selection skips blocked rows; the toast round-trip verified by emitting the keeper's `OrderCancelled` with the real reason bytes.

### Notes

- Overlaps #2767 (FEDEV-4131) only in the `.po` catalogs. Both branches port `isPositionLiquidatable` separately; unifying this gate with `getResultingPositionMarginState` is a follow-up once both land.
- A settle cancelled by `UnableToWithdrawCollateral` in the submit-to-execution race shows the shared "Can't withdraw collateral…" copy; rare now that the gate covers that check.
- `closingFeeUsd` in `PositionInfo` ignores the pro discount, so the gate is marginally stricter than the contract for pro accounts — the same known gap as FEDEV-4131.

Linear: https://linear.app/gmx-io/issue/FEDEV-4266/bug-settle-funding-fees-fails-silently-on-positions-below-min

### DIFF
--- a/sdk/src/utils/positions/utils.ts
+++ b/sdk/src/utils/positions/utils.ts
@@ -171,6 +171,50 @@ export function getLeverage(p: {
   return bigMath.mulDiv(sizeInUsd, BASIS_POINTS_DIVISOR_BIGINT, remainingCollateralUsd);
 }
 
+export function getLiquidationPriceImpactDeltaUsd(p: {
+  marketInfo: MarketInfo;
+  sizeInUsd: bigint;
+  sizeInTokens: bigint;
+  pendingImpactAmount: bigint;
+  isLong: boolean;
+  useMaxPriceImpact?: boolean;
+}) {
+  const { marketInfo, sizeInUsd, sizeInTokens, pendingImpactAmount, isLong, useMaxPriceImpact } = p;
+
+  const maxNegativePriceImpactUsd = -1n * applyFactor(sizeInUsd, marketInfo.maxPositionImpactFactorForLiquidations);
+
+  if (useMaxPriceImpact) {
+    return maxNegativePriceImpactUsd;
+  }
+
+  let priceImpactDeltaUsd = getPriceImpactForPosition(marketInfo, -sizeInUsd, isLong, {
+    fallbackToZero: true,
+    sizeDeltaInTokens: sizeInTokens,
+  }).priceImpactDeltaUsd;
+
+  if (priceImpactDeltaUsd > 0) {
+    priceImpactDeltaUsd = capPositionImpactUsdByMaxPriceImpactFactor(marketInfo, sizeInUsd, priceImpactDeltaUsd);
+  }
+
+  const pendingImpactUsd = convertToUsd(
+    pendingImpactAmount,
+    marketInfo.indexToken.decimals,
+    pendingImpactAmount > 0 ? marketInfo.indexToken.prices.minPrice : marketInfo.indexToken.prices.maxPrice
+  )!;
+
+  priceImpactDeltaUsd = priceImpactDeltaUsd + pendingImpactUsd;
+
+  if (priceImpactDeltaUsd > 0) {
+    return 0n;
+  }
+
+  if (priceImpactDeltaUsd < maxNegativePriceImpactUsd) {
+    return maxNegativePriceImpactUsd;
+  }
+
+  return priceImpactDeltaUsd;
+}
+
 export function getLiquidationPrice(p: {
   sizeInUsd: bigint;
   sizeInTokens: bigint;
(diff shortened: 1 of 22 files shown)


sdk/src/utils/positions/utils.ts
```diff
@@ -317,37 +338,65 @@ export function getMinCollateralUsdForLiquidationPrice(p: {
...
-  let liquidationCollateralUsd = applyFactor(sizeInUsd, marketInfo.minCollateralFactorForLiquidation);
-  if (liquidationCollateralUsd < minCollateralUsd) {
-    liquidationCollateralUsd = minCollateralUsd;
+  const collateralUsdAfterDelta = convertToUsd(
+    collateralAmount - collateralDeltaAmount,
+    collateralToken.decimals,
+    collateralToken.prices.minPrice
+  )!;
+
+  if (collateralUsdAfterDelta < applyFactor(sizeInUsd, getMinCollateralFactorForPosition(position, 0n))) {
```
divhead, line 385: This mirrors `willPositionCollateralBeSufficient`, but the same `if` in `DecreasePositionUtils.decreasePosition` has a second condition that also cancels a zero-size decrease ([L152-L157](https://github.com/gmx-io/gmx-synthetics/blob/b5523f7fddb1d86f9e7004080413c582eccd8d6d/contracts/position/DecreasePositionUtils.sol#L152-L157)):

```solidity
if (
    !willBeSufficient ||
    (estimatedRemainingCollateralUsd + cache.estimatedRemainingPnlUsd) < cache.minCollateralUsd
) {
    if (params.order.sizeDeltaUsd() == 0) {
        revert Errors.UnableToWithdrawCollateral(estimatedRemainingCollateralUsd);
    }
```

For a settle that is `(collateral - 1 wei) * collateral min price + position pnl < MIN_COLLATERAL_USD`, and `MIN_COLLATERAL_USD` is $1 on Arbitrum and Avalanche. The gate misses it whenever `size * minCollateralFactor` is below that. Example: a $20 long with 1.50 USDC of collateral and -$0.58 pnl is selectable, and the keeper cancels the settle.

Such positions are also liquidatable under the "min collateral" rule, so the impact is small: the window until the liquidation, plus the execution fee. But it's the one branch the gate doesn't cover, and the fix is a single line: block when `collateralUsdAfterDelta + pnl < minCollateralUsd`, with `minCollateralUsd` from `selectPositionConstants`.

divhead (commented):
I went through the gate against `DecreasePositionUtils` / `PositionUtils` in gmx-synthetics `release-v2.2c`: the checks it implements match, and the `getLiquidationPrice` refactor is behaviour-preserving. Before approving:

1. The `MIN_COLLATERAL_USD` condition in the same contract check isn't mirrored. Small reach, one line.
2. The toast for `UnableToWithdrawCollateral` describes a collateral withdrawal instead of a settle.

The other inline comments are minor.

This PR and #2767 conflict in all 10 `.po` catalogs, whichever merges first. I'd merge #2767 first, since AC7 here refers to its copy. Then this branch re-extracts the catalogs and can reuse `PositionMarginFailureReason.MinCollateralForLeverage` from #2767 instead of the local `MIN_COLLATERAL_FOR_LEVERAGE_REASON`.

CI is red only because of the SDK keeper consistency test fixed on `release` in #2831. Merging `release` should turn it green.

sdk/src/utils/positions/utils.ts
```diff
@@ -317,37 +338,65 @@ export function getMinCollateralUsdForLiquidationPrice(p: {
...
-  return liquidationCollateralUsd - pnl - priceImpactDeltaUsd + closingFeeUsd;
+  const priceImpactDeltaUsd = getLiquidationPriceImpactDeltaUsd({
+    marketInfo,
+    sizeInUsd,
+    sizeInTokens,
+    pendingImpactAmount,
+    isLong,
+  });
+
+  const marginUsd = remainingCollateralUsd + pnl + priceImpactDeltaUsd - closingFeeUsd;
```
divhead: `collateralDeltaAmount` only reaches the collateral-only check above. `marginUsd` ignores it, while the contract runs `validatePosition` after the withdrawal. For a settle that's 1 wei and doesn't matter, but this is exported from the SDK with a generic name and parameter, so a caller that passes a real withdrawal gets a wrong answer. Either subtract `convertToUsd(collateralDeltaAmount, collateralToken.decimals, collateralToken.prices.minPrice)` here too, or drop the parameter and name the function for the zero-size case.

src/components/StatusNotification/FeesSettlementStatusNotification.tsx
```diff
@@ -25,6 +27,16 @@ type Props = {
   marketsInfoData: MarketsInfoData | undefined;
 };
 
+function getCancellationReason(reasonBytes: string | undefined) {
```
divhead, line 34: `UnableToWithdrawCollateral` has no trade-history copy, so it falls through to `getContractErrorMessage`. The toast then reads "Long ETH failed to settle: Can't withdraw collateral. Remaining would be $1.50, below minimum", about a withdrawal the user never made. That's the revert both from the collateral-only check and from the `MIN_COLLATERAL_USD` branch, so it's what users see when a position crosses the line between submit and execution because of open-interest growth or a collateral price move. A settle is always a zero-size decrease, so here it's safe to map `UnableToWithdrawCollateral` to the same "Margin is below the minimum required for the position size" copy.

sdk/src/utils/positions/utils.ts
```diff
@@ -171,6 +171,50 @@ export function getLeverage(p: {
   return bigMath.mulDiv(sizeInUsd, BASIS_POINTS_DIVISOR_BIGINT, remainingCollateralUsd);
 }
 
+export function getLiquidationPriceImpactDeltaUsd(p: {
```
divhead, line 174: There's a third copy of this block in `getEstimatedLiquidationTimeInHours` (`src/domain/synthetics/positions/utils.ts`, around line 100). Worth moving it to this helper too.

src/components/SettleAccruedFundingFeeModal/SettleAccruedFundingFeeRow.tsx
```diff
@@ -85,16 +85,42 @@ export const SettleAccruedFundingFeeRow = ({
...
-        className="ClaimSettleModal-checkbox flex self-center"
-      >
-        <div className="Exchange-info-label ClaimSettleModal-checkbox-label">{label}</div>
-      </Checkbox>
+      {blockedTooltipContent === undefined ? (
+        checkbox
+      ) : (
+        <TooltipWithPortal
+          position="top-start"
+          variant="none"
```
divhead, line 118: With `variant="none"`, disabled-market rows lose the dotted underline that hinted at the tooltip, and unlike the blocked rows they have no warning icon either. Nothing shows there's an explanation behind them anymore.

divhead (approved):

PR #2838, opened 2026-09-08
Keep acceptable price impact on liquidity updates FEDEV-4287

timgmxio:
## Summary

- **Fix (the ticket):** the "Acceptable price impact" field flashed `N/A` and dropped the user's value on every market-data refresh. `useTradeboxChanges` treated the auto-backfilled Margin/Size input strings as user changes (keys added in a026c3720 for the since-removed pool auto-selection), and `useTradeboxAcceptablePriceImpactValues` reset both values to `undefined` on any change. Those keys are removed; the hook now resets only on user actions (market, direction, order type, limit price, leverage, tokens), keeps the recommended value live from `recommendedAcceptablePriceDeltaBps`, and lets the selected value follow it only while the user hasn't customized it (`selected === default`, the same convention `PercentageInput` uses for "cleared"). Vitest on the hook; two of its cases fail on the old code.
  - Behavior change: editing size/margin (including the size slider) no longer resets a customized acceptable price impact; the recommended value and the low/high warnings stay live.
- **`PercentageInput`:** ↑/↓ arrow keys step the value by 0.01% (1 bps), clamped to `[0, max]`; from an empty field the step starts at the placeholder. Applies to every `PercentageInput` (allowed slippage, network fee buffer). Done via `onKeyDown` rather than `type="number"` to keep the shared text `NumberInput` sanitizer and avoid spinner/locale/wheel side effects.
- **`SuggestionInput`:** an empty suggestion list no longer opens the (empty) panel, so `PercentageInput` no longer disables its warning tooltip while the field is focused. The high/low acceptable price impact warning with "Set recommended impact" now shows while typing.
- **`Tooltip`:** a tooltip opened by focus flickered whenever the pointer left the handle (floating-ui hover close followed by the focus-within reopen). Hover-initiated closes are now ignored while the handle has visible focus, and the tooltip closes on blur unless the handle or the popup is hovered (so a link inside the popup stays clickable). Side effect for every tooltip with focusable content: after tabbing away the popup now closes instead of staying open until an outside click.

Verified on a live Arbitrum stand (Long → Limit, 1000 USDC, limit price 2400, "Set acceptable price impact" enabled) with a MutationObserver on the form: before — 22 `N/A` flashes and 22 input remounts in 40 s, a typed 1% wiped within 0.4 s; after — 38 liquidity updates in 70 s with 0 flashes and the typed value intact; the focused warning tooltip survives repeated pointer sweeps with zero DOM changes.

Linear: https://linear.app/gmx-io/issue/FEDEV-4287/bug-acceptable-price-impact-field-flickers-when-available-liquidity

### DIFF
--- a/src/components/PercentageInput/PercentageInput.tsx
+++ b/src/components/PercentageInput/PercentageInput.tsx
@@ -1,5 +1,5 @@
 import cx from "classnames";
-import { ReactNode, useEffect, useMemo, useState } from "react";
+import { KeyboardEvent, ReactNode, useEffect, useMemo, useState } from "react";
 import { useLatest } from "react-use";
 
 import { BASIS_POINTS_DIVISOR } from "config/factors";
@@ -12,6 +12,8 @@ import TooltipWithPortal from "components/Tooltip/TooltipWithPortal";
 export const NUMBER_WITH_TWO_DECIMALS = /^\d+(\.\d{0,2})?$/; // 0.00 ~ 99.99
 
 export const MAX_PERCENTAGE_INPUT_VALUE = 99 * 100; // basis points
+
+export const PERCENTAGE_INPUT_STEP = 1; // basis points
 
 function getValueText(value: number) {
   return roundToTwoDecimals((value / BASIS_POINTS_DIVISOR) * 100).toString();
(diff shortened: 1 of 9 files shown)


qwinndev:
Custom price impact can still be overwritten — [hook, line 60](https://github.com/gmx-io/gmx-interface/blob/3e9aa9a7be799b06a83575a36f8d50df168819cd/src/components/TradeBox/hooks/useTradeboxAcceptablePriceImpactValues.ts#L60). Set 0.40% while the recommendation is 0.35%. If subsequent recommendations become 0.40%, then 0.45%, the selected value becomes 0.45%. Track customization separately from the changing recommendation. This violates FEDEV-4287’s preservation requirement.

qwinndev:
acceptable price impact gets stuck at N/A.

Reproduced on Arbitrum, ETH Short → Limit: leave the default impact at 0.30%, then change the limit price from 2500 to 2500.01. The input disappears and remains unavailable while liquidity updates—confirmed for 30 seconds.

The reset clears the value, but initialization doesn’t rerun when the recommendation stays unchanged. The previous implementation handles this correctly.

qwinndev (approved):

PR #2840, opened 2026-09-09
Say the annualized buy pressure is GMX only

divhead:
FEDEV-4292

The Total stats fee tooltip said "Annualized buy pressure" next to combined GMX and GMTrade fees, which left it open whether GMTrade contributes to GMX buybacks. It does not: its allocation bought GT and stopped entirely on 2026-01-16.

- The label reads "Annualized GMX buy pressure".
- A line under it states that GMTrade fees do not contribute to GMX buybacks. It wraps and is width-capped so it does not clip on mobile.
- The calculation was already correct: only the V1 30% and V2 27% allocations of the EVM networks feed it. The locals are renamed to `v1GmxBuyPressure` / `v2GmxBuyPressure` / `annualizedGmxBuyPressure` so the code says the same thing as the label. With zero eligible EVM fees and positive GMTrade fees the figure is zero.

Combined fees and the plain annualized figure keep including GMTrade.

The API side of this issue is gmx-io/gmx-api#184, which names the token each network's buyback allocation buys so a combined value can no longer be read as GMX.

### DIFF
--- a/src/locales/de/messages.po
+++ b/src/locales/de/messages.po
@@ -311,6 +311,10 @@ msgstr "Limit fehlgeschlagen"
 msgid "Performance"
 msgstr "Leistung"
 
+#: src/pages/Dashboard/OverviewCard.tsx
+msgid "Annualized data based on the past 7 days. GMTrade fees do not contribute to GMX buybacks."
+msgstr ""
+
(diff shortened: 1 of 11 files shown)


divhead:

PR #2841, opened 2026-09-09
Show one sorted row per network in the stats tooltips

divhead:
FEDEV-4293

The network breakdowns split a chain across "V1 Arbitrum" and "V2 Arbitrum" rows, labelled Solana as "GMTrade Solana" or "Solana (GMTrade)", and kept object insertion order, so comparing networks meant adding rows up by eye.

- One row per network: Arbitrum, Avalanche, MegaETH, Solana. V1 and V2 are summed into their network by `sumNetworkParts` before the row is built, so contributions and totals are unchanged.
- `ChainsStatsTooltipRow` sorts rows by their unrounded value, highest first. `Array#sort` is stable, so equal values keep a fixed order.
- TVL and GM pools used their own fixed-order rows; they now go through the same sort. A network whose value has not arrived yet sorts last rather than being ordered as zero.
- Total stays below the rows, and the annualized fees and GMX buy pressure block stays a separate summary below Total.

Presentation only: no metric definition, contribution or headline total changes.

Note on missing versus zero: today `useV2Stats` reports absent market data as `0n`, so a real zero and a loading network are the same value here. That is fixed in FEDEV-4295 (#2842); this PR keeps the current filter so loading does not start rendering "$0.00".

Merge order: this PR and #2842 both touch `ChainsStatsTooltipRow`, `OverviewCard` and `StatsCard`. Merging this one first keeps the conflict to renaming the `SOLANA_ENTRY` constant there.

### DIFF
--- a/src/components/StatsTooltip/ChainsStatsTooltipRow.tsx
+++ b/src/components/StatsTooltip/ChainsStatsTooltipRow.tsx
@@ -6,13 +6,22 @@ import { formatAmountHuman } from "lib/numbers";
 
 import "./StatsTooltip.css";
 
+type EntryValue = bigint | number | string | undefined;
+
 type Props = {
-  entries: { [key: string]: bigint | number | string | undefined };
+  entries: { [key: string]: EntryValue };
   showDollar?: boolean;
   decimalsForConversion?: number;
   symbol?: string;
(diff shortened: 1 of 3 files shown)


divhead:

PR #2842, opened 2026-09-09
Tell partial and stale stats apart from complete totals

divhead:
FEDEV-4295

`useV2Stats` reported every missing EVM value as `0n`, and Solana values that were unavailable were added as zero or dropped. A loading or failing source therefore looked like a confirmed zero inside a total presented as complete, which is how Arbitrum GM pools could read $0.00 next to a $34.54m total.

- `useV2Stats` returns `undefined` for market, volume, fee and user data that has not arrived. A zero now means zero.
- `totalGmTvl` and `displayTvl` stay `undefined` until every contribution is known, including Solana, which was previously folded in as `?? 0n`.
- `ChainsStatsTooltipRow` distinguishes three states: a known value, including a real zero, which used to be filtered out as falsy; an unavailable value; and a value the API flagged as behind. The total is the sum of what is known, and under it the tooltip names the sources that have not answered ("Partial total: no data yet from ...") and the ones included but lagging ("Included but not up to date: ...").
- `isProtocolStatsNetworkStale` reads `meta.sources[].health` from the stats API, which keeps summing a lagging source and only flags it, so the interface has to read the flag to show it.

Recovery is automatic: once the sources answer, the notes disappear and the totals and breakdowns come back.

Not in this PR: the headline number itself carries no marker, the state is communicated in its tooltip. Putting an indicator on the figure needs a design call.

Merge order: #2841 touches the same three files. Merging it first leaves only the `SOLANA_ENTRY` constant to reconcile here.

### DIFF
--- a/src/components/StatsTooltip/ChainsStatsTooltipRow.tsx
+++ b/src/components/StatsTooltip/ChainsStatsTooltipRow.tsx
@@ -12,6 +12,7 @@ type Props = {
   decimalsForConversion?: number;
   symbol?: string;
   subtotal?: ReactNode;
+  staleTitles?: string[];
 };
 
 export default function ChainsStatsTooltipRow({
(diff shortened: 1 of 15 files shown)


divhead:

PR #2845, opened 2026-09-09
Show a loading state instead of Connect wallet while the session is restored FEDEV-4253

timgmxio:
## Summary
- On reload the app rendered "Connect wallet" in the header and in the forms until Privy finished restoring the session. wagmi does remember the last connection (`wagmi.store.current`), but wagmi 2.19 `createConfig` persists its empty initial state before hydration, and with the `ssr: true` that `@privy-io/wagmi` forces this wiped the marker on every load. `getWagmiConfig` now reads the persisted store before `createConfig` and `WalletProvider` hands it back through wagmi's own `initialState`; the spec covers it and notes when the workaround can be dropped.
- `useIsWalletInitializing` is true only while a remembered connection exists, wagmi has no account yet and Privy has not resolved its wallets. Visitors without a session still get "Connect wallet" on the first frame, and in-app Disconnect clears the marker, so nothing stays stuck in the loading state.
- Header: while the session is restored the button slot shows a skeleton shaped like the account button (same box, avatar circle and address bar, app skeleton shimmer), so the address replaces it without a layout shift.
- Trade, GM/GLV and shift submit buttons show a disabled "Connecting wallet..." during that window instead of a clickable "Connect wallet" (new string translated in all locales). Referrals tabs show their loader and Earn distributions hide the connect CTA meanwhile; Earn Portfolio already used the hook.

Linear: https://linear.app/gmx-io/issue/FEDEV-4253/bug-connect-wallet-flashes-while-restoring-a-connected-session-on

## CI
Audit is red on the vitest 3.2.6 advisory (GHSA-82fw-gwwq-j7x9, `yarn npm audit`). This PR changes no dependencies, and the same job fails on the other release-based PRs today. The bump to vitest 4.1.11 (#2844) landed on `master`, not on `release` yet, so the check stays red here until that reaches `release`.

### DIFF
--- a/src/lib/wallets/walletConfig.spec.ts
+++ b/src/lib/wallets/walletConfig.spec.ts
@@ -1,8 +1,9 @@
+import { serialize } from "@wagmi/core";
 import { describe, expect, it } from "vitest";
 
 import { ARBITRUM, DEFAULT_SETTLEMENT_CHAIN_ID } from "config/chains";
 
-import { getSupportedChains } from "./walletConfig";
+import { getSupportedChains, getWagmiInitialState } from "./walletConfig";
 
 describe("getSupportedChains", () => {
   it("uses Arbitrum as the default settlement chain", () => {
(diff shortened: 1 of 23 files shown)


src/lib/wallets/walletConfig.spec.ts
```diff
@@ -21,3 +22,27 @@ describe("getSupportedChains", () => {
     expect(new Set(chainIds).size).toBe(chainIds.length);
   });
 });
+
+describe("getWagmiInitialState", () => {
+  // wagmi 2.19 runs store.setState(getInitialState()) inside createConfig, and with the ssr: true that
```
qwinndev, line 27: do we use ssr: true or it is about test env?

qwinndev (approved):

PR #2865, opened 2026-09-14
Fix homepage scrolling in Telegram iOS browser FEDEV-4286

qwinndev:
## Summary

Fix content appearing above the homepage header in Telegram's iOS browser for [FEDEV-4286](https://linear.app/gmx-io/issue/FEDEV-4286/bug-gmx-homepage-visual-glitch-in-telegram-ios-in-app-browser).

- Detect Telegram's injected bridge before rendering, keep the document stationary, and scroll inside a container clipped to the dynamic viewport.
- Use that container for existing section links and navigation scroll resets.

Telegram [allows its native scroll view to draw beyond its bounds](https://github.com/TelegramMessenger/Telegram-iOS/blob/master/submodules/BrowserUI/Sources/BrowserWebContent.swift#L350). Containing the page's scrolling prevents that content from appearing under its browser controls. The change applies only when Telegram's bridge is present.


landing/src/index.tsx
```diff
@@ -14,6 +14,10 @@ import { captureLandingUtmParams } from "./utils/utm";
 captureLandingReferralCode();
 captureLandingUtmParams();
 
+if ("TelegramWebviewProxy" in window) {
```
timgmxio: `TelegramWebviewProxy` isn't iOS-specific: Telegram Android's in-app browser injects the same bridge into regular pages ([BotWebViewContainer.java](https://github.com/DrKLO/Telegram/blob/62b56a07ca7e30e39f7fd00a6728d6bbd716ca1c/TMessagesProj/src/main/java/org/telegram/ui/web/BotWebViewContainer.java#L503-L512)). So Android gets the fixed-body mode too, although the root cause (`clipsToBounds = false` on the iOS scroll view) doesn't exist there and it hasn't been checked on Android. And it's not a no-op there: the Android browser sheet partly decides on its pull-down gesture from the WebView's own `scrollY` ([ChatAttachAlertBotWebViewLayout.java](https://github.com/DrKLO/Telegram/blob/62b56a07ca7e30e39f7fd00a6728d6bbd716ca1c/TMessagesProj/src/main/java/org/telegram/ui/bots/ChatAttachAlertBotWebViewLayout.java#L876)), which stays 0 in this mode.

Unless Android is intended, let's scope it to iOS (plus `import { isIOS } from "lib/headlessUiIsMobile";`):

```suggestion
if ("TelegramWebviewProxy" in window && isIOS()) {
```

timgmxio (dismissed):

timgmxio (approved):

PR #2873, opened 2026-09-14
Sunset legacy Buy GMX page and use the shared Buy GMX modal FEDEV-4188

timgmxio:


src/components/BuyGmxModal/BuyGmxModal.tsx
```diff
@@ -23,29 +17,7 @@ import SpinnerIcon from "img/ic_spinner.svg?react";
 
 import { BUY_GMX_MODAL_LINKS } from "./buyGmxModalConfig";
 
-const DIRECT_BUY_PATH = "/trade/swap";
-
-const ARB_USDC_ADDRESS = "0xaf88d065e77c8cC2239327C5EDb3A432268e5831";
-const ARB_GMX_ADDRESS = getContract(ARBITRUM, "GMX");
-
-function setArbitrumSwapToGmxOptionsInLocalStorage() {
```
qwinndev, line 31: With Short preferences saved on Arbitrum, fully reload Stats, wait for market data, then select Buy GMX → Buy GMX on GMX swap. This opens a GMX/USD Short instead of Market Swap USDC→GMX. The deferred tradebox initialization restores the saved preferences after the query processor queues the Swap configuration, rewriting the route to /trade/short.

src/pages/Earn/buyGmxDeepLink.ts
```diff
@@ -0,0 +1,28 @@
...
+
+  useEffect(() => {
+    if (!isRequested) {
+      return;
+    }
+
+    setIsVisible(true);
+
+    const params = new URLSearchParams(history.location.search);
+    params.delete(BUY_GMX_QUERY_PARAM);
```
qwinndev: With Arbitrum selected, open /#/buy_gmx?chainId=43114. With the wallet disconnected, the app switches to Avalanche but lands on Earn Portfolio without the Buy GMX modal. With a connected wallet, the requested network switch appears not to happen either.

The confirmed modal failure comes from immediately removing buyGmx: the subsequent network change remounts the route subtree through <SWRConfig key={chainId}>, resetting modal state without a remaining request to reopen it. The connected-wallet behavior needs separate investigation.

qwinndev (changes requested):

src/components/BuyGmxModal/BuyGmxModal.tsx
timgmxio: Confirmed, fixed in 5b5fdca81b.

The modal wasn't the cause: on the first tradebox sync, `handleChainChange` in `useTradeboxState` re-applied the saved localStorage options with a plain `setStoredOptionsOnChain(saved)`. Coming from Stats the markets are already cached, so it runs in the same commit as `useTradeParamsProcessor`, after its queued `setTradeConfig`; the saved Short overwrote the Swap config and the trade type sync rewrote the path to `/trade/short`.

The state is already initialized from that same entry, and the provider remounts on every network change (`SWRConfig key={chainId}`), so the saved options are now re-applied only when syncing to another chain. Checked: saved Short → Stats → Buy GMX on GMX swap opens Market Swap USDC → GMX; `/trade` still restores the saved Short and `/trade/long` still opens Long.

src/pages/Earn/buyGmxDeepLink.ts
timgmxio: Confirmed, fixed in 5b5fdca81b.

The visibility is now derived from `buyGmx` in the url, and the param is removed only when the modal is closed, so the remount on the network change reopens it (a refresh with the modal open keeps it open too). The `/buy_gmx` redirect also carries the incoming query now: `/earn/portfolio?chainId=43114&ref=…&buyGmx=1`.

The connected-wallet part isn't specific to this url: `/stats?chainId=43114` behaves the same, and this PR doesn't touch that handling. `AppRoutes` processes `chainId` on the first commit, before Privy restores the wallet connection, so `switchChain` has no connector: only the app selection changes, and the wallet chain overrides it once the wallet reconnects. FEDEV-2725 fixes this by deferring `chainId` processing until the wallet is initialized; carrying the query through the redirect lets that apply to `/#/buy_gmx?chainId=…` as well. Checked with the FEDEV-2725 work in progress: the wallet is asked to switch, and the page lands on Avalanche with the modal open.

qwinndev (approved):

PR #2875, opened 2026-09-15
Hotfix: keep the landing figures unmarked while their sources load

divhead:
FEDEV-4295 follow-up. The homepage figures turned yellow as partial for the second between the first and the last network answering on every page load. A figure now shows a dash until all its sources have answered once or five seconds have passed; after that a missing or stale network is reported in the hover notice only, with a dotted underline as the cue, and the figure itself is not coloured.


timgmxio (approved):

tanoeth:
**QA pass** on the previews for `acec688`: homepage at 3b4e9dba.gmx-interface-home.pages.dev, Stats at hotfix-landing-partial-flash.gmx-interface.pages.dev/stats. Tested in Chrome desktop, and the homepage also at phone width. I blocked the Solana stats source (`/v1/stats/*`) in the browser to create the missing and stale states.

Homepage (Traders, Open interest, Total volume) and Stats page (all 10 Overview and Stats figures):
- First load with all sources healthy: dash, then the full value in normal colour. No yellow and no partial number in between.
- Source blocked on load: dash for 5 s, then the partial total in normal colour. The notice "Partial total: no data yet from Solana." shows on hover, and on tap at phone width (homepage). Homepage figures get the dotted underline, and the Stats tooltip keeps its yellow notice line.
- Source recovers: full total on the next refresh, no reload, and the notice goes.
- Source fails after a healthy load: the value stays, no dash, and the notice reads "Included but not up to date: Solana (…)".

Seen once and accepted: on first load, a network slower than 5 s shows a short total in normal colour until it answers. Stats TVL showed $40.31m instead of $282m for 18 s while an Arbitrum source was slow. This matches the spec and only happens on first load.

PR #2877, opened 2026-09-15
Add targeted USDG pools announcement for top LP wallets FEDEV-4346

timgmxio:


src/components/WhatsNewToast/WhatsNewToastContainer.tsx
```diff
@@ -128,6 +133,13 @@ export function WhatsNewToastContainer() {
               </div>
             </motion.div>
           )}
+          {isUsdgPoolsAnnouncementVisible && (
```
qwinndev: ### [P2] Hide the USDG banner on `/announcements`

With `showUsdgPoolsAnnouncement.enabled = true`, an eligible, non-dismissed wallet connected on Arbitrum still sees this banner on `/announcements`. [FEDEV-4346](https://linear.app/gmx-io/issue/FEDEV-4346/in-app-targeted-announcement-usdg-whitelist-invitation-top-lp-wallets) explicitly requires it to be hidden on that page.

This condition only checks campaign eligibility. `WhatsNewToastContainer` is mounted outside the route switch, and the existing `!isAnnouncementsPage` guard applies only to the carousel. Please add that guard to the USDG render condition and cover `/announcements` with a regression test.

Reproduced in a rendering test using the actual campaign hook and banner, with mocked wallet/flag inputs. The live Arbitrum flag was absent when checked, so the issue becomes observable once the campaign is enabled for a qualifying wallet.

timgmxio: Fixed in 87c50032c0: the USDG render condition now carries the same `!isAnnouncementsPage` guard as the carousel, and `WhatsNewToastContainer.spec.tsx` covers it with the real campaign hook and mocked wallet/flag inputs (shown on `/trade`, hidden on `/announcements`; the spec fails without the guard).

qwinndev (dismissed):

qwinndev (approved):

timgmxio:
Superseded by #2878, which targets `master` as a hotfix. Same change, ported onto the master base: this branch is cut from `release`, so retargeting it would have carried the FEDEV-4343 commits into master as well, and `isAnnouncementsPage` does not exist on master (it arrived with FEDEV-4343), so the /announcements guard defines it locally there.

PR #2881, opened 2026-09-16
Drop an outdated One-Click Trading approval instead of resending it FEDEV-3297

divhead:
FEDEV-3297

An approval already used on-chain kept being attached to the next 1CT order until the subaccount data refreshed, so the order reverted with InvalidSubaccountApprovalNonce.


timgmxio (commented):
1. `SyntheticsEventsProvider.tsx:1181`: the new block does not set `isViewed = true`, unlike the `InsufficientExecutionFee` and `InvalidSignature` branches. For a new order with `metricId`, the metric branch writes `isViewed: false` at line 1210 over the `isViewed: true` set by `OrderStatusNotification`, so on the next effect pass the generic "failed" toast (line 1200) fires in addition to the outdated-approval toast shown 500 ms later. Setting `isViewed = true` where the outdated-approval toast is shown removes the second toast.

2. `errorToasts.tsx:147`: the `isOutdatedSubaccountApproval` check comes after the `isInternalSwapFallback` / `isExternalSwapFallback` branches. For a batch with an external swap, `getIsPossibleExternalSwapError` matches any "execution reverted" message, so a nonce revert on simulation gets the swap-fallback toast and `fallbackToInternalSwap` runs. The specific check should precede the fallback branches, and the fallback should be skipped when the nonce error is matched.

3. `SyntheticsEventsProvider.tsx:1188`: `extractRelayTaskError` is recomputed for the same task status; `relayError` from line 1127 can be hoisted and reused.

4. Locales: `msgstr` is empty for the new string in all non-English locales.

5. Perhaps out of scope for this PR: the stale window is created on the success side. A relayed txn with a non-empty `subaccountApproval` has consumed the nonce, but the approval keeps being attached until the next multicall refresh. `handleExpressTxnSuccess` already receives the approval; an optimistic update of the cached onchain data there could prevent the failed attempt instead of cleaning up after it. Plain removal would not be enough for first activation, since the cached `active` flag is still `false` at that point.

divhead:
Thanks, 1–4 are fixed in fecf663: the outdated-approval branch now marks the txn viewed, so the generic toast no longer follows it; the nonce check runs before the swap fallbacks and disables them; relayError is computed once; translations are filled for every locale except pseudo.

On 5: agreed, it would remove the remaining window (the multicall round trip after OrderCreated). The optimistic update has to mirror the contract's approval transitions (nonce vs multichain nonce, first activation, limits), so I'd take it as a separate issue with its own test.

timgmxio (approved):

